PR #75, opened 2015-09-30
[search] Fixed split of retrieved features to suggest/not suggest features.

ygorshenin:
For instance, when searching for "улица правды" only two results are displayed in the results list, while there are around 100 of them in Russia_Central and Russia_Southern. The problem is that some retrieved results are removed by mistake from the search output during suggestions process.


ygorshenin:
@vng PTAL

PR #112, opened 2015-10-03
FreeBSD support

AMDmi3:


mapsmetest:

AMDmi3:
Хм, test this please

AMDmi3:
test this please

routing/routing_integration_tests/routing_integration_tests.pro
```diff
@@ -11,7 +11,7 @@ CONFIG -= app_bundle
 TEMPLATE = app
 
 ROOT_DIR = ../..
-DEPENDENCIES = map routing search storage indexer platform geometry coding base osrm jansson protobuf tomcrypt succinct stats_client
+DEPENDENCIES = map routing search storage indexer platform geometry coding base osrm jansson protobuf tomcrypt succinct stats_client opening_hours
```
mnemonicflow: Any reason why the opening_hours target was added to the routing integration tests?

common.pri
```diff
@@ -61,7 +61,7 @@ RCC_DIR = $$TEMP_PATH
 MOC_DIR = $$TEMP_PATH
 UI_DIR = $$TEMP_PATH
 
-QMAKE_LIBDIR *= $$BINARIES_PATH
+LIBPATH = $$BINARIES_PATH $$LIBPATH
```
mnemonicflow: Why have you changed the variable name? Variable LIBPATH is deprecated; use QMAKE_LIBDIR instead

configure.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
Zverik: Почему? Bash-а нет на freebsd? Некоторые скрипты (едва ли этот) используют спец. возможности баша.

routing/routing_integration_tests/routing_integration_tests.pro
Zverik: Как-то это не в тему про freebsd?

platform/platform_freebsd.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+#include <stdlib.h>
+#include <unistd.h>
+
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/sysctl.h>
+
+
+/// @return directory where binary resides, including slash at the end
```
ygorshenin: We do not use static functions, we use anonymous namespaces instead.

platform/platform_freebsd.cpp
```diff
@@ -0,0 +1,142 @@
...
+static bool GetBinaryFolder(string & outPath)
```
ygorshenin: s/Folder/Dir/g

platform/platform_freebsd.cpp
```diff
@@ -0,0 +1,142 @@
...
+    outPath = path;
+    outPath.erase(outPath.find_last_of('/') + 1);
+    return true;
+  }
+  return false;
+}
+
+Platform::Platform()
+{
+  // init directories
```
ygorshenin: All comments except inline should start from a capital letter and end with a dot.

platform/platform_freebsd.cpp
```diff
@@ -0,0 +1,142 @@
...
+  string path;
+  CHECK(GetBinaryFolder(path), ("Can't retrieve path to executable"));
+
+  char const * homePath = ::getenv("HOME");
+  string const home(homePath ? homePath : "");
+
+  m_settingsDir = home + "/.config/MapsWithMe";
```
ygorshenin: Please, use JoinFoldersToPath() function.

platform/platform_freebsd.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+  if (!IsFileExistsByFullPath(m_settingsDir + SETTINGS_FILE_NAME))
+  {
+    mkdir((home + "/.config/").c_str(), 0755);
```
ygorshenin: Why not to use Platform::MkDir()

platform/platform_freebsd.cpp
```diff
@@ -0,0 +1,142 @@
...
+    mkdir(m_settingsDir.c_str(), 0755);
+  }
+
+  m_writableDir = home + "/.local/share/MapsWithMe";
+  mkdir((home + "/.local/").c_str(), 0755);
+  mkdir((home + "/.local/share/").c_str(), 0755);
+  mkdir(m_writableDir.c_str(), 0755);
+
+  char const * resDir = ::getenv("MWM_RESOURCES_DIR");
+  if (resDir)
```
ygorshenin: This function looks like platform/platform_linux.cpp. It's not good to just duplicate it's code here. Please, reuse old code instead of copy-and-paste.

platform/platform_freebsd.cpp
```diff
@@ -0,0 +1,142 @@
...
+    return content.substr(0, 32);
+  }
+  else
+    return "n0dbus0n0hostid00000000000000000";
+
+}
+
+void Platform::RunOnGuiThread(TFunctor const & fn)
+{
+  /// @todo
```
ygorshenin: Who will fix that @todo?

platform/platform_freebsd.cpp
```diff
@@ -0,0 +1,142 @@
+#include "platform/platform.hpp"
```
Zverik: Насколько сложно вписать это в platform_linux, насыпав `#ifdef`?

ygorshenin: The correct way is to create a separate platform_desktop_unix.{cpp|hpp} where all common Linux/FreeBSD code should reside. Linux is not FreeBSD and FreeBSD is not Linux. All Linux/FreeBSD-specific methods should be in platform_linux.cpp and platform_freebsd.cpp correspondingly.

therearesomewhocallmetim:
Please, note that currently we have __no__ FreeBSD servers, and in order to test this, we will have to install one.

common.pri
AMDmi3: Ok

configure.sh
AMDmi3: No, there's no bash on FreeBSD by default, and making a script that doesn't need it depend on it is a bad practice.

routing/routing_integration_tests/routing_integration_tests.pro
AMDmi3: Will split to separate PR.

platform/platform_freebsd.cpp
AMDmi3: (replying to random comment)

This file is based on platform_linux.cpp, and it actually feels quite strange to be asked for style fixes in the code I haven't touched and to follow style upstream code doesn't. Will use common linux/freebsd code and fix bits that can be fixes though.

ygorshenin: platform_linux.cpp was written before style guide was established, and, of course, it should be rewritten in accordance with our modern practices. But I can't approve *new* code (platform_freebsd.cpp) which follows old and wrong style.

Zverik: So basically, to get this PR accepted, you should make your changes in `platform_linux.cpp` instread.

ygorshenin: Yep, that's right. I'll fix it ASAP and notify you when it'll be fixed.

AMDmi3: Thank you!

platform/platform_freebsd.cpp
AMDmi3: No idea. I don't see why it should be different from Linux, and I don't know why it hasn't been implemented for Linux. I guess I'll leave it in unix-generic code.

ygorshenin: Agree, these functions will go to unix-generic code.

platform/platform_freebsd.cpp
deathbaba: The better way can be to refactor our Platform class, probably by returning to initial virtual class methods solution and their implementation for each platform. Current approach is faster at runtime but very hard to understand and support.

AMDmi3: :+1:, virtual base class would be much easier to handle. Now, as all stuff is initialized in ctor, you can't basically use virtual methods, so you have to parametrize base class ctor with e.g. executable path, which is non-flexible.

deathbaba: Exactly. We used virtual base class in 2010, then moved to current solution, but now I see that it was bad idea.

I think it would be great to refactor whole platform to use more clear virtual base class, rather than patch existing code (and people who port for other platforms will be grateful too). What do you think, is it feasible for you? (in a separate request, of course).

AMDmi3: Sure, I can try.

mpimenov: Some say that even Linux is not Linux.
It's GNU/Linux.

platform/platform_freebsd.cpp
```diff
@@ -0,0 +1,142 @@
...
+  LOG(LDEBUG, ("Tmp directory:", m_tmpDir));
+  LOG(LDEBUG, ("Settings directory:", m_settingsDir));
+  LOG(LDEBUG, ("Client ID:", UniqueClientId()));
+}
+
+string Platform::UniqueClientId() const
+{
+  bool filterdashes = false;
+  string machineFile = "/var/db/dbus/machine-id";
+  if (IsFileExistsByFullPath("/etc/hostid")) {
```
mpimenov:
```
if (IsFileExistsByFullPath("/etc/hostid")) 
{
}
```

platform/platform_freebsd.cpp
ygorshenin: @deathbaba , why do we need virtuals at all? There is always one and only one implementation of Platform on each platform. It doesn't make sense to compile PlatformWin on Android, for example. We have been used separate compilation for Platform instead of virtual inheritance and it works fine. From my point of view virtuals are not better that current approach, because you'll get a bunch of PlatformWin, PlatformLinux, PlatformPosix, PlatformUnix, ... classes instead of a bunch of source files.

AMDmi3: You'll still have the same bunch of source files and Platform classes (though it's better to use different names and typedef used one to Platform) - there's no need to use run-time polymorphism here as there's only a single Platform implementation at a time. But virtual class will give you have strictly defined interface (you can forget to fill a variable in current implementation; you can't forget to implement virtual method) and an ability to keep common code in intermediate classes. My idea is

```
// platform.h
class IPlatform {
   virtual string GetExecutableName() = 0;
   virtual string GetSettingsDir() = 0;
};
```
```
// platform_unix.h
class UnixPlatform : public IPlatform {
    virtual string GetSettingsDir() override { return getenv("HOME") + "./mapsme"; }
};
```
```
// platform_freebsd.h
class FreeBSDPlatform : public UnixPlatform {
    virtual string GetExecutablePath() final { return ... }
};

typedef FreeBSDPlatform Platform; // all consumers work with Platform -
                                  // a final class, no runtime polymorphism
```

ygorshenin: OK, that makes sense.

deathbaba: @ygorshenin Virtual classes will make code much simpler and easier to support. And it’s very hard to customize member variables now for Platform (check Android implementation for example). I agree that we don’t need runtime polymorphism, but it’s just lesser evil.
And idea with defines to avoid unnecessary virtual calls sounds interesting :)

AMDmi3: > I agree that we don’t need runtime polymorphism, but it’s just lesser evil.

By

> there's no need to use run-time polymorphism here

I've meant that there will be no runtime polymorphism in fact. With C++11 override & final keywords properly used, when final subclass is used directly (our case), compilers can do devirtualization and avoid vtbl lookups completely. There will not be any noticeable effect on performance here anyway.

deathbaba: Great! Didn’t know about that kind of optimization, thanks.

deathbaba:
@AMDmi3 Any news? Should we close this one?

deathbaba:
Closing request as we discussed, that Platform should be refactored first to have base virtual class.

PR #121, opened 2015-10-05
Add missing dependency

AMDmi3:
Fixes link error:

    ...omim/search/intermediate_result.cpp:44: undefined reference to OSMTimeRange::OSMTimeRange...

### DIFF
--- a/routing/routing_integration_tests/routing_integration_tests.pro
+++ b/routing/routing_integration_tests/routing_integration_tests.pro
@@ -13,6 +13,11 @@ TEMPLATE = app
 ROOT_DIR = ../..
 DEPENDENCIES = map routing search storage indexer platform geometry coding base osrm jansson protobuf tomcrypt succinct stats_client
 
+# this dependency is not built on Linux
+!linux* {
+  DEPENDENCIES += opening_hours
+}
+
 macx-*: LIBS *= "-framework IOKit" "-framework SystemConfiguration"
 
 include($$ROOT_DIR/common.pri)


mapsmetest:

ygorshenin:
We do not compile opening_hours for linux, please, add this dependency conditionally, for freebsd only.

AMDmi3:
Why don't you build it? Likely it should be the same for FreeBSD.

ygorshenin:
The problem is that libopening_hours can't be built with linux clang3.5 due to missing codecvt header. And by default clang3.5 is shipped with Debian (Jessie) and CentOS. Of course, there exists a workaround - manually download and build clang3.6 with libc++, but this is a headache. Is it possible to add this dependency conditionally for freebsd only?

AMDmi3:
Shouldn't it be ```!linux*``` instead?

ygorshenin:
Yes, I think this is correct. @syershov WDYT?

syershov:
I think same.

ygorshenin:
OK, @AMDmi3, could you please add this check?

AMDmi3:
Done

ygorshenin:
LGTM

routing/routing_integration_tests/routing_integration_tests.pro
```diff
@@ -13,6 +13,11 @@ TEMPLATE = app
 ROOT_DIR = ../..
 DEPENDENCIES = map routing search storage indexer platform geometry coding base osrm jansson protobuf tomcrypt succinct stats_client
 
+# this dependency is not built on Linux
```
mpimenov: Зачем этот комментарий?

AMDmi3: It is obvious, no? To explain why this dependency is not enabled on Linux. That is not obvious at all.

ygorshenin: But your comment doesn't explain why this dependency is not enabled on Linux. It just claims that opening_hours is not built on Linux.

AMDmi3: And it shouldn't. Why opening_hours is not build on Linux should be explained in 3party/3party.pro

PR #123, opened 2015-10-05
Fix library path: prepend local lib path instead of appending

AMDmi3:
This fixes picking up system libraries instead of ones from 3party/


mapsmetest:

Zverik:
LGTM

PR #124, opened 2015-10-05
Add FreeBSD awareness to build system

AMDmi3:


mapsmetest:

syershov:
LGTM

syershov:
test this please

therearesomewhocallmetim:
JTTP

Zverik:
LGTM

therearesomewhocallmetim:
JTTP

deathbaba:
If freebsd is mostly equal to linux in our code, does it make sense to rename our linux keyword to e.g. POSIX? And do not copy-paste similar code? And to add correct #ifdef OMIM_OS_FREEBSD only at the places where do we have the difference?

AMDmi3:
I've suggested that (i.e define OMIM_OS_UNIX along with LINUX and FREEBSD and use that in conditions where applicable) on IRC, but we decided to go straightforward way first which may be refactored later. It's not too hard to implement it right away, however we may want the same thing for qmake code. I can also test it on DragonFlyBSD, which is FreeBSD-but-not-quite to ensure it's implemented in a generic way.

deathbaba:
Would love to avoid copy-paste in straight-forward way :)
qmake has `unix` target, which is a parent for it’s linux* and mac* mkspec targets (and we use it in few places already), so it can be a good start.

AMDmi3:
Okay, let me try to implement cross-platform support in a more generic way then, including Platform class refactoring. This and #112 may be closed I guess.

ygorshenin:
Alex, Android and iOS are also POSIX, so it's wrong to merge Linux, FreeBSD, Android and iOS to POSIX.

AMDmi3:
We're speaking of `unix' here, not posix, it should have narrower scope. Well, with OMIM_OS we may define it as we like, I'm more worried of qmake. Does `unix' in qmake include ios and android?

deathbaba:
As far as I remember, yes. Check mkspecs folder in Qt, qmake defines QMAKE_PLATFORM variable via includes, so there are linux, unix and posix there.

deathbaba:
@AMDmi3 Any news? Should we close this one?

deathbaba:
Closing request as we discussed, that Platform should be refactored first to have base virtual class.
And in theory with a correct 'PlatformUnix' derived class these changes would be not needed any more.

PR #122, opened 2015-10-05
Switch configure.sh to sh interpreter

AMDmi3:
Thiere is nothing in this script which requires bash, and bash is not available everywhere by default

PR #129, opened 2015-10-05
Fix configure.sh for sh shell

Zverik:
После #122 перестал запускаться configure.sh. Этот PR его чинит.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -2,7 +2,7 @@
 # Please run this script to configure the repository after cloning it.
 
 # Stop on the first error.
-set -e
+set -e -u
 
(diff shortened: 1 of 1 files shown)


deathbaba:
@AMDmi3 We would love if you test future changes on more than one platform :)

AMDmi3:
Unfortunately there's no way to test this change for me on any platform as it implies access to private repository. But you're right, I've missed a bashism, my bad.

deathbaba:
I mean, in general, it would be great to test any change on as many platforms, as possible.
Anyway, we have our own autobuild system which can help sometime (but not with such script, of course).

PR #141, opened 2015-10-06
Buildscript

hybridOL:
Two minor fixes for Linux builds
Linux Qt target should change to linux-clang-libc++ as g++ stdlib headers gave some problems with thread model. This was at least with gcc 4.8.4. AFAIK libc++ is standard library for clang anyway, so might be best to avoid the mixup on systems where clang and g++ are found.
Second is a problem with smaller machines where only one processor is available. The build script terminates with command line 'make -j 0', as 1/2 processors gets rounded to 0. I'm working in a VM, so only one proc here. Could be extended to avoid the 0 instead, but having PROCESSES at num_proc seems more suitable.


mapsmetest:

PR #150, opened 2015-10-07
Czech translation update

R3gi:
Czech translation update:
• Fix of grammatical, typographical and contextual issues
• Improved translation consistency

### DIFF
--- a/android/res/values-cs/strings.xml
+++ b/android/res/values-cs/strings.xml
@@ -167,9 +167,9 @@
 	<!-- Add bookmark dialog - bookmark set, Bookmarks dialog - Bookmark set cell -->
 	<string name="set">Skupina</string>
 	<!-- Text hint in Bookmarks dialog when no any bookmarks are added -->
-	<string name="bookmarks_usage_hint">Zatím nemáš žádné záložky.\nZáložku můžeš přidat přidržením libovolného místa na mapě.\nV MAPS.ME mohou být také zobrazeny importované záložky z jiných zdrojů. Soubor ve formátu KML/KMZ s uloženými záložkami lze otevřít např. z emailu, Dropboxu nebo webového odkazu.</string>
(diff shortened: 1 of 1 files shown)


mapsmetest:

trashkalmar:
Thank you for corrections!
1. But please modify strings.txt file only, then run tools/unix/generate_localizations.sh to create platform strings.
2. Create two commits: the first with modified strings.txt, the second - with generated strings.xml and localizable.lproj files.

R3gi:
@trashkalmar Thanks, how can I do that? What do you mean by "strings.txt" file? Android is using xml… I am still not so familiar with GitHub web interface and I am not a developer, only translator.

trashkalmar:
There is "strings.txt" file in the project root. This file is the storage of strings used both by Android and iOS applications, thats why we modify it only, not strings.xml.
First, you should commit modifications of strings.txt, then run tools/unix/generate_localizations.sh script and commit its results. It should be done in console on your local computer.

R3gi:
Hmm, I've downloaded GitHub for desktop, cloned the repository (oh, it's so large), edited and saved the strings.txt file on my local HDD, but I can't find a way how to run mentioned script in GitHub app. And I am using Windows, which if I am right doesn't support running *.sh scripts directly via cmd.exe. Is there some other option to submit the translation file?


I am worried that this way of submitting even basic translation updates is too much complicated for translators, it requires a lot of time and I am not sure if every translator want to install X programs just for submitting one localization file. In my opinion this should be changed. Using Crowdin.net or other translation platform would be much more user friendly…

Zverik:
Could you please sign the CLA? http://maps.me/cla/

R3gi:
@Zverik Done! :)

trashkalmar:
@R3gi Thanks! We'll create a pull request with your modifications!

trashkalmar:
@R3gi Could you please send the file to alexm@mapswithme.com? Looks like Bitster.cz is inaccessible outside Czech Republic.

R3gi:
@trashkalmar Thank you for this info. I am very sorry for these complications, GitHub seems to be still very confusing for me…they really should do something with it and make it more friendly for newbies too O:-)

PR #168, opened 2015-10-08
[turns][sound] Using for turn notification a real distance to the closest turn instead of a distance based on speed.

bykoianko:
При проигрывании поворота правильно использовать расстояние оставшееся до поворота, а не расстояние, которое вычисляется из скорости и времени активации оповещения.
Бывают случаи (после перестроения маршрута), когда эти расстояния могут сильно различаться.

PR помечен DO NOT MERGE, поскольку я жду, когда в мастер попадет https://github.com/mapsme/omim/pull/164 и я смогу доделать юнит тесты без конфликтов.

https://trello.com/c/xTfj3uHV/1898--

### DIFF
--- a/routing/routing_tests/turns_sound_test.cpp
+++ b/routing/routing_tests/turns_sound_test.cpp
@@ -184,7 +184,7 @@ UNIT_TEST(TurnsSoundMetersTwoTurnsTest)
   string const engShortJson =
       "\
       {\
-      \"in_700_meters\":\"In 700 meters.\",\
+      \"in_600_meters\":\"In 600 meters.\",\
       \"make_a_sharp_right_turn\":\"Make a sharp right turn.\",\
       \"enter_the_roundabout\":\"Enter the roundabout.\"\
       }";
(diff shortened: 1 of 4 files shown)


mapsmetest:

routing/turns_sound_settings.cpp
```diff
@@ -73,7 +72,23 @@ double Settings::ConvertUnitsToMeters(double distanceInUnits) const
...
-  // m_lengthUnits is equal to LengthUnits::Undefined or to unknown value.
+  ASSERT(false, ());
+  return 0.;
+}
+
+double Settings::ConvertMetersToUnits(double distanceInMeters) const
+{
+  switch (m_lengthUnits)
+  {
+    case LengthUnits::Undefined:
```
deathbaba: Do we really need this Undefined state in the code? If it is not used, it's better to stick to some default value (meters?) or always get it from our settings::

deathbaba:
LGTM with minor comment.

routing/turns_sound_settings.cpp
bykoianko: Yes, you're right. I needed Undefined before but not now. 

I guess we need:
1. to remove Undefined form LengthUnits;
2. to move LengthUnits to platform/settings.hpp and to remove Unit enum there;
3. Use enum class LengthUnits instead of enum Unit;

But I think it's a question of separate PR.

deathbaba: Is it hard to refactor here?
Is it the only place where Undefined is used?

bykoianko: > Is it hard to refactor here?

I guess it'll be a PR with one commit of about 100 lines(addition/deletion). 

> Is it the only place where Undefined is used?

No. Of cause not. There are about 10 places where LengthUnits::Undefined is used. They shall be removed. 
Besides that there are some lines which shall be exchanged. Units::Metric -> LengthUnits::Meters and so on.

I guess we should do it but in a separate PR.

routing/routing_tests/turns_sound_test.cpp
```diff
@@ -203,9 +203,13 @@ UNIT_TEST(TurnsSoundMetersTwoTurnsTest)
   TEST(turnNotifications.empty(), ());
 
   // 700 meters till the turn. It's time to pronounce the first voice notification.
```
kshalnev: 600 meters till the turn, I guess.

kshalnev:
LGTM

PR #182, opened 2015-10-09
[search] Added test for RankTable loading from opposite endianness.

ygorshenin:
https://trello.com/b/z0fs5uJ2/search-data

### DIFF
--- a/indexer/rank_table.cpp
+++ b/indexer/rank_table.cpp
@@ -128,16 +128,19 @@ public:
   uint8_t Get(uint64_t i) const override { return m_coding.Get(i); }
   uint64_t Size() const override { return m_coding.Size(); }
   RankTable::Version GetVersion() const override { return V0; }
-  void Serialize(Writer & writer) override
+  void Serialize(Writer & writer, bool preserveHostEndianness) override
   {
     static uint64_t const padding = 0;
 
     uint8_t const version = GetVersion();
-    uint8_t const flags = IsBigEndian();
+    uint8_t const flags = preserveHostEndianness ? IsBigEndian() : !IsBigEndian();
     writer.Write(&version, sizeof(version));
     writer.Write(&flags, sizeof(flags));
     writer.Write(&padding, 6);
-    Freeze(m_coding, writer, "SimpleDenseCoding");
+    if (preserveHostEndianness)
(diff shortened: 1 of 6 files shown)


mapsmetest:

deathbaba:
LGTM, but not critical - all our devices are in the same endiannes now.

indexer/rank_table.cpp
```diff
@@ -199,7 +202,7 @@ void SerializeRankTable(RankTable & table, FilesContainerW & wcont)
   vector<char> buffer;
   {
     MemWriter<decltype(buffer)> writer(buffer);
-    table.Serialize(writer);
+    table.Serialize(writer, true /* hostEndianness */);
```
syershov: не поправлен комментарий

indexer/rank_table.hpp
```diff
@@ -56,8 +56,11 @@ class RankTable
   // Returns underlying data format version.
   virtual Version GetVersion() const = 0;
 
-  // Serializes rank table.
-  virtual void Serialize(Writer & writer) = 0;
+  // Serializes rank table. When |preserveHostEndianness| is true,
+  // table is serialized in host endianness, otherwise, opposite
+  // endianness is used. Please, don't set this parameter to false
+  // unless you know what you do.
+  virtual void Serialize(Writer & writer, bool preserveHostEndianness) = 0;
```
deathbaba: Why not use something like ```enum class Endianness; virtual void Serialize(Writer & writer, Endianness e = ELittleEndian) = 0; ``` ?

ygorshenin: Because we're not interested here in big/little endianness - we're interested in preserving host machine's endianness or not. It can be Little Endian or Big Endian - it doesn't matter. So one should read this method as "serialize this for the same architectures as the host is or serialize this for foreign architectures." Explicit endianness will cause difficulties in using this method as clients should always remember that we're Little Endian only. Moreover, this unit is designed in a such way that it's even possible to deserialize this data structure from foreign endianness, thus, the main reason to introduce the last parameter was for testing purposes.

indexer/rank_table.cpp
ygorshenin: Done

ygorshenin:
@syershov PTAL

syershov:
LGTM

PR #209, opened 2015-10-12
Release routing workaround.

gardster:


mapsmetest:

deathbaba:
What about future releases? Will they work correctly? Or after releasing this workaround and next "correct" release we need to support both work-around and "correct" implementations?

deathbaba:
What is exact problem this PR solves?

gardster:
Yes they will. It solves routing problems (and fixes several routing_integration_tests by the way).

PR #236, opened 2015-10-14
[ios] Added float equality comparison warning.

igrechuhin:


mapsmetest:

iphone/Maps/Classes/Common.h
```diff
@@ -89,3 +89,9 @@ static inline NSString * bcp47ToTwineLanguage(NSString const * bcp47LangName)
   // Taking two first symbols of a language name. For example ru-RU -> ru
   return [bcp47LangName substringToIndex:2];
 }
+
+// Use only for screen dimensions CGFloat comparison
+static inline BOOL fequal(CGFloat left, CGFloat right)
+{
+  return fabs(left - right) < 0.5;
+}
```
VladiMihaylenko: нужен перенос строки

ygorshenin: Мб отразить факт что эта функция будет использоваться только для сравнения размерностей в ее названии? equalScreenDimensions?

VladiMihaylenko: имхо, сложно :)

VladiMihaylenko:
LGTM

igrechuhin:
@deathbaba PTAL & merge

iphone/Maps/Classes/Common.h
ygorshenin: Ке? Тогда велик шанс что fequal будет переиспользован и для обычных чисел с плавающей точкой, для которых точность в 0.5 неприемлима. Что сложно? find-and-replace? :)

ygorshenin:
LGTM

iphone/Maps/Classes/CustomViews/MapViewControls/ZoomButtons/MWMZoomButtonsView.mm
```diff
@@ -118,7 +119,8 @@ - (CGFloat)topBound
 
 - (void)setBottomBound:(CGFloat)bottomBound
 {
-  if (_bottomBound == bottomBound)
+
```
deathbaba: extra \n

deathbaba:
Are there real bugs because of float comparison, or a "let's add extra warning key and fix everything, even if it's not necessarily needed" decision? It can lead to errors only if floats come from different sources/after math computations. If you always get screen bounds from system/resources it should be safe to compare floats for equality.
You can also consider using base/math.hpp AlmostEqualULPs() which was designed for efficient floats comparison.
LGTM if it is really needed to fix a bug, and it is better to remove -Wfloat-equal otherwise.

igrechuhin:
In UI due to animations we can get fractional values, however the only valuable change is the one greater than 0.5
Excess UI updated are not required and are unwanted.
Apple recommends UI coordinates to be without fractional part.
I believe we should enable this warning.

deathbaba:
If there are animation computations - I agree, it should be enabled on iOS code level.

PR #373, opened 2015-10-23
[search] Revert search logic.

vng:
In table + map view algorithm is working with the same logic as in table view.


mapsmetest:

PR #384, opened 2015-10-26
Moving downloader_tests into a separate executable.

therearesomewhocallmetim:
Separated the downloader tests from the platform tests to decrease the time during which the test server needs to run. Also, made all the tests that require a server to run, execute in the same bunch, for the same purpose. All that is needed to make the tests pass (and not produce false negatives).


therearesomewhocallmetim:
[skip ci]

therearesomewhocallmetim:
JTTP

therearesomewhocallmetim:
By the way, the tests failed because of Konstantin's tests, the ones that I tried to fix, passed.

PR #526, opened 2015-11-12
Removed expat include dependency which caused build error on linux for style_tests.

deathbaba:
This fix should be done 5 years ago when we imported expat.
It also reduces qmake dependencies check time and makes C++ compilation slightly faster.


Zverik:
LGTM

PR #535, opened 2015-11-13
[platform] Fixed logging severity for LocalCountryFileUtils.

ygorshenin:
LERROR is too serious and causes crashes for debug builds. It should be replaced by LWARNING.


rokuz:
LGTM

PR #560, opened 2015-11-17
[DO NOT MERGE] [base] Get back to boost regexp.

ygorshenin:
Ubuntu 14.04 has a serious bug in a libc++ which prevents us from usage of this library. This bug was fixed in a later versions, so, if our testing machine will be upgraded to modern versions of Ubuntu the bug will go away as libc++ will be updated. But, if this will not happen, the only possibility is to get back to Ubuntu 14.04 libstdc++, which does not support std::regex and other fancy features.

To sum up: in case of successful Jenkins linux bot upgrade this CL must be closed. Otherwise, it must be merged.


ygorshenin:
Closing the PR as bots happy now.

deathbaba:
What about libstdc++? Does it work with it?

ygorshenin:
Could you please paraphrase your question? It's unclear what are you talking about.

deathbaba:
Does your code compile on ubuntu with -spec linux-clang ?

ygorshenin:
Sorry, man. I can't check our project on all ubuntus and all libstdc++'s, especially on those that don't support c++11.

ygorshenin:
If you have enough resources for doing that, please, share results with us.

deathbaba:
I don't understand your reply.
My question is simple: does it work on the same ubuntu 14.04 but with libstdc++ instead of libc++? Our code compiles well without libc++ (only recent update of opening_hours broke it).

ygorshenin:
Our code does not compile with libstdc++s not supporting C++11. Our code compiles with libstdc++s supporting C++11.

deathbaba:
Of course, libstdc++ supports c++11 too. Why do you point to C++11? I just asked if you tried with another std library, not about turning off C++11.

ygorshenin:
Your question was " What about libstdc++? Does it work with it?", and I
gave you exact reply. If you think that I didn't, don't hesitate to reread
my previous messages instead of writing new ones.

deathbaba:
> Ubuntu 14.04 has a serious bug in a libc++ which prevents us from usage of this library. This bug was fixed in a later versions, so, if our testing machine will be upgraded to modern versions of Ubuntu the bug will go away as libc++ will be updated. But, if this will not happen, the only possibility is to get back to Ubuntu 14.04 libstdc++, which does not support std::regex and other fancy features.

Му question was "have you tried to compile your code in exactly the same environment, but without using libc++, which has a bug, and using libstdc++ instead? With exactly the same C++11 options, which are not related at all to the standard library used (e.g. remove -stdlib=libc++ option from command line).

Issue #559, opened 2015-11-17
Windows 7 build & apk exception error


stevegdavis:
I have managed to do a Windows 7 build (assembleWebDebug) but the resultant 'android-web-universal-debug-5.1.5-151117.apk' crashes with the usual 'Unfortunately, maps.me debug has stopped' when run on my HTC One device (copied via USB cable).

I had numerous lint errors on build to do with the language translation but suppressed those with 'disable MissingTranslation' in the build.gradle file so I am assuming that is not the reason it is crashing.

Looking at the logcat I see:-
11-17 10:35:46.743 E/AndroidRuntime(10725): FATAL EXCEPTION: main

11-17 10:35:46.743 E/AndroidRuntime(10725): java.lang.ExceptionInInitializerError

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at com.mapswithme.maps.MwmActivity.onCreate(MwmActivity.java:286)

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at android.app.Activity.performCreate(Activity.java:5165)

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1103)

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2419)

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2520)

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at android.app.ActivityThread.access$600(ActivityThread.java:162)

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1366)

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at android.os.Handler.dispatchMessage(Handler.java:99)

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at android.os.Looper.loop(Looper.java:158)

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at android.app.ActivityThread.main(ActivityThread.java:5751)

11-17 10:35:46.743 E/AndroidRuntime(10725): 	at java.lang.reflect.Method.invokeNative(Native Method)

[...]

PR #568, opened 2015-11-17
Minor faq & configure script fixes.

deathbaba:
Fixes issue with a crash mentioned in #559, when Flurry key is empty (internal Flurry library check for empty string).


vng:
LGTM

bykoianko:
LGTM

PR #569, opened 2015-11-17
Update strings.txt

freehorn3000:
Add language Persian

### DIFF
--- a/strings.txt
+++ b/strings.txt
@@ -30,6 +30,8 @@
     ro = Despre
     nb = Om
     fi = Tietoja
+    fr=درباره
+    
 
   [back]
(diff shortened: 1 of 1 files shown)


trashkalmar:
Sorry, we can not accept your PR, because it:
* uses 'fr' qualifier which corresponds to French but not to declared Persian;
* does not cover all string sections;
* has syntax errors.

deathbaba:
@freehorn3000 Please provide translations for all strings in a file, and use 'fa' language code, not fr.

freehorn3000:
Hi thanks for help me 
Ok I translations for all strings

PR #585, opened 2015-11-18
Strings for P2P and What's New

therearesomewhocallmetim:


strings.txt
```diff
@@ -14412,3 +14412,410 @@
...
+    ja = 目的地
+    ro = În
+    ar = إلى
+    sk = Do
+    es = A
+    pl = Do
+    nl = Naar
+    pt = Para
+
+  [p2p_only_from_current]
```
therearesomewhocallmetim: To Alex: you can start the follow mode only if you navigate from the current location, otherwise there is no point in navigating if you are miles away from the actual route.

strings.txt
```diff
@@ -14412,3 +14412,410 @@
...
+    ja = 現在位置からのルートを作成しますか？
+    ro = Doriți să vă planificăm o rută având ca punct de pornire locația actuală?
+    ar = هل ترغب في أن نرسم مسار لك من موقعك الحالي؟
+    sk = Doriți să vă planificăm o rută având ca punct de pornire locația actuală?
+    es = ¿Quieres que planeemos un ruta desde tu ubicación actual?
+    pl = Czy chcesz, byśmy zaplanowali trasę z Twojej bieżącej lokalizacji?
+    nl = Wilt u dat wij een route plannen vanaf uw huidige locatie?
+    pt = Deseja planejar uma rota a partir da sua localização atual?
+
+  [whats_new_where_to_turn]
```
therearesomewhocallmetim: There are 2 screens in the "what's new section", each contains a header and the body, thus we have 4 strings.

VladiMihaylenko:
LGTM

strings.txt
deathbaba: Do not understand it anyway. How is follow mode connected to p2p? May be designer screenshots make it clear.

PR #589, opened 2015-11-19
Run classificator tests for every style

kshalnev:
cherry-pick from Drape


kshalnev:
PTAL @vng @bykoianko

bykoianko:
LGTM

PR #604, opened 2015-11-19
Fixed drape qmake dependencies.

deathbaba:
It should fix build errors on Drape.
Technically, libraries do not depend on each other and can be built in parallel. Dependencies are needed only when libraries are linked to any executable.


Zverik:
Made a build successfully via command line (gradlew assembleWebDebug).

PR #625, opened 2015-11-20
[ios] Fixed bug with direction view after 180 degree rotation.

VladiMihaylenko:
https://trello.com/c/OG1Th4Bi/2057-ios-180


VladiMihaylenko:
@rokuz ptal

rokuz:
LGTM

PR #653, opened 2015-11-22
take QMAKE env variable.

easz:
updated script so that it will take QMAKE which is eventually defined as env variable


deathbaba:
Just for the reference about shell ${var:-} syntax: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

PR #657, opened 2015-11-23
[ios] Added checking for location state mode while routing is planning.

VladiMihaylenko:
@therearesomewhocallmetim @igrechuhin ptal


igrechuhin:
LGTM

Issue #654, opened 2015-11-23
Path planning can not be automatically corrected


xingdavis:
In the navigation process, when the traveling path shift, can not automatically correct. Is there any way to change it?

deathbaba:
Looks like a bug. Can you please provide more details (and probably screenshot)? Is it android/ios? Latest version from store?

xingdavis:
IOS,Is the latest version. I'll make a test again, and then send a screenshot

xingdavis:
The same with this issue #721

deathbaba:
Looks like it was already fixed on a master branch. Have you re-checked it?

xingdavis:
@deathbaba It's ok now.

PR #683, opened 2015-11-24
Nomem routing fix.

gardster:
https://github.com/mapsme/omim/issues/654
We will not unmap the routing facade for the current map while routing is active. iOS7 often returns NOMEM error when we are trying to map the fano matrix section.

### DIFF
--- a/coding/file_container.cpp
+++ b/coding/file_container.cpp
@@ -4,6 +4,7 @@
 #include "coding/internal/file_data.hpp"
 
 #ifndef OMIM_OS_WINDOWS
+  #include <errno.h>
   #include <unistd.h>
   #include <sys/mman.h>
   #include <sys/stat.h>
@@ -169,7 +170,7 @@ MappedFile::Handle MappedFile::Map(uint64_t offset, uint64_t size, string const 
 #else
   void * pMap = mmap(0, length, PROT_READ, MAP_SHARED, m_fd, alignedOffset);
   if (pMap == MAP_FAILED)
-    MYTHROW(Reader::OpenException, ("Can't map section:", tag, "with [offset, size]:", offset, size, "errno:", errno));
+    MYTHROW(Reader::OpenException, ("Can't map section:", tag, "with [offset, size]:", offset, size, "errno:", strerror(errno)));
 #endif
 
   char const * data = reinterpret_cast<char const *>(pMap);
--- a/coding/file_container.cpp
+++ b/coding/file_container.cpp
@@ -169,7 +169,7 @@ MappedFile::Handle MappedFile::Map(uint64_t offset, uint64_t size, string const 
 #else
   void * pMap = mmap(0, length, PROT_READ, MAP_SHARED, m_fd, alignedOffset);
   if (pMap == MAP_FAILED)
-    MYTHROW(Reader::OpenException, ("Can't map section:", tag, "with [offset, size]:", offset, size));
+    MYTHROW(Reader::OpenException, ("Can't map section:", tag, "with [offset, size]:", offset, size, "errno:", errno));
 #endif
 
   char const * data = reinterpret_cast<char const *>(pMap);
--- a/coding/file_container.cpp
+++ b/coding/file_container.cpp
@@ -5,6 +5,7 @@
 
 #ifndef OMIM_OS_WINDOWS
   #include <errno.h>
+  #include <stdio.h>
   #include <unistd.h>
   #include <sys/mman.h>
   #include <sys/stat.h>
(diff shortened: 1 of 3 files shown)


gardster:
DUDE
JTTP

coding/file_container.cpp
```diff
@@ -169,7 +169,7 @@ MappedFile::Handle MappedFile::Map(uint64_t offset, uint64_t size, string const
 #else
   void * pMap = mmap(0, length, PROT_READ, MAP_SHARED, m_fd, alignedOffset);
   if (pMap == MAP_FAILED)
-    MYTHROW(Reader::OpenException, ("Can't map section:", tag, "with [offset, size]:", offset, size));
+    MYTHROW(Reader::OpenException, ("Can't map section:", tag, "with [offset, size]:", offset, size, "errno:", errno));
```
ygorshenin: Please, use strerror(errno) and include all necessary headers.

kshalnev:
LGTM

coding/file_container.cpp
gardster: Done.

gardster:
DUDE
JTTP

coding/file_container.cpp
```diff
@@ -169,7 +170,7 @@ MappedFile::Handle MappedFile::Map(uint64_t offset, uint64_t size, string const
 #else
   void * pMap = mmap(0, length, PROT_READ, MAP_SHARED, m_fd, alignedOffset);
   if (pMap == MAP_FAILED)
-    MYTHROW(Reader::OpenException, ("Can't map section:", tag, "with [offset, size]:", offset, size));
+    MYTHROW(Reader::OpenException, ("Can't map section:", tag, "with [offset, size]:", offset, size, "errno:", strerror(errno)));
```
ygorshenin: And include please stdio.h for strerror.

routing/osrm_data_facade.hpp
```diff
@@ -272,6 +272,10 @@ template <class EdgeDataT> class OsrmDataFacade : public OsrmRawDataFacade<EdgeD
   {
     Clear();
 
+    // Load huge data first.
```
ygorshenin: How about to clarify why we map largest file first:
```c++
// Map huge data first, as we hope it will reduce fragmentation of the program address space.
```

ygorshenin:
LGTM after fixes.

mapsmetest:
JTTP, DUDE

therearesomewhocallmetim:
DUDE, JTTP

coding/file_container.cpp
gardster: DONE

therearesomewhocallmetim:
DUDE, JTTP

coding/file_container.cpp
kshalnev: I think it would be good to print errno too.
If strerror returns nullptr, will it work?

ygorshenin: As I know strerror never returns a NULL, moreover, numeric values of system errors may differ on different platforms.

ygorshenin: In case of unknown errno values, strerror returns "Unknown error" or something like that.

kshalnev: wait... i agree that text is dependent of platform/library, but errno value... what about macro like EIO, ENODIR (http://www.cplusplus.com/reference/system_error/errc/). I guess their values were standardized.

ygorshenin: Sorry, maybe I missed something, but integral values of such things as ENOMEM, EINVAL, ENOFILE, etc. are not specified (there is only explicit specification for  symbolic values of some subset of possible system errors, http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html). For example, on my MacOS, EDQUOT has integral value 69, while on Linux it's value is 122. That's why I do not understand why do you insist on logging errno as integer.

By the way, could you please describe a situation when you need an integral value of errno?

kshalnev: ok. thanks for info. then i agree with you.

LGTM

gardster:
DUDE, JTTP

kshalnev:
DUDE, JTTP

syershov:
DUDE
JTTP

gardster:
DUDE
JTTP

gardster:
DUDE, JTTP

syershov:
DUDE
JTTP

therearesomewhocallmetim:
DUDE 
JTTP

therearesomewhocallmetim:
DUDE JTTP

syershov:
DUDE
JTTP

syershov:
DUDE
JTTP

therearesomewhocallmetim:
DUDE JTTP

therearesomewhocallmetim:
DUDE JTTP

PR #711, opened 2015-11-25
[android] fix: Crash on device rotate while pedestrian navigation is active.

trashkalmar:
…active.


trashkalmar:
https://trello.com/c/G5UB1FRu/2098-android

PR #727, opened 2015-11-26
[android] fix: Next turn was incorrectly shown during pedestrian navigation.

trashkalmar:
…gation.


trashkalmar:
https://trello.com/c/yV0B5Tu0/2105-android

PR #749, opened 2015-11-27
[android] fix: Custom title for reroute suggest dialog.

trashkalmar:


trashkalmar:
https://trello.com/c/W2kPIUr1/2110-android

PR #760, opened 2015-11-30
[drape] Drape 3d, path text rendering without perspective deformation.

darina:
* Path text rendering without perspective deformation.
* Fixed rendering of position accuracy and position mark.

### DIFF
--- a/drape_frontend/path_text_shape.cpp
+++ b/drape_frontend/path_text_shape.cpp
@@ -1,7 +1,6 @@
 #include "drape_frontend/path_text_shape.hpp"
 #include "drape_frontend/text_handle.hpp"
 #include "drape_frontend/text_layout.hpp"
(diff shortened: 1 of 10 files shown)


drape_frontend/path_text_shape.cpp
```diff
@@ -48,17 +55,66 @@ class PathTextHandle : public df::TextHandle
     if (!df::TextHandle::Update(screen))
       return false;
 
-    return m_layout->CacheDynamicGeometry(m_centerPointIter, screen, m_normals);
+    m2::Spline pixelSpline(m_spline->GetSize());
+
+    vector<m2::PointD> const & globalPoints = m_spline->GetPath();
+    for (auto pos : globalPoints)
```
rokuz: auto const & pos

rokuz:
LGTM

PR #773, opened 2015-11-30
[android] fix: Restore slots state after device rotated.

trashkalmar:


trashkalmar:
https://trello.com/c/Shycln2q/2106-android-point-to-point-ios

PR #839, opened 2015-12-03
[android] Fixed bug with routing not being built.

yunikkk:
https://trello.com/c/g9V2sIYF/2133-android-2-don-t-keep-activities


yunikkk:
@trashkalmar PTAL

PR #868, opened 2015-12-07
Update INSTALL.md: clang 3.6

stangls:
It seems, a bug was fixed in llvm leading to a segmentation fault on some systems.
Upgrading to Version 3.6 (currently latest stable according to http://llvm.org/apt/) helped working around this issue.


deathbaba:
Which exactly issue are you referencing to?
Have you checked the most recent master branch with clang-3.5?

stangls:
Yes, I have checked out the master-branch with clang-3.5 which lead to segmentation faults during build on my Ubuntu 14.04 (the given documentation on how to install was followed 1:1).
At first, this problem occurred on the drape-2.0 branch (which I am using for evaluation purposes), so I switched back to the master branch to rule out any development-issues, but the problem still occurred.

I'm sorry If you require a ticket/issue first, I will create one.
There exist several issues with LLVM and clang concerning segmentation faults. I did not go into detail on how the segmentation fault occurred as upgrading to the newest stable fixed the problem for some cases and eventually also for me. I do not see a problem with upgrading to the next stable release in general as the build seems to be okay.
If there are any other hints on how to avoid segmentation faults during build I will happily try them with clang-3.5.

deathbaba:
Was it a debug or release build when segmentation fault occurred?

It would be very useful to know which command line and source caused segmentation fault. There can be an error in environment setup documentation for example.

stangls:
I did a release build with the command ```CONFIG=gtool tools/unix/build_omim.sh -cro``` on master-branch and then after upgrading to 3.6 with ```CONFIG=gtool tools/unix/build_omim.sh -cro```.
I can reproduce the problem if you need further details.

I seem to understand your concern, there may in fact be differences when using a different version of clang, as for example the switch to GNU C11, which may be avoided using the flag -std=gnu99, as depicted on http://llvm.org/releases/3.6.0/tools/clang/docs/ReleaseNotes.html#c-language-changes-in-clang

I am not able to run all tests with satisfaction, but I do see other problems there, for example opening-count.lst is missing for OpeningHoursSupportedFeatures and online_cross_tests.cpp::OnlineRussiaNorthToSouthTest using HTTP-Requests with invalid URL '/mapsme?loc=61.76,34.45&loc=45.07,38.94' (could be due an empty repository-URL?)...

deathbaba:
Have you tried a debug build?
Can you please share exact command line which causes crash?

Invalid URL is due to empty private.h file - we do not publish our functional test server for security reasons (but you can always run your own one).

@mgsergio Can we put opening-count.lst into the repo? Or may be zip it (and unzip into a memory via our ZipReader)?

stangls:
Host: Ubuntu 14.04.3 LTS
Clang version: 3.5.0-4ubuntu2~trusty2
Clang repository: http://ppa.launchpad.net/jhe/llvm-toolchain/ubuntu
Branch: master
Git status: clean (= no changes and no unversioned files)
Latest git commit: 650c1e5aa

Build type: Debug build
Command: tools/unix/build_omim.sh -cdro 2>&1 | tee build.out

Please see attached file for full output ( search for first occurrance"route.o" )
[build.out.zip](https://github.com/mapsme/omim/files/55683/build.out.zip)

stangls:
Retried with newest master-branch and non-german environment (via ```export LC_ALL=C```)

Latest git commit: 3badc404e2e
... (everything else the same)

Same error, different (english) output: [build2.out.zip](https://github.com/mapsme/omim/files/55704/build2.out.zip)

deathbaba:
Thanks for your build logs! The crash was caused by 3party/sdf_image/sdf_image.cpp source (probably by base/buffer_vector.hpp implementation) http://pastebin.com/EDPGFyuH

After we have merged drape-2.0 branch into master it does not compile with clang 3.5.2 too due to other errors (compiler can't cast non-const member function pointer to a const one, like this: http://pastebin.com/iX5CGjfH ).

So you are right, and only clang 3.6+ can now be used.

Thanks again for your time and contribution!

PR #900, opened 2015-12-09
fix generate_mwm

stangls:
It seems that ```generate_mwm.sh``` is unable to compile ```osmconvert``` with GNU C Compiler 4.8.4 on Ubuntu 14.04 if -lz is provided as an early argument. It fails with the error message ```undefined reference to `gzopen'```

~~Also I added boost-libraries to the docs. This is not ultimately neccessary as the build fails with appropriate messages if the libraries are unavailable and can be reverted if seen fit.~~

### DIFF
--- a/docs/MAPS.md
+++ b/docs/MAPS.md
@@ -25,8 +25,6 @@ Specify `TARGET` variable for changing that (e.g. `TARGET=.`). The script runs `
 see `find_generator_tool.sh` script for an algorithm on how it finds it. All temporary files are created
 with `mktemp` and then removed.
 
-If you do not have osmconvert installed in ```$HOME/osmctools/osmconvert```, the script may compile it from source. In this case you need the following boost libraries: Thread, System, Program Options, Filesystem, Date Time, Atomic, Chrono, Serialization. On Ubuntu you may want to install them using ```apt-get install libboost-thread-dev libboost-system-dev libboost-program-options-dev libboost-filesystem-dev libboost-date-time-dev```
-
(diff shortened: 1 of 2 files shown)


stangls:
There is also an official documentation about the position of the linker parameter ```-lz``` on https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html :
> It makes a difference where in the command you write this option; the linker searches and processes libraries and object files in the order they are specified. Thus, ‘foo.o -lz bar.o’ searches library ‘z’ after file foo.o but before bar.o. If bar.o refers to functions in ‘z’, those functions may not be loaded.

docs/MAPS.md
```diff
@@ -25,6 +25,8 @@ Specify `TARGET` variable for changing that (e.g. `TARGET=.`). The script runs `
 see `find_generator_tool.sh` script for an algorithm on how it finds it. All temporary files are created
 with `mktemp` and then removed.
 
+If you do not have osmconvert installed in ```$HOME/osmctools/osmconvert```, the script may compile it from source. In this case you need the following boost libraries: Thread, System, Program Options, Filesystem, Date Time, Atomic, Chrono, Serialization. On Ubuntu you may want to install them using ```apt-get install libboost-thread-dev libboost-system-dev libboost-program-options-dev libboost-filesystem-dev libboost-date-time-dev```
```
deathbaba: I think boost libraries are not needed for osmconvert. They are needed for osrm project to build maps routing.

stangls: You are correct. I removed the docs from this pull request.

PR #972, opened 2015-12-14
[search] Numerous fixes to Geocoder (search query v2).

ygorshenin:
This CL includes following fixes:
* fix to house numbers matching --- all tokens from query must be matched in feature.
* fix to lookup radius - now it's shared between ReverseGeocoder and Geocoder.
* fix to matching algorithm - optimized layers intersection, now DFS is incorporated into layers intersection.
* fixed architecture - stack of layers is constant now and can't be modified by path finder or matcher.

### DIFF
--- a/search/reverse_geocoder.cpp
+++ b/search/reverse_geocoder.cpp
@@ -13,7 +13,6 @@ namespace search
 {
 namespace
 {
-double constexpr kLookupRadiusM = 500.0;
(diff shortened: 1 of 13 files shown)


ygorshenin:
@vng PTAL

ygorshenin:
@mpimenov PTAL

mpimenov:
LGTM

PR #1036, opened 2015-12-17
[search] Enable SearchQueryV2 by default.

ygorshenin:
Usage of USE_SEARCH_QUERY_V2 macro makes it hard to build new-search version for platforms. I think it would be nice to enable it by default, as it lives on a separate branch.


ygorshenin:
@vng PTAL

PR #1147, opened 2015-12-25
Restored qmake call for all Android and iOS builds to fix build errors when switching branches in git.

deathbaba:
Typical error appears when some new source files were added on a one branch and are missing on another one. Another possible fix would be to create an unique build folder for each branch to avoid their sharing, but it's a bigger overhead.

We had correct implementation for years but it was changed around a year ago to make builds faster.

I prefer to have slower but stable builds than faster but unstable.


bykoianko:
LGTM

PR #1150, opened 2015-12-26
Adding two small islands missed in Taiwan.poly

kcwu:
https://www.openstreetmap.org/way/9216199 東碇島
https://www.openstreetmap.org/relation/3339702 烏坵鄉

PR #1155, opened 2015-12-28
[borders] Add missing islands to Taiwan, and remove from China

Zverik:
Fixed the border myself, instead of #1150.


vng:
polygons.bin ?

Zverik:
Собрал.

PR #1223, opened 2016-01-05
[search][tmp] Show address in PP. Need to recheck PP layout for the production.

vng:


ygorshenin:
LGTM

map/address_finder.cpp
```diff
@@ -1,6 +1,7 @@
 #include "map/framework.hpp"
 
 #include "search/categories_holder.hpp"
+#include "search/reverse_geocoder.hpp"
```
mpimenov: sort

mpimenov:
LGTM

vng:
PTAL

mpimenov:
LGTM

PR #1234, opened 2016-01-05
Removing dependency. Storage does not depend on tests now.

bykoianko:
https://trello.com/c/QcIMhU0f/2232-storage-testing

### DIFF
--- a/storage/storage_tests/storage_tests.cpp
+++ b/storage/storage_tests/storage_tests.cpp
@@ -845,7 +845,7 @@ UNIT_TEST(StorageTest_GetRootId)
                            "id": "Countries",
                            "v": 151227,
                            "g": []
-                         })"));
+                         })"), new TestMapFilesDownloader());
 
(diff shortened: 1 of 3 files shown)


storage/storage_tests/storage_tests.cpp
```diff
@@ -845,7 +845,7 @@ UNIT_TEST(StorageTest_GetRootId)
                            "id": "Countries",
                            "v": 151227,
                            "g": []
-                         })"));
+                         })"), new TestMapFilesDownloader());
```
ygorshenin: Memory leak here, right?

ygorshenin: OK, no leak, but in any case, please, pass test downloader as unique_ptr, not as raw pointer.

bykoianko: Sure. I'm working on a new PR.

bykoianko: https://github.com/mapsme/omim/pull/1246

PR #1244, opened 2016-01-06
Added sending statistics about GPU

rokuz:


rokuz:
@deathbaba PTAL

drape/support_manager.cpp
```diff
@@ -12,6 +14,15 @@ void SupportManager::Init()
   string const version = GLFunctions::glGetString(gl_const::GLVersion);
   LOG(LINFO, ("Renderer =", renderer, "Version =", version));
 
+  // On Android the engine may be recreated. Here we guarantee that GPU info is sent once.
+  static bool gpuInfoSent = false;
+  if (!gpuInfoSent)
+  {
+    alohalytics::TStringMap details {{ "renderer", renderer }};
+    alohalytics::Stats::Instance().LogEvent("GPU", details);
```
deathbaba: You don't need <string, map> event here. LogEvent("GPU renderer", renderer) is enough.
Btw, GLFunctions::glGetString() never returns nullptr, right?

rokuz: Да, nullptr не будет

rokuz:
@deathbaba PTAL

PR #1364, opened 2016-01-13
[search] Correct check for RankTable validity.

deathbaba:
Fixes assert in Debug with old MWM files.

### DIFF
--- a/indexer/rank_table.cpp
+++ b/indexer/rank_table.cpp
@@ -221,7 +221,10 @@ void SerializeRankTable(RankTable & table, string const & mapPath)
 template <typename TRegion>
 unique_ptr<RankTable> LoadRankTable(unique_ptr<TRegion> && region)
 {
-  if (!region || !region->ImmutableData() || region->Size() < kHeaderSize)
+  if (!region || !region->ImmutableData())
+    return unique_ptr<RankTable>();
+
+  if (region->Size() < kHeaderSize)
   {
     LOG(LERROR, ("Invalid RankTable format."));
     return unique_ptr<RankTable>();


vng:
LGTM

indexer/rank_table.cpp
```diff
@@ -221,7 +221,11 @@ void SerializeRankTable(RankTable & table, string const & mapPath)
 template <typename TRegion>
 unique_ptr<RankTable> LoadRankTable(unique_ptr<TRegion> && region)
 {
-  if (!region || !region->ImmutableData() || region->Size() < kHeaderSize)
+  if (!region || !region->ImmutableData())
+  {
+    return unique_ptr<RankTable>();
+  }
+  else if (region->Size() < kHeaderSize)
```
ygorshenin: no else after return, please.

ygorshenin:
What's the point in this change?

ygorshenin:
I mean, could y

ygorshenin:
ou please describe a case where old check fails?

indexer/rank_table.cpp
deathbaba: ?

deathbaba:
We just discussed it in Skype chat. Old code fails with old MWMs in debug when you try to search for anything.

deathbaba:
@ygorshenin PTAL please.

indexer/rank_table.cpp
ygorshenin: Please, rewrite the code like:
```c++
if (...)
  return make_unique<RankTable>();
if (...)
```

deathbaba: Already done!

ygorshenin:
LGTM

PR #1376, opened 2016-01-13
[downloader] Storage test fixes

syershov:
- Add storage tests in Xcode projects
- Add support path options into testing framework


bykoianko:
LGTM

PR #1387, opened 2016-01-14
[Downloader] Part implementation of method which gets closest countryId list by point.

bykoianko:
https://trello.com/c/TZtVVewj/2261-mwm


bykoianko:
@igrechuhin PTAL

PR #1422, opened 2016-01-15
F.A.Q. update

ForkWithMe:
Deleted Blue Stack player reference, since it doesn’t work with current version from Google Play


ForkWithMe:
PTAL

PR #1436, opened 2016-01-18
[map_download] Write dir changer

kshalnev:
WriteDirChanger has been moved to hpp/cpp to be reused in another tests


syershov:
LGTM

PR #1447, opened 2016-01-18
[android] fix: Do not show rating dialogs while in navigation.

trashkalmar:


trashkalmar:
https://trello.com/c/JZKR7dmV/2219--

PR #1486, opened 2016-01-20
[downloader] Remove test dir before and after test

kshalnev:


storage/storage_integration_tests/write_dir_changer.cpp
```diff
@@ -12,20 +12,9 @@ WritableDirChanger::WritableDirChanger(string testDir)
...
-  case Platform::ERR_FILE_ALREADY_EXISTS:
-    Platform::EFileType type;
-    TEST_EQUAL(platform.GetFileType(m_testDirFullPath, type), Platform::ERR_OK, ());
-    TEST_EQUAL(type, Platform::FILE_TYPE_DIRECTORY, ());
-    break;
-  default:
-    TEST(false, ("Can't create directory:", m_testDirFullPath));
-    break;
-  }
+  platform.RmDirRecursively(m_testDirFullPath);
```
ygorshenin: I guess it's not good to delete something you didn't created.

ygorshenin: What if, by mistake, during experiments, I'll put something important (like, big mwms) in a test dir, and this code will delete them?

PR #1580, opened 2016-01-24
Updated German localization

claudiush:
- More consistent glossary use
- Corrected punctuation
- Removed "Einkaufenladen", which is not a common German word

### DIFF
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -1249,7 +1249,7 @@ amenity-place_of_worship
 en:place of worship|4temple|sights|U+1F64F|U+26EA
 ru:3храм|церковь|достопримечательность
 uk:3храм|церква|собор|пам’ятка|алтарь|мавзолей|пам’ятні місця
-de:Kirche|Synagoge|Moschee|4Tempel|Anbetungsstätte|Dom|Sehenswürdigkeit
+de:Anbetungsstätte|Sehenswürdigkeit
 fr:4lieu de culte|site touristique
 it:4tempio|cattedrale|duomo|turistico
 es:4templo|lugar de culto|turismo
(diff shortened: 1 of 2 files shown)


data/categories.txt
```diff
@@ -1249,7 +1249,7 @@ amenity-place_of_worship
 en:place of worship|4temple|sights|U+1F64F|U+26EA
 ru:3храм|церковь|достопримечательность
 uk:3храм|церква|собор|пам’ятка|алтарь|мавзолей|пам’ятні місця
-de:4Tempel|Kirche|Dom|Sehenswürdigkeit
+de:Kirche|Synagoge|Moschee|4Tempel|Anbetungsstätte|Dom|Sehenswürdigkeit
```
deathbaba: This translation should have only generic words, not related to any specific religion.
We have recently introduced religion-specific types but still not translated them all at the moment.

data/categories.txt
```diff
@@ -1510,6 +1510,7 @@ amenity-bench
 en:bench
 ru:скамейка|лавка
 uk:лавка
+en:Parkbank
```
deathbaba: de: instead of en:

data/categories.txt
```diff
@@ -2814,7 +2815,7 @@ highway-primary|highway-primary_link|highway-residential|highway-secondary|highw
 en:street|st|str|ave|road|rd
 ru:улица|переулок|проезд|тракт|проспект|ул
 uk:вулиця|провулок|проїзд|проспект|бульвар|вул
-de:Strasse|Weg|Avenue|Gasse|str
+de:Strasse|Weg|Avenue|Gasse|Str
```
deathbaba: @vng @ygorshenin @mpimenov Does it fit well with our search_string_utils values? Should we store all synonyms in one place only instead of two?

claudiush:
Thanks for the review. I have incorporated the fixes you mentioned and also added the new separate category entries for religion specific place of worship POIs.

data/categories.txt
ygorshenin: There is some kind of duplication, you're right. Seems that we should parse somehow this file to create a list of synonyms and use it during search.

PR #1585, opened 2016-01-25
[strings] Fixed recent strings commit.

deathbaba:


deathbaba:
@claudiush Please take a look too.
In the future, any modification of strings.txt should be followed by launching tools/unix/generate_localizations.sh

PR #1620, opened 2016-01-26
[editor] Temporarily ignore edited features from old mwm files.

deathbaba:
Fixes possible crash for mwm migration case,


mgsergio:
LGTM

PR #1650, opened 2016-01-27
[search] Fixed tests.

mpimenov:
Migration to small MWMs conicided with removing separate
routing files (thus incorporating routing into regular map files).
This CL fixes the corresponding behaviour of search integration tests.

Also, fixed quality tests to account for a new World file.


ygorshenin:
LGTM

PR #1691, opened 2016-01-28
case-sensitive search on linux-systems

stangls:
It seems the compilation fails due to case-sensitivity under Linux.


syershov:
LGTM

PR #1728, opened 2016-01-29
[coding] A clean-up of the coding library.

mpimenov:


mpimenov:
@syershov PTAL

ygorshenin:
We do not use arithmetic coder right now, but it may be useful in the future. On the other hand, it's not a big deal to resurrect it from the git history.

LGTM

syershov:
LGTM

deathbaba:
There is a problem with resurrecting from a git history: someone in the team should know that codec existed in the past and was deleted.
What was wrong with this code? Why not simply remove it from .pro files but retain grepable/searchable online?

mpimenov:
It was not used anywhere, the tests were commented out because they did not pass, the code produced compiler warnings.
I tried to leave commit messages as greppable as possible.

PR #1741, opened 2016-02-01
Improving Dutch translations

sanderd17:
I have two feature requests for the translations:
* Split up the shops a bit more, and add some more types. Currently there are a lot of shops that just fall under "shop"
* Get rid of the religion translation. F.e. a Christian religious place can be anything from a wayside cross over a chapel to a cathedral. So I can't just put "church", as in a well-mapped area, most of the religious places will be the smaller wayside crosses and shrines. An existing proposal to solve this is by using the tag place_of_worship:type=* (see http://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dplace_of_worship#place_of_worship:type_.28Chapels.2C_Small_POWs.29 and http://taginfo.openstreetmap.org/keys/place_of_worship%3Atype#overview ). it would be nice if this could be supported instead of the religion tag. It's not yet very popular, but the tag is about the only one that is able to describe these different features, and a bit of tool support would improve its usage I think.

Thanks

### DIFF
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -68,7 +68,7 @@ ko:1환전|돈|교환을
 ja:1為替|マネー|お金|通貨|金|両替
 cs:3směnárna|peníze
 sk:3zmenáreň|peniaze
-nl:3geld wisselen|3wisselen|geld
+nl:3wisselkantoor|3geld wisselen|wisselen|geld
 zh-Hant:1匯率|理財|外幣|兌換|貨幣
 pl:4kantor|5wymiana|pieniądze
 pt:3câmbio|3troca|dinheiro
@@ -97,7 +97,7 @@ ko:바|술집|맥주|음료|음식
 ja:1居酒屋|バー|パブ|ビール|飲食|飲み屋|呑み屋|酒
 cs:2bar|pivo|pití|jídlo|voda|pitná voda|tekutiny
 sk:2bar|pivo|pitie|jedlo|voda|pitná voda|tekutiny
-nl:2bar|2pub|bier|drinken|eten
+nl:2bar|2pub|2café|bier|drinken|eten
 zh-Hant:1酒吧|酒館|飲酒|PUB|吧台|飲食
 pl:2bar|2pub|piwo|napoje|jedzenie
 pt:2bar|2pub|cerveja|beber|alimentação
@@ -126,7 +126,7 @@ ko:1카페|레스토랑|음식
 ja:1カフェ|喫茶|食べ物|飲食|コーヒー|珈琲|お茶|茶|紅茶|食事|軽食
 cs:3kavárna|restaurace|hospoda|jídlo
 sk:3kaviareň|reštaurácia|pohostinstvo|jedlo
-nl:3café|3restaurant|3eten
+nl:3tearoom|3brasserie|3café|3restaurant|3eten
 zh-Hant:2咖啡廳|咖啡|3cafe|餐廳|飲食
 pl:3kawiarnia|4restauracja|kawa|jedzenie
 pt:3café|3restaurante|alimentação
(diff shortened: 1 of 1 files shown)


data/categories.txt
```diff
@@ -126,7 +126,7 @@ ko:1카페|레스토랑|음식
 ja:1カフェ|喫茶|食べ物|飲食|コーヒー|珈琲|お茶|茶|紅茶|食事|軽食
 cs:3kavárna|restaurace|hospoda|jídlo
 sk:3kaviareň|reštaurácia|pohostinstvo|jedlo
-nl:3café|3restaurant|3eten
+nl:3tearoom|3brasserie|3café|3restaurant|3eten
```
deathbaba: Is it better to leave restaurant here as a search synonym? When you are hungry, you probably can type restaurant and would be happy to go into a nearest cafe if it is nearby.

data/categories.txt
```diff
@@ -97,7 +97,7 @@ ko:바|술집|맥주|음료|음식
 ja:1居酒屋|バー|パブ|ビール|飲食|飲み屋|呑み屋|酒
 cs:2bar|pivo|pití|jídlo|voda|pitná voda|tekutiny
 sk:2bar|pivo|pitie|jedlo|voda|pitná voda|tekutiny
-nl:2bar|2pub|bier|drinken|eten
+nl:2bar|2pub|2café|bier|drinken|eten
```
deathbaba: When user types 'cafe' in the search, he probably doesn't want to see a bar. Why do you think it should be here as a search synonym?

deathbaba:
Thanks for your pull request! Please check my comments.

For us to accept it, please accept our CLA (it's just a formality): http://maps.me/cla/

deathbaba:
Can you please also make a list of shops types which should be there (but they don't)? Please also check the latest categories.txt from master or new-search branch, not the publicly released version. We have split shops and temples in the code but did not release it yet.
Split by religions is not ideal, but it's more popular in the OSM (and it's enough for most people, looking for their specific religious places). We can introduce other tags in the future when they become popular.

data/categories.txt
sanderd17: That's what I thought too, so the restaurant is still there.

data/categories.txt
sanderd17: Because that's how it's translated. A café in Belgium and the Netherlands is a place where they mostly sell beer, and it has nothing to do with coffee. That's also why I switched café with tearoom as the default in the amenity=cafe category. 

See https://nl.wikipedia.org/wiki/Caf%C3%A9 which has "bar" as synonym (kroeg and knijp are also synonyms, but they're a bit older and not used a lot anymore). It also translates to the English page titled "Bar". The terminology changed during history, where first the more expensive ones called themselves "café" (which was a fancy French term in those days), but as cheaper pubs also started to call themselves "café", the more expensive businesses changed their names to even more fancy names like "brasserie" (more focused on selling meals, but still open for drinks only) and when English became popular "tearoom" (focused on being open during the day, and selling coffee and cake).

I even doubted to remove cafe from the amenity=cafe section, but decided to not remove it as mappers also often mis-tag it (thinking the English cafe matches the Dutch café). So by including café in both, the correct results will certainly be in there.

(note that coffeeshop should also be avoided as a translation of café, a coffeeshop is an euphemism for a place where legal soft-drugs are sold in the Netherlands, and it's never used as a place where only coffee is sold.)

I know it's all a bit complicated, but terms evolve as fancier businesses make up their new fancy names, and cheaper businesses copy the fancy names from more expensive ones.

data/categories.txt
deathbaba: Ouch, my mistake, thanks!

sanderd17:
I accepted the CLA

About the shops, this list could easily be split as a start:

`shop|shop-convenience|shop-mall|shop-doityourself|shop-mobile_phone|shop-chemist|shop-garden_centre`

I guess other tags should grow organically.

deathbaba:
We’ll split them when translations will be ready (soon). This list was a very long one, and was already split few weeks ago :)

PR #1753, opened 2016-02-01
[categories][search] Translations for different religions/places of worship.

deathbaba:


deathbaba:
This one replaces #1745

PR #1814, opened 2016-02-04
[android] [DNM] Don't send statistics about background location checks, that are used to suggest user to download missing maps on wifi.

yunikkk:


trashkalmar:
LGTM

PR #1845, opened 2016-02-08
Fixed the server for Djibouti.

therearesomewhocallmetim:
For some reason, we couldn't get the list of siblings on that particular machine.

### DIFF
--- a/tools/SiblingKiller.py
+++ b/tools/SiblingKiller.py
@@ -8,6 +8,8 @@ import urllib2
 import socket
 from subprocess import Popen, PIPE
 from time import sleep
+import sys
+
 
 import logging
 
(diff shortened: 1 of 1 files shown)


tools/SiblingKiller.py
```diff
@@ -89,7 +91,7 @@ def kill(self, pid=0, pids=list()):
 
     def siblings(self):
         my_name = self.my_process_name()
-        return filter(lambda x: x < self.__my_pid, 
+        return filter(lambda x: x < self.__my_pid,
```
ygorshenin: Are you sure that siblings pid is less than self.__my_pid?

therearesomewhocallmetim: They make up a queue. In order not to have all siblings kill each other so that none is left, I kill only those whose id's are lower than "mine", thus we can be sure that one will be left. There were situations in the past when the server couldn't start because SiblingKillers all killed each other.

PR #1856, opened 2016-02-08
Create temp. folders for certain tests

therearesomewhocallmetim:
Some tests fail when running in parallel because they compete for certain resources. To eliminate that situation we now create temporary folders for those tests and pass those folders as parameters.

### DIFF
--- a/tools/integration_tests_runner.py
+++ b/tools/integration_tests_runner.py
@@ -1,12 +1,15 @@
 from __future__ import print_function
-from optparse import OptionParser
-import subprocess
+
+import logging
 import multiprocessing
+from optparse import OptionParser
+from os import path
+import shutil
+import subprocess
+import tempfile
 from threading import Lock
 from threading import Thread
 import traceback
-import logging
-from os import path
 
 
 from Queue import Queue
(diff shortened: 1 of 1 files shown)


therearesomewhocallmetim:
@ygorshenin @gardster @syershov PTAL

tools/integration_tests_runner.py
```diff
@@ -1,11 +1,14 @@
...
+
+import logging
 import multiprocessing
+from optparse import OptionParser
+import tempfile
 from threading import Lock
 from threading import Thread
 import traceback
-import logging
+import shutil
```
gardster: sort please

tools/integration_tests_runner.py
```diff
@@ -58,13 +64,29 @@ def exec_tests_in_queue(self):
                     return
 
                 test_file, test = self.tests.get()
-                self.exec_test(test_file, test)
+
+#                    test_file = '{} "--user-resource-path={}"'.format(test_file, tmpdir) #--data_path
```
gardster: What is this?

tools/integration_tests_runner.py
```diff
@@ -58,13 +64,29 @@ def exec_tests_in_queue(self):
...
                 logging.error(traceback.format_exc())
 
 
-    def exec_test(self, test_file, test):
-        out, err, result = self.get_tests_from_exec_file(test_file, '--filter={test}'.format(test=test))
+    def exec_test(self, test_file, test, clean_env=False):
+        keys = '--filter={test}'.format(test=test)
+        if clean_env:
+            tmpdir = tempfile.mkdtemp()
+            logging.debug("Temp dir: {}".format(tmpdir))
```
gardster: Some of our systems has a old python interpreter. They can't handle {} string formatting. Please, use {0} instead.

ygorshenin:
LGTM

gardster:
LGTM

PR #2002, opened 2016-02-24
Use THREAD_CHECKER under Android

kshalnev:
С архитектурой DRAPE очереди NV больше нет, мы можем использовать ThreadChecker.

### DIFF
--- a/base/thread_checker.hpp
+++ b/base/thread_checker.hpp
@@ -23,9 +23,7 @@ private:
   DISALLOW_COPY_AND_MOVE(ThreadChecker);
 };
 
-// ThreadChecker checker is not valid on Android.
-// UI thread (NV thread) can change it's handle value during app lifecycle.
-#if defined(DEBUG) && !defined(OMIM_OS_ANDROID)
+#if defined(DEBUG)
   #define DECLARE_THREAD_CHECKER(threadCheckerName) ThreadChecker threadCheckerName
   #define ASSERT_THREAD_CHECKER(threadCheckerName, msg) ASSERT(threadCheckerName.CalledOnOriginalThread(), msg)
 #else


kshalnev:
@rokuz @trashkalmar @yunikkk @bykoianko @deathbaba  PTAL

ygorshenin:
What the point in this CL? Thread id of the UI thread may be changed during the application lifetime.

deathbaba:
@yunikkk is @ygorshenin right? Does UI thread change in an app lifetime?

P.S. @kshalnev have you tested it on Android?

ygorshenin:
I mean, the main reason why thread checker was disabled on Android is because main thread may be changed when application goes to/back from background. From the multithreaded point of view everything is ok as there're synchronization edges between old and new UI thread, but from the ThreadChecker's point of view there is a data rce.

kshalnev:
@ygorshenin, we need to know that Storage is accessed from the one thread all time. For that thread checkers have been added. Thread checkers work fine under iOS/Qt, but disabled for Android (due to legacy NV architecture). 
I didn't find info about changing of UI thread id. Do you have any link to doc?
@deathbaba, yes, I tested it on  Android.

deathbaba:
@ygorshenin it was true for the old rendering engine, but probably is not true any more :)

deathbaba:
"main" or "UI" thread in the old engine actually was our own thread, which was created/destroyed by some events. Now we have only "true" system-created "main" thread, as far as I understand it.

deathbaba:
LGTM

yunikkk:
@deathbaba @kshalnev wait, UI thread on android does not change if app wasn't destroyed completely and then recreated from scratch.

yunikkk:
So it's correct to add thread checks. LGTM.

ygorshenin:
LGTM

rokuz:
LGTM

PR #2005, opened 2016-02-24
[search] Added a calculation of address search completeness.

mpimenov:
MAPSME-108

### DIFF
--- a/search/search_quality_tests/search_quality_tests.cpp
+++ b/search/search_quality_tests/search_quality_tests.cpp
@@ -75,7 +75,7 @@ struct CompletenessQuery
 {
   string m_query;
   unique_ptr<TestSearchRequest> m_request;
-  string m_country;
+  string m_mwmName;
   uint64_t m_featureId = 0;
   double m_lat = 0;
   double m_lon = 0;
@@ -202,10 +202,10 @@ void Split(string const & s, char delim, vector<string> & parts)
 
 // Returns the position of the result that is expected to be found by geocoder completeness
 // tests in the |result| vector or -1 if it does not occur there.
-int FindResult(TestSearchEngine & engine, string const & country, uint64_t const featureId,
+int FindResult(TestSearchEngine & engine, string const & mwmName, uint64_t const featureId,
                double const lat, double const lon, vector<search::Result> const & results)
 {
-  auto const mwmId = engine.GetMwmIdByCountryFile(platform::CountryFile(country));
+  auto const mwmId = engine.GetMwmIdByCountryFile(platform::CountryFile(mwmName));
   FeatureID const expectedFeatureId(mwmId, featureId);
   for (size_t i = 0; i < results.size(); ++i)
   {
@@ -272,7 +272,7 @@ void CheckCompleteness(string const & path, m2::RectD const & viewport, TestSear
       ++malformedQueries;
       continue;
     }
-    string const mwmName = parts[0].substr(0, idx);
+    string mwmName = parts[0].substr(0, idx);
     string const kMwmSuffix = ".mwm";
     if (!strings::EndsWith(mwmName, kMwmSuffix))
     {
@@ -300,12 +300,14 @@ void CheckCompleteness(string const & path, m2::RectD const & viewport, TestSear
     string const street = parts[5];
     string const house = parts[6];
 
-    string country = mwmName.substr(0, mwmName.size() - kMwmSuffix.size());
+    mwmName = mwmName.substr(0, mwmName.size() - kMwmSuffix.size());
+    string country = mwmName;
+    replace(country.begin(), country.end(), '_', ' ');
     string const query = country + " " + city + " " + street + " " + house + " ";
 
     CompletenessQuery q;
     q.m_query = query;
-    q.m_country = move(country);
+    q.m_mwmName = mwmName;
     q.m_featureId = featureId;
     q.m_lat = lat;
     q.m_lon = lon;
@@ -320,7 +322,7 @@ void CheckCompleteness(string const & path, m2::RectD const & viewport, TestSear
 
     LOG(LDEBUG, (q.m_query, q.m_request->Results()));
     int pos =
-        FindResult(engine, q.m_country, q.m_featureId, q.m_lat, q.m_lon, q.m_request->Results());
+        FindResult(engine, q.m_mwmName, q.m_featureId, q.m_lat, q.m_lon, q.m_request->Results());
     if (pos >= 0)
       ++expectedResultsFound;
     if (pos == 0)
--- a/search/search_quality_tests/search_quality_tests.cpp
+++ b/search/search_quality_tests/search_quality_tests.cpp
@@ -68,6 +68,20 @@ map<string, m2::RectD> const kViewports = {
 string const kDefaultQueriesPathSuffix = "/../search/search_quality_tests/queries.txt";
 string const kEmptyResult = "<empty>";
 
+// todo(@m) We should not need that much.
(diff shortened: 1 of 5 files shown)


mpimenov:
@ygorshenin PTAL

search/search_quality_tests/search_quality_tests.cpp
```diff
@@ -160,8 +176,163 @@ void CalcStatistics(vector<double> const & a, double & avg, double & maximum, do
...
+
+  my::Timer timer;
+
+  uint32_t totalQueries = 0;
+  uint32_t malformedQueries = 0;
+  uint32_t expectedResultsFound = 0;
+  uint32_t expectedResultsTop1 = 0;
+
+  // todo(@m) Process the queries on the fly and do not keep them.
+  vector<string> queries;
```
ygorshenin: It seems that you need to create a structure containing country, id, lat, lon and create a single vector of these structures.

search/search_quality_tests/search_quality_tests.cpp
```diff
@@ -160,8 +176,163 @@ void CalcStatistics(vector<double> const & a, double & avg, double & maximum, do
...
+      LOG(LINFO, ("Bad mwm name:", s));
+      ++malformedQueries;
+      continue;
+    }
+    string const featureIdStr = parts[0].substr(idx + 1);
+    uint64_t featureId;
+    if (!strings::to_uint64(featureIdStr, featureId))
+    {
+      LOG(LINFO, ("Bad feature id:", s));
+      ++malformedQueries;
```
ygorshenin: How about to simply throw an exception, and in the exception handler increment number of malformed queries and log error. I suggest you to extract query parsing code into individual function which will fill a structure with country, city, street, house, etc. and return false (or error code, or exception) in case of malformed entry.

search/search_quality_tests/search_quality_tests.cpp
```diff
@@ -160,8 +176,163 @@ void CalcStatistics(vector<double> const & a, double & avg, double & maximum, do
...
+  cout << "Malformed queries: " << malformedQueries << endl;
+  cout << "Expected results found: " << expectedResultsFound << " ("
+       << expectedResultsFoundPercentage << "%)." << endl;
+  cout << "Expected results found in the top1 slot: " << expectedResultsTop1 << " ("
+       << expectedResultsTop1Percentage << "%)." << endl;
+}
+
 int main(int argc, char * argv[])
 {
+  ChangeMaxNumberOfOpenFiles(4000);
```
ygorshenin: Don't use 4000, just update soft limit to the hard limit.

search/search_quality_tests/search_quality_tests.cpp
```diff
@@ -160,8 +176,163 @@ void CalcStatistics(vector<double> const & a, double & avg, double & maximum, do
...
+    {
+      LOG(LINFO, ("Could not find \':\':", s));
+      ++malformedQueries;
+      continue;
+    }
+    string const mwmName = parts[0].substr(0, idx);
+    string const kMwmSuffix = ".mwm";
+    if (!strings::EndsWith(mwmName, kMwmSuffix))
+    {
+      LOG(LINFO, ("Bad mwm name:", s));
```
ygorshenin: LERROR should be used here, right?

search/search_quality_tests/search_quality_tests.cpp
syershov: As i know, on mac, getrlimit returns uint64_t::max() for hard limit. 
``` c++
msk-wire25-darkbookpro-yershov:~ $ cat fileno_test.cpp
#include <iostream>
#include <sys/resource.h>

int main() {
  struct rlimit rl;
  getrlimit (RLIMIT_NOFILE, &rl);
  std::cout << "soft: " << rl.rlim_cur << " hard: " << rl.rlim_max << std::endl;
  return 0;
}

msk-wire25-darkbookpro-yershov:~ $ ./fileno_test
soft: 7168 hard: 9223372036854775807
msk-wire25-darkbookpro-yershov:~ $
```

search/search_quality_tests/search_quality_tests.cpp
```diff
@@ -60,6 +68,14 @@ map<string, m2::RectD> const kViewports = {
 string const kDefaultQueriesPathSuffix = "/../search/search_quality_tests/queries.txt";
 string const kEmptyResult = "<empty>";
 
+void ChangeMaxNumberOfOpenFiles(size_t n)
+{
+  struct rlimit rlp;
+  getrlimit(RLIMIT_NOFILE, &rlp);
```
deathbaba: It won't compile on Windows.

ygorshenin: So what? We are not planning to use this tool on Windows.

deathbaba: If you exclude it from windows builds in pro file or in #ifdef OMIM_OS_WINDOWS, then it’s ok.

ygorshenin: Or we can protect this rlimit magic with ifdef. @mpimenov whyt?

search/search_quality_tests/search_quality_tests.cpp
mpimenov: done

search/search_quality_tests/search_quality_tests.cpp
mpimenov: No, why? One bad query should not stop the program.

search/search_quality_tests/search_quality_tests.cpp
mpimenov: I'll stick with 4000. It is already more than we should use.

search/search_quality_tests/search_quality_tests.cpp
mpimenov: Added ifdefs.

search/search_quality_tests/search_quality_tests.cpp
mpimenov: done

search/search_quality_tests/search_quality_tests.cpp
mpimenov: I don't see how it can clarify this code.
I don't like the overall organisation now: completeness test should probably be a different binary (and then your considerations should be applied) but as it is now I think it's better to keep the completeness code less scattered.

mpimenov:
PTAL

mpimenov:
PTAL

mpimenov:
@ygorshenin @deathbaba @syershov PTAL

search/search_quality_tests/search_quality_tests.cpp
```diff
@@ -160,8 +191,166 @@ void CalcStatistics(vector<double> const & a, double & avg, double & maximum, do
...
+  }
+
+  // Another attempt. If the queries are stale, feature id is useless.
+  // However, some information may be recovered from (lat, lon).
+  double const kEps = 1e-6;
+  for (size_t i = 0; i < results.size(); ++i)
+  {
+    auto const & r = results[i];
+    if (r.HasPoint() &&
+        my::AlmostEqualAbs(r.GetFeatureCenter(), MercatorBounds::FromLatLon(lat, lon), kEps))
```
ygorshenin: Please, add a curly braces around branch.

mpimenov: done

mpimenov:
Переписал на исключения. PTAL

ygorshenin:
LGTM

deathbaba:
LGTM

syershov:
LGTM

PR #2026, opened 2016-02-25
Added Swahili Translations

choquette33:
Peace Corps Volunteers worked with their local community members in Tanzania to create this list.

PR #2029, opened 2016-02-25
[categories] Swahili translations from our Peace Corps friends.

deathbaba:
This requests is a fixed version of #2026.
@choquette33 for correct Swahili support we need also translations for strings.txt file. Our wiki has more details: https://github.com/mapsme/omim/wiki/strings.txt

Not all translations are present, we have added several new types.

I would suggest to take strings.txt from new-search branch, as it's more fresh and under active development now ( https://github.com/mapsme/omim/blob/new-search/strings.txt )

### DIFF
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -26,6 +26,7 @@ id:2Atm|uang
 ro:3Bancomat|bani
 nb:2Penger|minibank
 fi:2Raha|pankkiautomaatti
+sw:Benki|fedha
 
 amenity-bank
 en:3Bank|money|U+1F3E6|U+1F4B0|U+1F4B2|U+1F4B3|U+1F4B4|U+1F4B5|U+1F4B6|U+1F4B7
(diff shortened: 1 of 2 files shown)


Zverik:
I don't see any autocompletion numbers in these strings. Are they needed?

Also, without strings.txt translations these I pretty useless, I presume?

deathbaba:
No, they are not completely useless - Place Page on a map will have a correct translation.

Zverik:
But does the app support Swahili? Shouldn't the language be added to some list?

deathbaba:
I’ll do it on a new-search branch first and then pull needed commits into the master.

Zverik:
What do you mean? This PR is for the master branch.
As for the question, I meant, do we have a list of supported languages somewhere, so if we don't add `sw` to it, this translation won't be used?

deathbaba:
Ok, ok, here is it, I’ve added in a new commit.

choquette33:
Ok- I'll take a look at this one: https://github.com/mapsme/omim/blob/new-search/strings.txt

PR #2121, opened 2016-03-04
Fixed editor crashes after merging editor config.

deathbaba:
@mgsergio Please check it out.


data/editor.config
```diff
@@ -124,6 +124,12 @@
     </fields>
 
     <types>
+      <type id="building">
```
Zverik: can_add="no"

mgsergio:
LGTH

PR #2128, opened 2016-03-04
[search] Fixed gen_samples script.

ygorshenin:
gen-samples script now reads samples specification from stdin.


ygorshenin:
@mpimenov PTAL

mpimenov:
LGTM

PR #2152, opened 2016-03-09
[new downloader][android] add: Show confirmation when trying to delete map with unsaved editor changes.

trashkalmar:


yunikkk:
LGTM

PR #2155, opened 2016-03-09
Fixed crash when searchresult doesn't contain valid featureId

yunikkk:


map/framework.cpp
```diff
@@ -1132,12 +1132,13 @@ bool Framework::QueryMayBeSkipped(search::SearchParams const & params,
 
 void Framework::LoadSearchResultMetadata(search::Result & res) const
 {
-  if (res.m_metadata.m_isInitialized)
+  if (res.m_metadata.m_isInitialized ||
+      res.GetResultType() != FeatureType::FeatureType)
```
deathbaba: Странно, что тут перенесся код. В одну строку смотрелось бы хорошо.

deathbaba: != search::Result::RESULT_FEATURE

deathbaba: Кстати, по-хорошему надо таки выставлять res.m_metadata.m_isInitialized в true для не-фич.

map/framework.cpp
```diff
@@ -1132,12 +1132,13 @@ bool Framework::QueryMayBeSkipped(search::SearchParams const & params,
...
     return;
 
   FeatureID const & id = res.GetFeatureID();
-  if (id.IsValid())
-    search::ProcessMetadata(*GetFeatureByID(id), res.m_metadata);
+  ASSERT(id.IsValid(), ("Search result doesn't contain valid FeatureID."));
+  search::ProcessMetadata(*GetFeatureByID(id), res.m_metadata);
```
deathbaba: Тут нужна TODO на будущее, что-то вроде Reuse the same formatting metadata code which is used in MapObject and place_page::Info.

map/framework.cpp
yunikkk: Упс.

map/framework.cpp
yunikkk: PTAL

map/framework.cpp
```diff
@@ -1132,12 +1132,13 @@ bool Framework::QueryMayBeSkipped(search::SearchParams const & params,
 
 void Framework::LoadSearchResultMetadata(search::Result & res) const
 {
-  if (res.m_metadata.m_isInitialized)
+  if (res.m_metadata.m_isInitialized || res.GetResultType() != search::Result::RESULT_FEATURE)
```
deathbaba:
```
if (res.m_metadata.m_isInitialized)
  return;
if (res.GetResultType() != search::Result::RESULT_FEATURE)
{
  res.m_metadata.m_isInitialized = true;
  return;
}
```

ygorshenin: Disagree. Why metadata should be marked as initialized for non feature results?

deathbaba: Because it can potentially make code faster. And it’s already set initialized in some cases (see one of search::Result constructor).

But I’m not insisting on it. This code should be refactored and thrown out anyway.

ygorshenin: Non feature results do not have metadata, how it can be initialized? Lets leave this field as is.

deathbaba: ok.

ygorshenin: On the other hand, we can treat non feature results as having empty metadata, and in this case we can (as you mentioned) reduce number of redundant calls. In any case, if this code will gone away, both approaches seems good to me now.

yunikkk: So, who will merge it first?)

ygorshenin:
LGTM

PR #2162, opened 2016-03-10
[search][build] Fixed build deps for search quality tools.

ygorshenin:
search quality tools depends on search_tests_support, so it must be declared first in SUBDIRS.


ygorshenin:
@deathbaba PTAL

PR #2180, opened 2016-03-11
[search] Relax the condition of matching a locality to a country.

mpimenov:
It is enough that search model reports a feature to be a country
and there is no need for additional checks. Besides, the infoGetter
check looks broken: "Kingdom of Lesotho" does not have "Lesotho" as
a prefix.

This change cannot be applied to the current implementation of
states checker, however, so the prefix check is infoGetter is not
removed in this CL.


mpimenov:
PTAL

ygorshenin:
LGTM

ygorshenin:
But we need to get statistics about localities and select a better pre-localities filter.

PR #2230, opened 2016-03-15
[search] Removed wrong and faulty cache.

ygorshenin:
All features layer matchers are kept in Geocoder cache, and it's ok since we do not want to re-read street s or buildings vicinities, but it's wrong to keep m_houseToStreetTable in features layer matcher, because:
* HouseToStreetTable contains Reader, therefore, Geocoder actually had been kept Readers to all mwms, even when idle
* there is no reason to keep HouseToStreetTable between search sessions, because it's impossible to use FeaturesLayerMatcher without MwmHandle, and MwmHandle is impossible without MwmValue, and MwmValue contains a Reader.
* we already have a cache of MwmValues in MwmSet, so some kind of caching already exists

@mpimenov PTAL

### DIFF
--- a/search/v2/features_layer_matcher.cpp
+++ b/search/v2/features_layer_matcher.cpp
@@ -1,6 +1,7 @@
 #include "search/v2/features_layer_matcher.hpp"
 
 #include "search/reverse_geocoder.hpp"
+#include "search/v2/house_to_street_table.hpp"
 
 #include "indexer/scales.hpp"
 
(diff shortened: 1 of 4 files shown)


ygorshenin:
https://jira.mail.ru/browse/MAPSME-244

mpimenov:
Также https://jira.mail.ru/browse/MAPSME-261

mpimenov:
LGTM

PR #2261, opened 2016-03-16
From string conversion fixes.

deathbaba:


deathbaba:
DUDE JTTP

base/base_tests/string_utils_test.cpp
```diff
@@ -165,6 +165,10 @@ UNIT_TEST(to_double)
 
   s = "123.456 we don't parse it.";
   TEST(!strings::to_double(s, d), ());
+
+  TEST(!strings::to_double("INF", d), ());
```
ygorshenin: Why? What the problem to parse Inf or Nan?

base/base_tests/string_utils_test.cpp
```diff
@@ -165,6 +165,10 @@ UNIT_TEST(to_double)
...
+  TEST(!strings::to_double("NAN", d), ());
+  TEST(!strings::to_double((ostringstream() << numeric_limits<long double>::max()).str(), d), ());
```
ygorshenin: The size of long double is not specified, as far as I know. I.e. it can be that long double is the same as double.

base/base_tests/string_utils_test.cpp
deathbaba: Our code is designed to work with finite values from OSM.

deathbaba: The same is true for overflows - they are not handled now.

deathbaba:
Fixed @ygorshenin comment, PTAL please.
DUDE JTTP

deathbaba:
@ygorshenin merge please.
Падают тесты реверс геокодера
map_tests/address_tests.cpp:25 TEST(stName == key) советская свердлова

ygorshenin:
Okay, LGTM

PR #2341, opened 2016-03-18
[COMMIT AS SOON AS SMALL MWMS WILL BE ON ALL OUR SERVERS] Removed hacks for maps downloading ULRs.

ygorshenin:


ygorshenin:
@Zverik , I think you're the best person to review and merge this when it will be appropriate.

PR #2370, opened 2016-03-21
[ios] Fixed small back button tap area.

igrechuhin:


iphone/Maps/Classes/UIViewController+Navigation.mm
```diff
@@ -1,21 +1,33 @@
...
+  UIBarButtonItem * spacer = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemFixedSpace target:nil action:nil];
+  spacer.width = -kButtonExtraWidth;
+  return spacer;
+}
+
+- (UIBarButtonItem *)backButton
 {
   UIImage * backImage = [UIImage imageNamed:@"ic_nav_bar_back"];
-  UIButton * button = [[UIButton alloc] initWithFrame:{{}, backImage.size}];
+  CGSize buttonSize = {backImage.size.width + kButtonExtraWidth, backImage.size.height};
```
VladiMihaylenko: const

PR #2414, opened 2016-03-22
[search] Paranoid revert of GetCenter.

mpimenov:
Does not change the behaviour but neither did the previous change.


ygorshenin:
LGTM

PR #2444, opened 2016-03-23
Scale changing after resize in follow and rotate mode on Android fixed.

darina:
https://jira.mail.ru/browse/MAPSME-449


rokuz:
LGTM

PR #2493, opened 2016-03-24
[search][quality] Implemented set balancing for samples.

ygorshenin:
@mpimenov PTAL

Set balancing is needed for better SVM class separation. For biased sets, SVM generally produces worse results.


ygorshenin:
PTAL

mpimenov:
LGTM

PR #2637, opened 2016-03-30
[user-marks] DebugPrint for UserMark::Type.

ygorshenin:
Follow-up fixes for https://github.com/mapsme/omim/pull/2384

@deathbaba PTAL


deathbaba:
:+1:

PR #2793, opened 2016-04-06
[new downloader][android] fix: Show bottom panel instead of toolbar buttons in downloader.

trashkalmar:


trashkalmar:
@yunikkk PTAL

PR #2799, opened 2016-04-06
[search] Don’t calculate address for viewport search result points.

vng:
Cherry pick #2791


ygorshenin:
DUDE JTTP

ygorshenin:
LGTM, but probably we need to wait for Jenkins.

Issue #2805, opened 2016-04-06
How does it connect to OpenStreetMap?


rugk:
Especially: Do you use HTTPS? (And is the certificate correctly checked?)

For the [login pages](https://github.com/mapsme/omim/blob/d47d55febe6a718fed1487bbdb92501b8a41ca29/android/src/com/mapswithme/util/Constants.java#L35-L36) I already found HTTPS links, but what about all other requests, e.g. when submitting new objects or comments?

deathbaba:
Yes, https is used.

rugk:
Okay, but as you can see in my Pull Request there are still HTTP links...

PR #2806, opened 2016-04-06
Convert HTTP OpenStreetMap links to HTTPS ones

rugk:
Using this regexp: https://regex101.com/r/pD7aO7/2 (created by using the [HTTPS Everywhere ruleset](https://github.com/EFForg/https-everywhere/blob/375124aa311c82db8c87c1882c41b082da4b84a4/src/chrome/content/rules/OpenStreetMap.org.xml))

I did not tested whether this compiles correctly. It would be nice if you could check it.
But as I basically only changed added few characters I doubt that this PR will break something.

Closes https://github.com/mapsme/omim/issues/2805


deathbaba:
Hi and thanks for contributing to MAPS.ME!
Can you please sign our CLA before merging? http://maps.me/cla/

About this change: Please check our guide about changing localization strings. In general, you need to:
1. First commit with changes to strings.txt
2. Second commit with auto generated localizations from strings.txt
3. Third commit with changes to data/faq.html

rugk:
> Can you please sign our CLA before merging?

Done.

> 1. First commit with changes to strings.txt
2. Second commit with auto generated localizations from strings.txt
3. Third commit with changes to data/faq.html

In all these files the changes should be represented as I just searched and replaced the strings through all files.
As I did not really translated the strings, but more or less just added one character in some lines, is it okay to just use this one commit I made? I am not really keen on splitting this commit.

deathbaba:
No, it is not okay. Because:
1. Looks like you replaced also some whitespaces.
2. We always split changes to strings.txt from regenerated strings for easier merge conflicts resolving.

rugk:
Okay, if anyone else wants to take this regexp or so, he can do.

PR #2818, opened 2016-04-07
[android] fix: Prevent crash if the TTS is corrupted.

trashkalmar:


trashkalmar:
Копия #2761, но на мастер.
@yunikkk, @deathbaba PTAL and MERGE.

deathbaba:
LGTM

PR #2858, opened 2016-04-10
[strings] Fixed translations of zh-Hant and zh-Hans

jiazheng0609:
Sorry, I don't know how to run the script. I'm not a unix user.


deathbaba:
Ok, thanks, I'll do that.

PR #2859, opened 2016-04-11
[strings] Regenerated localizations fixed by contributors.

deathbaba:
@trashkalmar or @vng please check and merge.


trashkalmar:
Is there commit with changes in ```strings.txt```?

PR #2887, opened 2016-04-14
[ios] Fixed crash.

igrechuhin:
MWM<Base>MapDownloaderViewController retains its dataSources.
However its tableView does not retain its dataSource.
During switch of active dataSource, tableView was loosing its dataSource.
Table reload during dataSource switch led to crashes.


syershov:
LGTM

PR #2903, opened 2016-04-17
Improved German translation

Atalanttore:
I've replaced some anglicisms and made the German translation a little bit more consistent.

### DIFF
--- a/strings.txt
+++ b/strings.txt
@@ -8927,7 +8927,7 @@
     hu = %@ MB-t kell letöltened, de csak %@ MB hely szabad.
     el = Πρέπει να κάνετε λήψη %@ MB, αλλά μόνο %@ MB είναι διαθέσιμα.
     ar = تحتاج إلى تحميل %@ ميجابايت، ولكن لا يتوافر سوى %@ ميجا بايت.
-    de = Sie müssen %@ MB herunterladen, aber es sind nur %@ MB verfügbar.
+    de = Sie wollen %@ MB herunterladen, der Speicherplatz reicht aber nur für %@ MB.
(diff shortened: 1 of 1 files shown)


strings.txt
```diff
@@ -8925,7 +8927,7 @@
     hu = %@ MB-t kell letöltened, de csak %@ MB hely szabad.
     el = Πρέπει να κάνετε λήψη %@ MB, αλλά μόνο %@ MB είναι διαθέσιμα.
     ar = تحتاج إلى تحميل %@ ميجابايت، ولكن لا يتوافر سوى %@ ميجا بايت.
-    de = Der Download erfordert %@ MB, es sind jedoch nur %@ MB verfügbar.
+    de = Sie müssen %@ MB herunterladen, aber es sind nur %@ MB verfügbar.
```
hybridOL: This one should be improved. The new one sounds as if the download does not provide enough MB. Maybe better:
Sie wollen %@ MB herunterladen, der Speicherplatz reicht aber nur für %@ MB.

deathbaba:
Thanks!
Please check and sign our CLA http://maps.me/cla/ before merge.
And this instruction shows how to correctly update app interface strings after modifying strings.txt https://github.com/mapsme/omim/wiki/strings.txt

strings.txt
Atalanttore: According to your suggestion, I have improved this string.

Atalanttore:
CLA is checked and signed.

deathbaba:
Great! Thank you very much!

trashkalmar:
@deathbaba What about generated strings for platforms?

deathbaba:
Oops, forgot to recheck it. #2915 contains generated strings.

Issue #2906, opened 2016-04-18
Add static code analysis


sils1297:
Hi,

I'm a very happy user of this app and coincidentally I'm the author of a static code analysis framework coala and the service connected to it GitMate which can automatically review your pull requests (gitmate.io).

Apparently you don't have static code analysis yet, I can offer setting it up for you (we're also interested in trying the setup on some more sophisticated projects so this should be a win win). It can run on all languages and your docs as well without you having to bother with lots of different tools. It's easily integratable into any CI system or GitMate can be used for free (it'll just comment about newly introduced issues in the GitHub diff).

coala is AGPL and runs totally without any proprietary piece of software.

Is that interesting for you? What all would you like to have analyzed?

deathbaba:
Hi, can you please analyze our omim repository? Let's check the result :)

sils1297:
Alright, let's see what we have here

/me runs cloc.

blobgo:
Just FYI there are many other code analyzers: E.g. https://codacy.com/ and [many other](https://github.com/integrations/feature/code)...

sils1297:
Yes, there are lots of alternatives. If you want the *best* analysis for C your best bet is probably coverity scan for open source. (Good luck setting it up though and it's proprietary.) All the other platforms (codacy, scrutinizer, ...) are usually also proprietary and don't offer a real deal breaker.

You can either go for the good results (coverity etc.) or have something simple (codacy, gitmate, etc.) and you can choose between free and nonfree software. (HoundCI is a free alternative that does style checking only though. GitMate is coala based and all algorithms are free.) If you want to do analysis on multiple languages at once you'll have a hard time finding something that beats coala in simplicity, you can set everything up in one config and rule a whole fleet of open source tools and our native algorithms at once.

I'm of course biased towards GitMate and coala but hope the overview helps a bit. We're spending a bit of time on the competition and obviously they didn't convince us :P

PR #2912, opened 2016-04-18
WIP: Add static code analysis

sils1297:
related to https://github.com/mapsme/omim/issues/2906

still pretty early, I just did some basic PEP8ing to your python code, more stuff can follow and I think some stuff needs manual refactoring.

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -1,0 +1,11 @@
+[Default]
+ignore = 3party/**
+
+[cpplint]
+files = **/*.cpp
+bears = ClangComplexityBear
+
+[python-pep8]
+files = **/*.py
+bears = PEP8Bear
+tab_width = 2
--- a/.gitignore
+++ b/.gitignore
@@ -131,3 +131,6 @@ tools/android/yota.keystore
 android/secure.properties
 android/fabric.properties
 server
+
+# Git and coala backup files
+*.orig


deathbaba:
Why do we need that? I mean, just formatting changes?

3party folder should stay untouched, it will be hard to update other libraries after changes.

And python is not a crucial language in our project, we are more interested in C++/Java(Android)/ObjectiveC++ static code analyzers which can help us to find potential errors before they show up to our users.

sils1297:
This is formatting only, analysis can happen on any level. That was what I meant with "what do you want/need"

deathbaba:
Formatting fixes usually break line history (git blame), not sure that it's a good idea.

ygorshenin:
In any case, it's not a good idea to modify 3party - this code mostly written by non-maps.me developers and we sometimes pull latest versions.

sils1297:
I can totally exclude the folder. Also we can improve codestyle iterative with GitMate (shows only new issues) if you don't want an initial cleanup

deathbaba:
I was more interested in what your static analyzes say about our current code. Which issues can it find and highlight right now.

sils1297:
Hi,

I'll look into that, it might take some time because a lot of other stuff is eating up all my time.

To clarify: coala currently can do some light analysis language independently. For C we currently have (apart from the usual style stuff) a clang integration which can do a lot but clang usually needs a rather complicated setup, a rather sophisticated code clone detection but I'm afraid this one's not performant enough for this kind of code base (those algorithms have ugly complexities). A cyclomatic complexity and a more performant code clone detection is in the works. I'd look into setting up clang, te complexity measurement and the code clone detection then.

sils1297:
btw. your complexity looks good, we found one false positive only and evaluating now on how to improve our algorithm

PR #2921, opened 2016-04-19
Fixed an out-of-bounds access.

mpimenov:


mpimenov:
@bykoianko The only place where this function is used is routing/turns_generator.cpp. Do we even need it?

vng:
LGTM

ygorshenin:
LGTM

bykoianko:
Yes. We need to for cases when a name of a road contains a kilometer(mile) counter. For example, "MKAD, 60 km", "MKAD, 59 km".

LGTM

PR #2922, opened 2016-04-19
[generator, indexer] Fixed postcodes handling.

ygorshenin:
Fixed postcodes handling in generator - now they're stored in the
metadata.  Also, fixed iteration over features in features vector.


ygorshenin:
@vng @mpimenov @Zverik PTAL

ygorshenin:
@Zverik, we probably need to re-generate all our maps with this patch and check how size of maps will change.

mpimenov:
LGTM

PR #2926, opened 2016-04-20
[omim] dropped pin as default bookmark name.

VladiMihaylenko:
MAPSME-745


VladiMihaylenko:
@deathbaba ptal

VladiMihaylenko:
Решили не делать

PR #2959, opened 2016-04-22
[ios] Crash fix.

igrechuhin:
Take a copy of pasteboard string since it can accidentally become nil while we still use it.


syershov:
Cherry pick from release branch. LGTM

PR #3056, opened 2016-04-28
[ios] Fixed different table section header views in downloader.

igrechuhin:


iphone/Maps/Classes/MapDownloader/Cells/MWMMapDownloaderCellHeader.mm
```diff
@@ -0,0 +1,24 @@
...
+  {
+    self.font = [UIFont regular12];
+    self.textColor = [UIColor blackSecondaryText];
+  }
+  return self;
+}
+
+- (void)drawTextInRect:(CGRect)rect
+{
+  UIEdgeInsets const insets = {.top = 0, .left = 16, .bottom = 0, .right = 0};
```
VladiMihaylenko: {.left = 16};

PR #3077, opened 2016-04-29
[ios] Fixed download from map.

igrechuhin:
Added ability to download from map in case when routing is built but not yet started.


syershov:
LGTM

PR #3283, opened 2016-05-20
Revert "[search] Limit geocoder by max num of results."

ygorshenin:
Reverts mapsme/omim#3281


ygorshenin:
The problem is that search quality drops with this PR. Need to be investigated after https://github.com/mapsme/omim/pull/3259.

mpimenov:
LGTM

PR #3443, opened 2016-06-03
[search] A probabilistic tweak to locality scorer.

mpimenov:
https://jira.mail.ru/browse/MAPSME-1380

### DIFF
--- a/search/geocoder.cpp
+++ b/search/geocoder.cpp
@@ -694,9 +694,14 @@ void Geocoder::FillLocalityCandidates(coding::CompressedBitVector const * filter
   for (size_t startToken = 0; startToken < m_numTokens; ++startToken)
   {
     CBVPtr intersection;
+    CBVPtr unfilteredIntersection;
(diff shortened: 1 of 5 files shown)


search/geocoder.cpp
```diff
@@ -694,9 +694,14 @@ void Geocoder::FillLocalityCandidates(coding::CompressedBitVector const * filter
   for (size_t startToken = 0; startToken < m_numTokens; ++startToken)
   {
     CBVPtr intersection;
+    CBVPtr unfilteredIntersection;
```
ygorshenin: I guess unfilteredIntersection must be set to full unconditionally.

mpimenov: All uses of ```unfilteredIntersection``` are guarded. But setting to full is cheap, so

done

mpimenov:
PTAL

mpimenov:
PTAL

ygorshenin:
LGTM

PR #3427, opened 2016-06-02
Added support of dashes in route visualization

rokuz:

PR #3451, opened 2016-06-06
Added support of dashes in route visualization

rokuz:
https://github.com/mapsme/omim/pull/3427 на релизную ветку


rokuz:
DUDE JTTP

PR #3503, opened 2016-06-09
Keeping altitude in features. Linear least squares for feature point altitude.

bykoianko:
Вычисление энтропии отклонения точек фич по высоте от прямых полученных методом наименьших квадратов.

Вычисление минимально и максимальной высоты mwm-ок.

@ygorshenin @mpimenov обязательно посмотрите. Завтра на встрече буду говорить о результатах полученных на базе этого PR.

### DIFF
--- a/generator/feature_segments_checker/feature_segments_checker.cpp
+++ b/generator/feature_segments_checker/feature_segments_checker.cpp
@@ -122,6 +122,8 @@ public:
   uint32_t m_roadPointCount;
   /// Number of features for GetBicycleModel().IsRoad(feature) != true.
   uint32_t m_notRoadCount;
+  generator::SrtmTile::THeight m_minAltitude = generator::SrtmTile::kInvalidHeight;
+  generator::SrtmTile::THeight m_maxAltitude = generator::SrtmTile::kInvalidHeight;
 
   Processor(generator::SrtmTileManager & manager)
     : m_srtmManager(manager), m_roadCount(0), m_emptyRoadCount(0), m_roadPointCount(0), m_notRoadCount(0)
@@ -169,6 +171,15 @@ public:
       double const segmentLengthMeters = MercatorBounds::DistanceOnEarth(f.GetPoint(i - 1), f.GetPoint(i));
       distFromStartMeters += segmentLengthMeters;
       pointDists[i] = distFromStartMeters;
+    }
+
+    // Min and max altitudes.
+    for (auto const a : pointAltitudes)
+    {
+      if (m_minAltitude == generator::SrtmTile::kInvalidHeight || a < m_minAltitude)
+        m_minAltitude = a;
+      if (m_maxAltitude == generator::SrtmTile::kInvalidHeight || a > m_maxAltitude)
+        m_maxAltitude = a;
     }
 
     // Feature length and feature segment length.
@@ -320,7 +331,8 @@ int main(int argc, char ** argv)
   PrintCont(processor.m_leastSquaresDiff, "Altitude deviation of feature points from least squares line.",
             " internal feature point(s) deviate from linear model with ", " meter(s)");
 
-  cout << endl << "Road feature count = " << processor.m_roadCount << endl;
+  cout << endl << FLAGS_mwm_file_path << endl;
+  cout << "Road feature count = " << processor.m_roadCount << endl;
   cout << "Empty road feature count = " << processor.m_emptyRoadCount << endl;
   cout << "Unique road points count = " << processor.m_uniqueRoadPoints.size() << endl;
   cout << "All road point count = " << processor.m_roadPointCount << endl;
@@ -329,5 +341,7 @@ int main(int argc, char ** argv)
        << CalculateEntropy(processor.m_diffFromLinear) << endl;
   cout << "Entropy for altitude deviation of feature points from least squares line = "
        << CalculateEntropy(processor.m_leastSquaresDiff) << endl;
+  cout << "Min altitude of the mwm = " << processor.m_minAltitude << endl;
+  cout << "Max altitude of the mwm = " << processor.m_maxAltitude << endl;
   return 0;
 }
--- a/generator/feature_segments_checker/feature_segments_checker.cpp
+++ b/generator/feature_segments_checker/feature_segments_checker.cpp
@@ -20,6 +20,7 @@
 #include "std/fstream.hpp"
 #include "std/iostream.hpp"
 #include "std/map.hpp"
+#include "std/numeric.hpp"
 #include "std/set.hpp"
 #include "std/string.hpp"
 
@@ -75,14 +76,10 @@ void WriteCSV(Cont const & cont, string const & fileName)
     fout << a.first << "," << a.second << endl;
 }
 
-/// \returns expected point altitude in meters according to linear model.
-double ExpectedPointAltitude(int16_t startAltitudeMeters, int16_t endAltitudeMeters,
-                             double distFromStartMeters, double featureLengthMeters)
+/// \returns y = k * x + b. It's an expected altitude in meters.
+double ExpectedPointAltitude(double k, double b, double distFromStartMeters)
 {
-  if (featureLengthMeters == 0.0)
-    return 0;
-  double const k = (endAltitudeMeters - startAltitudeMeters) / featureLengthMeters;
-  return startAltitudeMeters + k * distFromStartMeters;
+  return k * distFromStartMeters + b;
 }
 
 class Processor
@@ -113,6 +110,10 @@ public:
   /// from linear model.
   /// Value is a number of features.
   map<int32_t, uint32_t> m_diffFromLinear;
+  /// Key is number of meters. It shows altitude deviation of intermediate feature points
+  /// from line calculated base on least squares method for all feature points.
+  /// Value is a number of features.
+  map<int32_t, uint32_t> m_leastSquaresDiff;
   /// Number of features for GetBicycleModel().IsRoad(feature) == true.
   uint32_t m_roadCount;
   /// Number of features for empty features with GetBicycleModel().IsRoad(feature).
@@ -151,12 +152,31 @@ public:
     for (uint32_t i = 0; i < numPoints; ++i)
       m_uniqueRoadPoints.insert(RoughPoint(f.GetPoint(i)));
 
+    // Preparing feature altitude and length.
+    vector<generator::SrtmTile::THeight> pointAltitudes(numPoints);
+    vector<double> pointDists(numPoints);
+    double distFromStartMeters = 0;
+    for (uint32_t i = 0; i < numPoints; ++i)
+    {
+      // Feature segment altitude.
+      pointAltitudes[i] = m_srtmManager.GetHeight(MercatorBounds::ToLatLon(f.GetPoint(i)));
+      if (i == 0)
+      {
+        pointDists[i] = 0;
+        continue;
+      }
+      // Feature segment length.
+      double const segmentLengthMeters = MercatorBounds::DistanceOnEarth(f.GetPoint(i - 1), f.GetPoint(i));
+      distFromStartMeters += segmentLengthMeters;
+      pointDists[i] = distFromStartMeters;
+    }
+
     // Feature length and feature segment length.
     double realFeatureLengthMeters = 0.0;
     for (uint32_t i = 0; i + 1 < numPoints; ++i)
     {
       // Feature segment length.
-      double const realSegmentLengthMeters = MercatorBounds::DistanceOnEarth(f.GetPoint(i), f.GetPoint(i + 1));
+      double const realSegmentLengthMeters = pointDists[i + 1] - pointDists[i];
       m_segLength[static_cast<uint32_t>(floor(realSegmentLengthMeters))]++;
 
       // Feature length.
@@ -165,10 +185,8 @@ public:
     m_featureLength[static_cast<uint32_t>(realFeatureLengthMeters)]++;
 
     // Feature altitude difference.
-    generator::SrtmTile::THeight const startAltitude =
-        m_srtmManager.GetHeight(MercatorBounds::ToLatLon(f.GetPoint(0)));
-    generator::SrtmTile::THeight const endAltitude =
-        m_srtmManager.GetHeight(MercatorBounds::ToLatLon(f.GetPoint(numPoints - 1)));
+    generator::SrtmTile::THeight const startAltitude = pointAltitudes[0];
+    generator::SrtmTile::THeight const endAltitude = pointAltitudes[numPoints - 1];
     int16_t const altitudeDiff = endAltitude - startAltitude;
     m_altitudeDiffs[altitudeDiff]++;
 
@@ -178,9 +196,7 @@ public:
     int32_t down = 0;
     for (uint32_t i = 0; i + 1 < numPoints; ++i)
     {
-      auto const segAltDiff =
-          (m_srtmManager.GetHeight(MercatorBounds::ToLatLon(f.GetPoint(i + 1))) -
-           m_srtmManager.GetHeight(MercatorBounds::ToLatLon(f.GetPoint(i))));
+      auto const segAltDiff = pointAltitudes[i + 1] - pointAltitudes[i];
       if (altitudeDiff >= 0)
       {
         // If the feature goes up generaly calculates how many meters it's necessary
@@ -212,19 +228,36 @@ public:
     if (realFeatureLengthMeters == 0.0)
       return;
 
-    double distFromStartMeters = 0;
+    double const k = (endAltitude - startAltitude) / realFeatureLengthMeters;
     for (uint32_t i = 1; i + 1 < numPoints; ++i)
     {
-      // Feature segment length.
-      double const segmentLengthMeters =
-          MercatorBounds::DistanceOnEarth(f.GetPoint(i - 1), f.GetPoint(i));
-      distFromStartMeters += segmentLengthMeters;
-
-      generator::SrtmTile::THeight const pointAltitude =
-          m_srtmManager.GetHeight(MercatorBounds::ToLatLon(f.GetPoint(i)));
-      int32_t const deviation = static_cast<int32_t>(ExpectedPointAltitude(startAltitude, endAltitude, distFromStartMeters,
-                                                                           realFeatureLengthMeters)) - pointAltitude;
+      int32_t const deviation =
+          static_cast<int32_t>(ExpectedPointAltitude(k, startAltitude, pointDists[i])) - pointAltitudes[i];
       m_diffFromLinear[deviation]++;
+    }
+
+    // Linear least squares for feature points.
+    {
+      long double const distSum = accumulate(pointDists.begin(), pointDists.end(), 0.0);
+      uint64_t const altSum = accumulate(pointAltitudes.begin(), pointAltitudes.end(), 0);
+      long double const distSquareSum = accumulate(pointDists.begin(), pointDists.end(), 0.0,
+                                                   [](double x, double y) { return x + y * y; });
+      long double distMultiplyAltSum = 0;
+      for (size_t i = 0; i < pointDists.size(); ++i)
+        distMultiplyAltSum += pointDists[i] * pointAltitudes[i];
+
+      if (distSum == 0.0 || distSquareSum == 0.0)
+        return;
+
+      double const b = (altSum/distSum - distMultiplyAltSum/distSquareSum) / (numPoints/distSum - distSum/distSquareSum);
+      double const k = (altSum - b * numPoints) / distSum;
+
+      for (uint32_t i = 0; i < numPoints; ++i)
+      {
+        int32_t const deviation =
+            static_cast<int32_t>(ExpectedPointAltitude(k, b, pointDists[i])) - pointAltitudes[i];
+        m_leastSquaresDiff[deviation]++;
+      }
     }
   }
 };
@@ -284,12 +317,17 @@ int main(int argc, char ** argv)
   PrintCont(processor.m_diffFromLinear, "Altitude deviation of internal feature points from linear model.",
             " internal feature point(s) deviate from linear model with ", " meter(s)");
 
+  PrintCont(processor.m_leastSquaresDiff, "Altitude deviation of feature points from least squares line.",
+            " internal feature point(s) deviate from linear model with ", " meter(s)");
+
   cout << endl << "Road feature count = " << processor.m_roadCount << endl;
   cout << "Empty road feature count = " << processor.m_emptyRoadCount << endl;
   cout << "Unique road points count = " << processor.m_uniqueRoadPoints.size() << endl;
   cout << "All road point count = " << processor.m_roadPointCount << endl;
   cout << "Not road feature count = " << processor.m_notRoadCount << endl;
-  cout << "Binary entropy for altitude deviation of internal feature points from linear model = "
+  cout << "Entropy for altitude deviation of internal feature points from linear model = "
        << CalculateEntropy(processor.m_diffFromLinear) << endl;
+  cout << "Entropy for altitude deviation of feature points from least squares line = "
+       << CalculateEntropy(processor.m_leastSquaresDiff) << endl;
   return 0;
 }
--- a/generator/feature_segments_checker/feature_segments_checker.cpp
+++ b/generator/feature_segments_checker/feature_segments_checker.cpp
@@ -15,6 +15,7 @@
 #include "platform/platform.hpp"
 
 #include "base/logging.hpp"
+#include "base/math.hpp"
 
 #include "std/cmath.hpp"
 #include "std/fstream.hpp"
@@ -31,6 +32,9 @@ DEFINE_string(mwm_file_path, "", "Path to an mwm file.");
 
 namespace
 {
+using TAltitude = generator::SrtmTile::THeight;
+using TAltitudeVec = vector<TAltitude>;
+
 routing::BicycleModel const & GetBicycleModel()
 {
   static routing::BicycleModel const instance;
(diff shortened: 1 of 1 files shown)


generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -212,20 +239,37 @@ class Processor
...
-      int32_t const deviation = static_cast<int32_t>(ExpectedPointAltitude(startAltitude, endAltitude, distFromStartMeters,
-                                                                           realFeatureLengthMeters)) - pointAltitude;
+      int32_t const deviation =
+          static_cast<int32_t>(ExpectedPointAltitude(k, startAltitude, pointDists[i])) - pointAltitudes[i];
       m_diffFromLinear[deviation]++;
     }
+
+    // Linear least squares for feature points.
+    {
+      long double const distSum = accumulate(pointDists.begin(), pointDists.end(), 0.0);
```
ygorshenin: A lot of misleading names. Please, create an individual method that takes a vector of xs and vector of ys and returns coeffs for linear model.

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -212,20 +239,37 @@ class Processor
...
+      long double const distSquareSum = accumulate(pointDists.begin(), pointDists.end(), 0.0,
+                                                   [](double x, double y) { return x + y * y; });
+      long double distMultiplyAltSum = 0;
+      for (size_t i = 0; i < pointDists.size(); ++i)
+        distMultiplyAltSum += pointDists[i] * pointAltitudes[i];
+
+      if (distSum == 0.0 || distSquareSum == 0.0)
+        return;
+
+      double const b = (altSum/distSum - distMultiplyAltSum/distSquareSum) / (numPoints/distSum - distSum/distSquareSum);
```
ygorshenin: It's much simpler to calculate k, then b:
```c++
k = (distMultiplyAltSum - distSum * altSum / numPoints) / (distSquareSum - distSum * distSum);
b = (altSum - k * distSum) / numPoints;
```

In any case, your expression for b is quite complex and may be simplified.

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,32 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+/// \returns false in case of error (e.g. if the line is parallel with ordinate axis)
+/// and true otherwise.
+bool LinearLeastSquaresFactors(vector<double> const & xs, TAltitudeVec const & ys, double & k, double & b)
 {
-  if (featureLengthMeters == 0.0)
-    return 0;
-  double const k = (endAltitudeMeters - startAltitudeMeters) / featureLengthMeters;
-  return startAltitudeMeters + k * distFromStartMeters;
+  long double const distSum = accumulate(xs.begin(), xs.end(), 0.0);
+  uint64_t const altSum = accumulate(ys.begin(), ys.end(), 0);
```
kshalnev: TAltitude const altSum?

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,32 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+  long double const distSquareSum = accumulate(xs.begin(), xs.end(), 0.0,
+                                               [](double x, double y) { return x + y * y; });
+  long double distMultiplyAltSum = 0;
+  for (size_t i = 0; i < xs.size(); ++i)
+    distMultiplyAltSum += xs[i] * ys[i];
+
+  if (distSum == 0.0 || distSquareSum == 0.0)
```
kshalnev: это можно проверять до вычисления distMultiplyAltSum

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Below the correct formulas if to calculate k first:
```c++
k = (altSum/numPoints - distMultiplyAltSum/distSum) / (distSum/numPoints - distSquareSum/distSum);
b = (altSum - k * distSum) / numPoints;
```
But if to do do the entropy would be a little more. I rewrote the code in another way. The entropy is a little less this way. 

> n any case, your expression for b is quite complex and may be simplified.

Yes it's possible. But in that case we'll have too big values while calculation which effects the precision of the result.

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Done

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Done

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Нельзя. TAltitude (int16_t) слишком маленькая.

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -151,12 +194,41 @@ class Processor
...
 
+    // Preparing feature altitude and length.
+    TAltitudeVec pointAltitudes(numPoints);
+    vector<double> pointDists(numPoints);
+    double distFromStartMeters = 0;
+    for (uint32_t i = 0; i < numPoints; ++i)
+    {
+      // Feature segment altitude.
+      TAltitude altitude = m_srtmManager.GetHeight(MercatorBounds::ToLatLon(f.GetPoint(i)));
+      pointAltitudes[i] = altitude == generator::SrtmTile::kInvalidHeight ? 0 : altitude;
```
bykoianko: Тут поправлена причина, по которой у нас для Ямала (где нет высот) вычислялась довольно большая энтропия. Теперь она равна 0 для Ямала.

bykoianko:
@ygorshenin @kshalnev PTAL

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,47 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+  long double result = 0.0;
+  for (auto item : vec)
+    result += item/devisor;
+  return result;
+}
+
+/// \brief Calculates factors |k| and |b| of a line using linear least squares method.
+/// \returns false in case of error (e.g. if the line is parallel with ordinate axis)
+/// and true otherwise.
+bool LinearLeastSquaresFactors(vector<double> const & xs, TAltitudeVec const & ys, double & k, double & b)
```
ygorshenin: vector<double> ys. No need to pass TAlititudeVec here.

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,47 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+{
+  long double const distSum = accumulate(xs.begin(), xs.end(), 0.0);
```
ygorshenin: Please, get rid of this domain-specific names like dist, alt, etc. All what you need is to compute mean of xs, mean of (xs * ys), etc ...

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,47 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+  uint64_t const altSum = accumulate(ys.begin(), ys.end(), 0);
+  long double const distSquareSum = accumulate(xs.begin(), xs.end(), 0.0,
+                                               [](double x, double y) { return x + y * y; });
+
+  if (distSum == 0.0 || distSquareSum == 0.0)
+    return false;
+
+  uint32_t const numPoints = xs.size();
```
ygorshenin: s/numPoints/n/g

generator/feature_segments_checker/feature_segments_checker.cpp
ygorshenin: I suggest you to use:
```mx``` for mean of xs
```mxy``` for mean of xs and ys.

Comment (ygorshenin) [inline generator/feature_segments_checker/feature_segments_checker.cpp]: Etc.

File: generator/feature_segments_checker/feature_segments_checker.cpp [hunk 9]
```diff
@@ -75,14 +79,47 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
 {
-  if (featureLengthMeters == 0.0)
-    return 0;
-  double const k = (endAltitudeMeters - startAltitudeMeters) / featureLengthMeters;
-  return startAltitudeMeters + k * distFromStartMeters;
+  return k * distFromStartMeters + b;
+}
+
+template <class T>
+long double SumWithDevisor(vector<T> const & vec, long double devisor)
```
ygorshenin: s/Devisor/Divisor/
You use this method only once. Do you really need it?

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,47 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+  long double const distSquareSum = accumulate(xs.begin(), xs.end(), 0.0,
+                                               [](double x, double y) { return x + y * y; });
+
+  if (distSum == 0.0 || distSquareSum == 0.0)
+    return false;
+
+  uint32_t const numPoints = xs.size();
+  long double const yDivX = SumWithDevisor(ys, distSum);
+  vector<double> xMulY(xs.size());
+  transform(xs.begin(), xs.end(), ys.begin(), xMulY.begin(), [](double x, double y){ return x*y; });
```
ygorshenin: Why do you need to create an aux vector here? Why not to simply iterate over it and calculate dot product?

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: > s/Devisor/Divisor/

Yes. It was an error.

> You use this method only once. Do you really need it?

I use it 3 times.

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,47 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+  long double const multDivXSquare = SumWithDevisor(xMulY, distSquareSum);
+  long double const xDivXSquare = SumWithDevisor(xs, distSquareSum);
```
ygorshenin: This formula is quite complex. Please, simplify it.

generator/feature_segments_checker/feature_segments_checker.cpp
ygorshenin: So, what's the point in this method? Why do you need it at all?

ygorshenin: Why not to create a method, say, Sum(), that will be a nice and readable shortcut for accumulate, and everywhere where you need to divide the sum of some vector by some value, directly divide them?

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Done

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: I got rid of domain specific names.

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Done

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: I need the aux vector with dot product to be able to use SumWithDivisor then. 

> Why not to simply iterate over it and calculate dot product?

What do you offer? To do something like that:
```c++
long double dotDivXSquare = 0.0;
for (int i = 0; i < xs.size(); ++i)
{
  dotDivXSquare += xs[i] * ys[i] / xSquareSum;
}
```
?

bykoianko:
@ygorshenin @kshalnev PTAL

kshalnev:
LGTM

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,47 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
-  return startAltitudeMeters + k * distFromStartMeters;
+  return k * distFromStartMeters + b;
+}
+
+template <class T>
+long double SumWithDivisor(vector<T> const & vec, long double divisor)
+{
+  long double result = 0.0;
+  for (auto item : vec)
+    result += item/divisor;
```
kshalnev: в конце поделить не быстрее? и для наглядности еще и суммирование на accmulate поменять?

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,47 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+  long double const ySum = accumulate(ys.begin(), ys.end(), 0);
+  long double const xSquareSum = accumulate(xs.begin(), xs.end(), 0.0,
+                                            [](double x, double y) { return x + y * y; });
+
+  if (xSum == 0.0 || xSquareSum == 0.0)
+    return false;
+
+  uint32_t const n = xs.size();
+  long double const yDivX = SumWithDivisor(ys, xSum);
+  vector<double> xDotY(xs.size());
```
kshalnev: была же красивая функция O(1) по памяти

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,47 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+  for (auto item : vec)
+    result += item/divisor;
+  return result;
+}
+
+/// \brief Calculates factors |k| and |b| of a line using linear least squares method.
+/// \returns false in case of error (e.g. if the line is parallel with ordinate axis)
+/// and true otherwise.
+bool LinearLeastSquaresFactors(vector<double> const & xs, vector<double> const & ys, double & k, double & b)
+{
```
ygorshenin:
```
mx = 0, my = 0, mxy = 0, mx2 = 0;
for (int i = 0; i < n; ++i) {
  mx += x[i];
  my += y[i];
  mxy += x[i] * y[i];
  mx2 += x[i] * x[i];
}
mx /= n;
my /= n;
mxy /= n;
mx2 /= n;

k = (mxy - mx * my) / (mx2 - mx * mx);
b = my - k * mx;
```

ygorshenin: Much easier, right? Coincides with https://en.wikipedia.org/wiki/Simple_linear_regression and most statistics literature.

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,47 @@ void WriteCSV(Cont const & cont, string const & fileName)
     fout << a.first << "," << a.second << endl;
 }
 
-/// \returns expected point altitude in meters according to linear model.
-double ExpectedPointAltitude(int16_t startAltitudeMeters, int16_t endAltitudeMeters,
-                             double distFromStartMeters, double featureLengthMeters)
+/// \returns y = k * x + b. It's an expected altitude in meters.
+double ExpectedPointAltitude(double k, double b, double distFromStartMeters)
```
ygorshenin: double GetY(double k, double x, double b)
{
}

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Быстрее и нагляднее, но тогда result достигает больших значений, что отрицательно сказывается на точности.

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Done

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Ok. Now I got the idea. But your code should be changed. It'll be something like that:
```c++
bool LinearLeastSquaresFactors(vector<double> const & xs, vector<double> const & ys, double & k, double & b)
{
  size_t const n = xs.size();
  long double mx = 0, my = 0, mxy = 0, mx2 = 0;
  for (int i = 0; i < n; ++i)
  {
    mx += xs[i];
    my += ys[i];
    mxy += xs[i] * ys[i];
    mx2 += xs[i] * xs[i];
  }
  mx /= n;
  my /= n;
  mxy /= n;
  mx2 /= n;

  if (mx == 0.0 || (mx - mx2/mx) == 0.0)
    return false;

  k = (my - mxy/mx) / (mx - mx2/mx);
  b = my - k*mx;
  return true;
}
```

bykoianko: Ok. I see. Your formulas are the same but without extra divisions.

bykoianko:
@ygorshenin PTAL

generator/feature_segments_checker/feature_segments_checker.cpp
ygorshenin: Sorry, but I didn't get the point why you modified the expression for k.

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,37 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+  }
+  mx /= n;
+  my /= n;
+  mxy /= n;
+  mx2 /= n;
+
+  if (mx == 0.0 || (mx - mx2/mx) == 0.0)
+    return false;
+
+  k = (my*mx - mxy) / (mx*mx - mx2);
```
ygorshenin: Clang format

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,37 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+  return k * d + b;
+}
+
+/// \brief Calculates factors |k| and |b| of a line using linear least squares method.
+/// \returns false in case of error (e.g. if the line is parallel with ordinate axis)
+/// and true otherwise.
+bool LinearLeastSquaresFactors(vector<double> const & xs, vector<double> const & ys, double & k, double & b)
+{
+  size_t const n = xs.size();
+  long double mx = 0, my = 0, mxy = 0, mx2 = 0;
```
ygorshenin: Please, don't use long double.

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,37 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+    my += ys[i];
+    mxy += xs[i] * ys[i];
+    mx2 += xs[i] * xs[i];
+  }
+  mx /= n;
+  my /= n;
+  mxy /= n;
+  mx2 /= n;
+
+  if (mx == 0.0 || (mx - mx2/mx) == 0.0)
```
ygorshenin: You only need to check that mx * mx is almost equal to mx2.

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +79,37 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+  size_t const n = xs.size();
+  long double mx = 0, my = 0, mxy = 0, mx2 = 0;
+  for (int i = 0; i < n; ++i)
+  {
+    mx += xs[i];
+    my += ys[i];
+    mxy += xs[i] * ys[i];
+    mx2 += xs[i] * xs[i];
+  }
+  mx /= n;
```
ygorshenin: You can move this division to the loop. Minus four lines, better precision.

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: I didn't copypaste your code. I wrote(rewrote) it based on my formulas. Anyway if you have a look at the diff, you'll a variant like yours without extra divisions.

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Done

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Why? If to use double the entropy is bigger on MacOsX. 
2.39986 (double) instead of 2.39789 (long double).

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Ok

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Yes. But if I do so the entropy will be more.
For Moscow I get 2.40125 (- 4 lines) and 2.39789 (current code).

bykoianko:
@ygorshenin PTAL

generator/feature_segments_checker/feature_segments_checker.cpp
ygorshenin: So, you really think that error in second place makes the difference?

generator/feature_segments_checker/feature_segments_checker.cpp
ygorshenin: The variant I proposed is more precise. Again, the error is
```R
> abs(2.40125 - 2.39789)
[1] 0.00336
```

ygorshenin: Moreover, there's a variance in measurements, if you wish, try to estimate it. It's quite big, I bet that it's bigger than 0.003 )

ygorshenin: The covariance matrix for k and b is the inverse of
```c++
N, N * mx
N * mx, N * mx2
```
multiplied by 1 (1 meter).

ygorshenin: So, the variance for k and b are diagonal values of the matrix.

bykoianko:
I did as you'd offered but tell me pls why you don't like "long double"?
@ygorshenin PTAL

ygorshenin:
I don't like or dislike standard types - they're standard and I'm taking them as is. Please, tell me why you like "long double"? Could you please tell me why it's better and should be used?

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: I don't thing it's an important question. I guess double and long double suits for these case. And I asked you about advantage of long double.

ygorshenin: Sorry, but I don't see any benefits of long double over double. sizeof(long
double) is unspecified and varies on different platforms, therefore results
computed with long double are not representative and reproducible.
Moreover, long double are usually slower than doubles, because of some
strange sizes, like 10 or 12 bytes. So, please, don't use long double just
because it's wider (sometimes).

ygorshenin:
LGTM

ygorshenin:
@mpimenov would you like to take a look?

bykoianko:
You suggested to exchange "long double" type with "double" type. I asked you about reasons for such request. Why do you thing long double is better in that case?

As for me long double gives us better exactness but takes more memory/time for execution. But anyway I do not thing this question is worth discussing. So I did as you had offered.

ygorshenin:
It sometimes gives you better precision, but this is very very platform
dependent. There is no IEEE standard for long double, for example. There is
no such types in other languages - Java, Go, Haskell, Lisp lack long
double, as it's quite arguable data type - sometimes it's precise and fast,
sometimes it's just a slow and bigger version of double.

To sum up and close the discussion, please, don't use long double unless
you know all pitfalls and benefits of this type and unless you are 100%
sure that you need it.

bykoianko:
Thanks.

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +80,32 @@ void WriteCSV(Cont const & cont, string const & fileName)
     fout << a.first << "," << a.second << endl;
 }
 
-/// \returns expected point altitude in meters according to linear model.
-double ExpectedPointAltitude(int16_t startAltitudeMeters, int16_t endAltitudeMeters,
-                             double distFromStartMeters, double featureLengthMeters)
+/// \returns y = k * x + b. It's an expected altitude in meters.
+double GetY(double k, double b, double d) { return k * d + b; }
```
mpimenov: s/d/x/ ?

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +80,32 @@ void WriteCSV(Cont const & cont, string const & fileName)
     fout << a.first << "," << a.second << endl;
 }
 
-/// \returns expected point altitude in meters according to linear model.
-double ExpectedPointAltitude(int16_t startAltitudeMeters, int16_t endAltitudeMeters,
-                             double distFromStartMeters, double featureLengthMeters)
+/// \returns y = k * x + b. It's an expected altitude in meters.
```
mpimenov: s/an/the/

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -75,14 +80,32 @@ void WriteCSV(Cont const & cont, string const & fileName)
...
+double GetY(double k, double b, double d) { return k * d + b; }
+
+/// \brief Calculates factors |k| and |b| of a line using linear least squares method.
+/// \returns false in case of error (e.g. if the line is parallel with ordinate axis)
```
mpimenov: s/ordinate/vertical/
s/parallel with/parallel to/

mpimenov: и артикль

generator/feature_segments_checker/feature_segments_checker.cpp
```diff
@@ -212,20 +267,29 @@ class Processor
...
+    {
+      double k;
+      double b;
+      vector<double> const pointAltitudesMeters(pointAltitudes.begin(), pointAltitudes.end());
+      if (!LinearLeastSquaresFactors(pointDists, pointAltitudesMeters, k, b))
+        return;
+
+      for (uint32_t i = 0; i < numPoints; ++i)
+      {
+        int32_t const deviation =
```
mpimenov: TAltitude?
Кстати, имеет какое-то значение, что TAltitude это int16_t, а здесь int32_t?

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: yes

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Done

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Done

generator/feature_segments_checker/feature_segments_checker.cpp
bykoianko: Можно TAltitude. deviation - это отклонение от прямых точек фичи. Не думаю, что оно будет больше TAltitude.

bykoianko:
@mpimenov PTAL&Merge

mpimenov:
LGTM

PR #3532, opened 2016-06-10
Fix for some integration tests:

therearesomewhocallmetim:
* Added storage integration tests to those that require temporary forlders for running normally;
* Added logging the process id for the shells running our tests (needed when a test hangs);


syershov:
LGTM

PR #3645, opened 2016-06-22
[strings] Move types translations from strings.txt to categories.txt

Zverik:
Промахнулись чутка.


Zverik:
@therearesomewhocallmetim

syershov:
LGTM

PR #3679, opened 2016-06-28
[bicycle] Process cycleway:left/:right

### LINKED ISSUE #3678: Bicycle mode routing makes an odd decision
Currently the beta calculates a long route with many (absurd) turns when starting roughly [here](https://www.openstreetmap.org/node/3272433826) and going [here](https://www.openstreetmap.org/node/3397947279)
https://jmuehldorfer.de/owncloud/apps/gpxpod/publink?filepath=/Dateien/GPSTracks/current.gpx&user=jmuehldorfer
The obvious answer is just to follow the street, which is correctly tagged with "cycleway:right=lane"
https://jmuehldorfer.de/owncloud/apps/gpxpod/publink?filepath=/Dateien/GPSTracks/expexted.gpx&user=jmuehldorfer

Zverik:
Велодорожка иногда только с одной стороны дороги — всё равно по ней ведём, пусть и без учёта стороны.

Will fix  #3678.

### DIFF
--- a/generator/osm2type.cpp
+++ b/generator/osm2type.cpp
@@ -529,6 +529,8 @@ namespace ftype
           { "bicycle", "!", [&params] { params.AddType(types.Get(CachedTypes::NOBICYCLE)); }},
           { "bicycle", "~", [&params] { params.AddType(types.Get(CachedTypes::YESBICYCLE)); }},
           { "cycleway", "~", [&params] { params.AddType(types.Get(CachedTypes::YESBICYCLE)); }},
+          { "cycleway:right", "~", [&params] { params.AddType(types.Get(CachedTypes::YESBICYCLE)); }},
+          { "cycleway:left", "~", [&params] { params.AddType(types.Get(CachedTypes::YESBICYCLE)); }},
           { "oneway:bicycle", "!", [&params] { params.AddType(types.Get(CachedTypes::BICYCLE_BIDIR)); }},
           { "cycleway", "opposite", [&params] { params.AddType(types.Get(CachedTypes::BICYCLE_BIDIR)); }},
         });


syershov:
LGTM

generator/osm2type.cpp
```diff
@@ -529,6 +529,8 @@ namespace ftype
           { "bicycle", "!", [&params] { params.AddType(types.Get(CachedTypes::NOBICYCLE)); }},
           { "bicycle", "~", [&params] { params.AddType(types.Get(CachedTypes::YESBICYCLE)); }},
           { "cycleway", "~", [&params] { params.AddType(types.Get(CachedTypes::YESBICYCLE)); }},
+          { "cycleway:right", "~", [&params] { params.AddType(types.Get(CachedTypes::YESBICYCLE)); }},
```
ygorshenin: Maybe it worth to use these tags to detect orientation of the road? I mean, cycleway:left implies that the road is oneway, right?

utack: Not sure I understand if you mean the road for cars is oneway, but that is not the case
You can for example add
cycleway:left=track
cycleway:right=lane
if the bicycles have some space on the right side of the line in OSM and a dedicated separated way on the left side of the OSM way, but that does not imply anything for the direction cars can go

ygorshenin: Agree, that does not imply anything for cars, but if cycleway:right is set, does it mean that the highway has a dedicated space on the right side, and, so, bicyclers may move only in a single direction on this space?

therearesomewhocallmetim:
FORMAT JTTP

generator/osm2type.cpp
Zverik: @ygorshenin you are right, the wiki states: "It should then be assumed that cycle traffic is allowed to flow in the customary direction for traffic on that side of the road". The same with `cycleway=opposite_lane`, which is even more complicated. But the trouble is, to correctly use this data in routing, we need to start storing and using driving side hints, which we don't have now. Otherwise `cycleway:left=lane` does not tell us in which direction that lane goes.

So I propose to leave this as is. @bykoianko?

ygorshenin:
LGTM

PR #3690, opened 2016-06-29
[geometry] Fixed a warning message.

mpimenov:
It is already annoying as is, let it at least pretend to be
grammatically correct.


ygorshenin:
LGTM

PR #3739, opened 2016-07-07
[search] Moved Results from Engine to Ranker.

mpimenov:


search/processor.cpp
```diff
@@ -580,8 +668,10 @@ void Processor::InitParams(QueryParams & params)
     params.m_langs.insert(GetLanguage(i));
 }
 
-void Processor::InitGeocoderParams(Geocoder::Params & params, bool viewportSearch)
+void Processor::InitGeocoderParams(Geocoder::Params & params)
```
ygorshenin: InitGeocoderParams not only inits params, but also inits geocoder. We need to think about new method name.

mpimenov: InitGeocoder?

search/processor.cpp
```diff
@@ -591,21 +681,30 @@ void Processor::InitGeocoderParams(Geocoder::Params & params, bool viewportSearc
...
   params.m_currentLocaleCode = m_currentLocaleCode;
-  if (m_mode == Mode::Viewport)
+  if (viewportSearch)
+  {
     params.m_viewport = GetViewport();
+    params.m_limit = kResultsCount;
+  }
+  else
+  {
+    params.m_limit = numeric_limits<size_t>::max();
```
ygorshenin: Why? I guess we should use kResultsCount here too, right?

search/processor.cpp
```diff
@@ -591,21 +681,30 @@ void Processor::InitGeocoderParams(Geocoder::Params & params, bool viewportSearc
...
+  size_t const kResultsCount = 30;
+  bool const viewportSearch = m_mode == Mode::Viewport;
   Ranker::Params params;
 
   params.m_currentLocaleCode = m_currentLocaleCode;
-  if (m_mode == Mode::Viewport)
+  if (viewportSearch)
+  {
     params.m_viewport = GetViewport();
+    params.m_limit = kResultsCount;
```
ygorshenin: I'm afraid that for viewport there should be 200?

mpimenov: This is completely rotten. See FlushViewportResults in ranker.cpp: the limit is not used in viewport search; the effective limit is 200 because that's the limit used in PreRanker.

search/processor.cpp
mpimenov: Now that's a valid point: the cases should be swapped.

search/processor.cpp
mpimenov: Changed to kPreResultsCount (200) here.

search/processor.cpp
mpimenov: done

search/processor.cpp
ygorshenin: Sounds good.

mpimenov:
PTAL

search/processor.cpp
mpimenov: done

search/ranker.hpp
```diff
@@ -62,6 +64,9 @@ class Ranker
     m2::PointD m_accuratePivotCenter = m2::PointD(0, 0);
 
     TLocales m_categoryLocales;
+
+    size_t m_limit;
```
ygorshenin: Please, zero-initialize this.

mpimenov: done

ygorshenin:
LGTM

ygorshenin:
DUDE JTTP

PR #3773, opened 2016-07-13
Correct danish for amenity=shelter

Hjart:
Danish translation of amenity=shelter was misleading. It was translated to "home" which lead to quite a few amenity=shelter's spuriously located on top of houses and other weird places. This has puzzled me for quite some time now...


Zverik:
Thanks!

PR #3790, opened 2016-07-14
[search] Fixed slooow locality matching.

ygorshenin:
Locality matching code haven't supported a lot of mwms, and iterated
over the whole mwms set on each search result (sic!).  Moreover, for
locality selection it directly read all features in the rect instead of
RankTable usage.

This CL significantly speeds up search results generation stage, when
there're a lot of maps and a lot of search results.


ygorshenin:
@mpimenov PTAL
@vng, if you read this, PTAL

mpimenov:
LGTM

PR #3979, opened 2016-08-08
fix implicit cast compilation for archlinux with clang++-3.8

fijemax:
Add explicit cast on number in max function under archlinux with clang++-3.8.
This implicit cast causes this compilation error :
```
/home/fred/mapotempo/code/maps.me/omim/drape_frontend/text_layout.cpp:563:30: error: no matching function for call to 'max'
   double const textCount = max(floor(pathLength / minPeriodSize), 1.0);
                            ^~~
/usr/bin/../include/c++/v1/algorithm:2635:1: note: candidate template ignored: deduced conflicting types for parameter '_Tp'
     ('float' vs. 'double')
max(const _Tp& __a, const _Tp& __b)
```

### DIFF
--- a/drape_frontend/text_layout.cpp
+++ b/drape_frontend/text_layout.cpp
@@ -560,7 +560,7 @@ void PathTextLayout::CalculatePositions(vector<float> & offsets, float splineLen
   }
   else
   {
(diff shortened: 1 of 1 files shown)


fijemax:
issue is gnu++11was set instead of c++11

PR #4013, opened 2016-08-10
[android] Whats'new, new.

yunikkk:


trashkalmar:
Надо бы удалить старые картинки и тексты.

rugk:
Just a question: Can't you just use SVGs instead of many, many PNG images?

trashkalmar:
@rugk yep. However in this case SVGs will be too large.

rugk:
SVGs can scale to any size or do you mean the file size?

trashkalmar:
Yes, file size. SVG are good for simple icons. But for complex images they are too huge and too slow.

rugk:
And you cannot [minify](https://github.com/svg/svgo) them, so that they are small enough?

trashkalmar:
Honestly, I do not know. Our designer may take a look. @igortomko

PR #4022, opened 2016-08-11
[ios][android] Exclude Default name from "Add language" window

milchakov:
@VladiMihaylenko 
@mgsergio

PR #4027, opened 2016-08-11
[ios][android]Exclude Default name from "Add language" window

milchakov:


milchakov:
request into master https://github.com/mapsme/omim/pull/4022

PR #4029, opened 2016-08-11
Fixed indexer_tests on Master

therearesomewhocallmetim:
Swapped the first and the second sinonyms for shop-alcohol in the Danish language.


Zverik:
LGTM

PR #4043, opened 2016-08-12
Build type fix.

therearesomewhocallmetim:
The names of some permissions differ depending on the build type, such as beta, release or debug, plus we have some permissions in some build types that are not present in the others.


syershov:
LGTM

Zverik:
LGTM

PR #4064, opened 2016-08-16
[android] Fixed crash.

yunikkk:


android/src/com/mapswithme/maps/location/LocationHelper.java
```diff
@@ -209,6 +209,11 @@ public void run()
   {
     mLogger.d("ctor()");
 
+    // TODO consider refactoring.
+    // Actually we shouldn't initialize Framework here,
+    // to allow app components to retrieve location updates without all heavy framework's stuff initialized.
+    // For now this is necessary to connect mModeChangeListener below.
+    MwmApplication.get().initNativeCore();
```
ygorshenin: How this may happen, that LocationHelper is initialized but framework is not?

ygorshenin: And why not to just skip nativeSetListener if framework is not initialized?

trashkalmar: Framework is initialized (in general) much later than other components. And here `nativeSetListener()` is the crucial operation for this class.

ygorshenin: As far as I know framework is not needed when track recording is enabled and application is killed.

ygorshenin: I.e. when track recording works in background, in daemon mode, framework is not needed, right?

mpimenov:
Merged in #4066.

android/src/com/mapswithme/maps/location/LocationHelper.java
trashkalmar: It seems framework required anyway. At least to `nativeSetListener()`

yunikkk: Yep, now it's required at least to set that listener. Left a TODO to refactor that.

PR #4076, opened 2016-08-17
[ios] Fix interactive search on Ipad.

mpimenov:


mpimenov:
@igrechuhin PTAL

singalen:
Can you please tell, what exactly does this one fix?
For me, iPad mini is totally unable to find anything by address. On a first launch, suggestions appear as I type, but then they vanish and never appear again.
Is this that bug, or should I register a new one?
Thanks.

PR #4117, opened 2016-08-22
[doc] Update routing and non routing index on MAPS.md

frodrigo:
I keep the Steps part of MAPS.md unchanged, as I not sure of current steps order (step 6).

### DIFF
--- a/docs/MAPS.md
+++ b/docs/MAPS.md
@@ -1,7 +1,7 @@
 # MWM Files
 
 MAPS.ME uses maps in its own vector format, MWM. It contains classified features sorted and simplified by zoom level.
-For car routing, it needs a separate routing index in a `.mwm.routing` file. We build maps for the entire planet:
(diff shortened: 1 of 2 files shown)


docs/MAPS.md
```diff
@@ -1,7 +1,7 @@
 # MWM Files
 
 MAPS.ME uses maps in its own vector format, MWM. It contains classified features sorted and simplified by zoom level.
-For car routing, it needs a separate routing index in a `.mwm.routing` file. We build maps for the entire planet:
+We build maps for the entire planet:
```
Zverik: I think the removed part should be replaced with "It also can include a pre-calculated routing index for car routing."

Zverik:
Step 6 should also be rewritten to "building routing indices out of `*.osrm*` files".

frodrigo:
Requested changes done

Zverik:
Thanks!

PR #4131, opened 2016-08-23
[Android] Adding a splash screen on the application start.

fijemax:
- Creating the empty class "SplashActivity" who extends "DownloadResourcesActivity". Created if some work need to do here in future.
 - Adding a new style to define the splash screen.


trashkalmar:
This actually not needed. If we need splash, we'd not extend it from DownloadRecourcesActivity as it is completely different component.

PR #4150, opened 2016-08-25
Sergey's fix for tests

therearesomewhocallmetim:


therearesomewhocallmetim:
DUDE JTTP

storage/country_tree.hpp
```diff
@@ -153,7 +153,7 @@ class CountryTree
   };
 
 private:
-  using TCountryTreeHashTable = unordered_multimap<TKey, Node *>;
+  using TCountryTreeHashTable = multimap<TKey, Node *>;
```
ygorshenin: Why not to modify instead tests? I.e. to check equality of sets, not lists?

therearesomewhocallmetim:
"Why not modify tests instead?" - First, this modification has already been merged into the release branch. We will have conflicts if we don't merge this one (even if we then go on to modifying the tests). Second, we would have to modify three tests instead of 1 line in the code. Third, we need to ask Sergey about it :)

@syershov Please, help with the discussion

ygorshenin:
Ah, this is already in release branch...
Okay, LGTM...

PR #4158, opened 2016-08-26
Td locs 160826

therearesomewhocallmetim:
Three Russian translations replaced.

### DIFF
--- a/strings.txt
+++ b/strings.txt
@@ -4,7 +4,7 @@
     comment = About button text (should be short)
     tags = tizen
     en = About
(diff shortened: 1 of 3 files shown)


mpimenov:
LGTM

strings.txt
```diff
@@ -4,7 +4,7 @@
     comment = About button text (should be short)
     tags = tizen
     en = About
-    ru = О программе
+    ru = О приложении
```
mpimenov: The tag "tizen" is probably wrong, though.

PR #4175, opened 2016-08-29
[android, search] Removed unnecessary synchronization.

ygorshenin:
Get rid of sync in Android - as all calls to core comes from the UI thread, and as all callbacks from SearchEngine are posted to the UI thread.


ygorshenin:
@mpimenov and ( @trashkalmar or @yunikkk ) PTAL

trashkalmar:
LGTM

mpimenov:
LGTM

PR #4193, opened 2016-09-01
Generating route altitude chart image.

bykoianko:
Генерация диаграммы изменения высоты маршрута. В ядре. agg. 
Тесты.

@rokuz @ygorshenin @igrechuhin  PTAL


map/framework.cpp
```diff
@@ -2195,6 +2196,9 @@ void Framework::BuildRoute(m2::PointD const & start, m2::PointD const & finish,
     storage::TCountriesVec absentCountries;
     if (code == IRouter::NoError)
     {
+//      vector<uint8_t> frameBuffer;
+//      GenerateRouteAltitudeChart(400, 100, frameBuffer);
+
```
rokuz: Может убрать отсюда закомментированный код?

map/framework.cpp
```diff
@@ -2979,3 +2983,18 @@ bool Framework::OriginalFeatureHasDefaultName(FeatureID const & fid) const
 {
   return osm::Editor::Instance().OriginalFeatureHasDefaultName(fid);
 }
+
+bool Framework::HasRouteAltitude() const
+{
+  return !m_routingSession.HasRouteAltitude();
```
ygorshenin: Sorry? Is it intended? The name of method diverges with its body.

map/generate_chart.cpp
```diff
@@ -0,0 +1,127 @@
+#include "map/generate_chart.hpp"
+
+//#include "coding/png_memory_encoder.hpp"
+
+#include "base/assert.hpp"
+
+#include "std/fstream.hpp"
```
ygorshenin: std before base, please.

map/generate_chart.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include "3party/agg/agg_path_storage.h"
+#include "3party/agg/agg_pixfmt_rgba.h"
+#include "3party/agg/agg_rasterizer_scanline_aa.h"
+#include "3party/agg/agg_renderer_primitives.h"
+#include "3party/agg/agg_renderer_scanline.h"
+#include "3party/agg/agg_scanline_p.h"
+
+namespace
+{
+template <class TColor, class TOrder>
```
ygorshenin: We concluded (@ygorshenin, @mpimenov, @syershov, @mgsergio) that these "T" prefixes are ugly. Join us and get rid of them!

map/generate_chart.cpp
```diff
@@ -0,0 +1,127 @@
...
+struct TBlendAdaptor
```
ygorshenin: s/TBlendAdaptor/BlendAdaptor/

map/generate_chart.cpp
```diff
@@ -0,0 +1,127 @@
...
+{
+  using order_type = TOrder;
```
ygorshenin: Please, use our naming conventions.

map/generate_chart.cpp
```diff
@@ -0,0 +1,127 @@
...
+  using color_type = TColor;
+  using TValueType = typename color_type::value_type;
+  using TCalcType = typename color_type::calc_type;
+
+  enum EBaseScale
```
ygorshenin: Please, use our naming rules, not agg's.

map/generate_chart.cpp
```diff
@@ -0,0 +1,127 @@
...
+                   vector<uint8_t> & frameBuffer)
+{
+  agg::rgba8 const kBackgroundColor(255, 255, 255, 255);
+  agg::rgba8 const kLineColor(0, 100, 100, 255);
+  double constexpr kLineWidth = 2;
+  uint32_t constexpr kBPP = 4;
+
+  if (width == 0 || height == 0 || geometry.empty())
+  {
+    ASSERT(false, ());
```
ygorshenin: What's the problem?

map/generate_chart.cpp
```diff
@@ -0,0 +1,127 @@
...
+  using TPixelFormat = agg::pixfmt_custom_blend_rgba<TBlender, agg::rendering_buffer>;
+  using TBaseRenderer = agg::renderer_base<TPixelFormat>;
+  using TPrimitivesRenderer = agg::renderer_primitives<TBaseRenderer>;
+  using TSolidRenderer = agg::renderer_scanline_aa_solid<TBaseRenderer>;
+
+  agg::rendering_buffer renderBuffer;
+  TPixelFormat pixelFormat(renderBuffer, agg::comp_op_src_over);
+  TBaseRenderer m_baseRenderer(pixelFormat);
+
+  frameBuffer.resize(width * kBPP * height);
```
ygorshenin:
```c++
frameBuffer.assign(width * kBPP * height, 0);
```

map/generate_chart.cpp
```diff
@@ -0,0 +1,127 @@
...
+//  file.write(reinterpret_cast<char *>(output.data()), output.size());
+  return true;
+}
+
+bool GenerateChart(size_t width, size_t height, feature::TAltitudes const & yAxisNormalizedData,
+                   vector<uint8_t> & frameBuffer)
+{
+  uint32_t constexpr kHeightIndent = 10;
+
+  auto const minMaxAltitudeIt = minmax_element(yAxisNormalizedData.begin(), yAxisNormalizedData.end());
```
ygorshenin: Please, clear frameBuffer and return true if yAxisNormalizeddata is empty.

map/generate_chart.cpp
```diff
@@ -0,0 +1,127 @@
...
+  double const deltaAlt = maxAlt - minAlt;
+  double const oneSegLenPix = static_cast<double>(width) / yAxisNormalizedData.size();
+
+  vector<m2::PointD> geometry(yAxisNormalizedData.size());
+  for (size_t i = 0; i < yAxisNormalizedData.size(); ++i)
+  {
+    geometry[i] = m2::PointD(i * oneSegLenPix,
+                             kHeightIndent + static_cast<double>(height - 2 * kHeightIndent) * (yAxisNormalizedData[i] - minAlt) /
+                             deltaAlt);
+    geometry[i].y = height - geometry[i].y;
```
ygorshenin: Why not to include "height - " to the formula that computes geometry[i] for the first time?

map/generate_chart.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "indexer/feature_altitude.hpp"
+
+#include "geometry/point2d.hpp"
+
+#include "std/vector.hpp"
+
+void NormalizeChartData(vector<double> const & xAxisData, vector<double> const & yAxisData,
```
ygorshenin: Where this method is defined?

routing/road_graph_router.cpp
```diff
@@ -247,10 +249,15 @@ void RoadGraphRouter::ReconstructRoute(vector<Junction> && path, Route & route,
...
 
+  feature::TAltitudes altitudes(path.size());
+  for (size_t i = 0; i < path.size(); ++i)
+    altitudes[i] = path[i].GetAltitude();
+
   route.SetGeometry(geometry.begin(), geometry.end());
   route.SetSectionTimes(times);
   route.SetTurnInstructions(turnsDir);
   route.SetStreetNames(streetNames);
+  route.SetAltitudes(altitudes);
```
ygorshenin: I guess it's not ok that Set* method modifies its argument. Rename to SwapAltitudes or pass altitudes by rvalue-reference.

routing/routing_session.cpp
```diff
@@ -523,6 +523,26 @@ void RoutingSession::EmitCloseRoutingEvent() const
...
+  return !m_route.GetAltitudes().empty();
+}
+
+bool RoutingSession::HasRouteAltitude() const
+{
+  threads::MutexGuard guard(m_routeSessionMutex);
+  return HasRouteAltitudeImpl();
+}
+
+bool RoutingSession::CopyRouteAltitudes(feature::TAltitudes & routeAltitudes) const
```
ygorshenin: s/CopyRouteAltitudes/GetRouteAltitudes/g

map/framework.cpp
bykoianko: Конечно уберу, перед тем как сниму пометку DNM

map/framework.cpp
bykoianko: It's an error. Thanks.

map/generate_chart.cpp
bykoianko: Are you sure? We usually place base before std.

map/generate_chart.cpp
bykoianko: No problem. Let's remove all these T(s)

map/generate_chart.cpp
bykoianko: Done

map/generate_chart.cpp
bykoianko: I can not. BlendAdaptor is used in other agg template:
agg::pixfmt_custom_blend_rgba<BlendAdaptor<agg::rgba8, agg::order_rgba>, agg::rendering_buffer>
and agg::pixfmt_custom_blend_rgba wants BlendAdaptor to have types with names order_type and color_type

map/generate_chart.cpp
bykoianko: Done.

map/generate_chart.cpp
bykoianko: I guess there's no sense to generate an empty image or generate a chart without geometry. GenerateChart(...) is called with such params it's a sign that something goes wrong.

map/generate_chart.cpp
bykoianko: Yes!

map/generate_chart.cpp
bykoianko: Although, I removed these ASSERTs

map/generate_chart.cpp
bykoianko: Done.

map/generate_chart.hpp
bykoianko: I'll be implemented later in this PR before it's ready for merge.

map/generate_chart.cpp
bykoianko: Done

routing/road_graph_router.cpp
bykoianko: I agree. But it's not enough to rename SetAltitude only. I renamed all the group of such methods.

routing/routing_session.cpp
bykoianko: I don't like the method looks like a light weight method. If I call it GetRouteAltitudes() users could consider it as a quick getter but it copies all the content of vector with significant length.

map/generate_chart.cpp
ygorshenin: Sorry, you're right.

routing/routing_session.cpp
ygorshenin: Getters may not be lightweight. Moreover, the method signature indicates that it's not lightweight.

routing/road_graph_router.cpp
ygorshenin: Wait, wait! You decided to rename methods instead of passing vector by rvalue-reference?

routing/routing_session.cpp
bykoianko: Ok.

bykoianko:
@rokuz @ygorshenin @igrechuhin  PTAL

map/framework.cpp
```diff
@@ -2979,3 +2981,26 @@ bool Framework::OriginalFeatureHasDefaultName(FeatureID const & fid) const
 {
   return osm::Editor::Instance().OriginalFeatureHasDefaultName(fid);
 }
+
+bool Framework::HasRouteAltitude() const { return m_routingSession.HasRouteAltitude(); }
+bool Framework::GenerateRouteAltitudeChart(size_t width, size_t height, bool day,
```
rokuz: Пустая строка между?

bykoianko: Сделал. (clang format ее удалил ранее)

map/generate_chart.cpp
```diff
@@ -0,0 +1,215 @@
...
+    heightIndent = 0;
+  }
+
+  auto const minMaxAltitudeIt = minmax_element(altitudeDataM.begin(), altitudeDataM.end());
+  double const minAltM = *minMaxAltitudeIt.first;
+  double const maxAltM = *minMaxAltitudeIt.second;
+  double const deltaAltM = maxAltM - minAltM;
+  uint32_t const drawHeightPxl = height - 2 * heightIndent;
+  double const meterPerPxl = max(minMetersPerPxl, deltaAltM / static_cast<double>(drawHeightPxl));
+  if (meterPerPxl == 0.0)
```
rokuz: Такое сравнение здесь норм?

bykoianko: Думаю, да. Пока minMetersPerPxl задан корректно - meterPerPxl будет далеко не 0. Чуть позже мы делим на meterPerPxl. Так что я сделал на всякий случай, если кто-то позже поменяет minMetersPerPxl не подумав. В этом случае дебаг будет падать по LOG(LERROR, ...)

rokuz:
LGTM

map/framework.cpp
```diff
@@ -2979,3 +2981,26 @@ bool Framework::OriginalFeatureHasDefaultName(FeatureID const & fid) const
...
   return osm::Editor::Instance().OriginalFeatureHasDefaultName(fid);
 }
+
+bool Framework::HasRouteAltitude() const { return m_routingSession.HasRouteAltitude(); }
+
+bool Framework::GenerateRouteAltitudeChart(size_t width, size_t height, bool day,
+                                           vector<uint8_t> & imageRGBAData) const
+{
+  feature::TAltitudes altitudes;
+  if (!m_routingSession.GetRouteAltitudes(altitudes) && altitudes.empty())
```
kshalnev: GetRouteAltitudes returns false if altitude is empty, right?

bykoianko: You're right. I see your point. I'll fix it.

routing/routing_session.hpp
```diff
@@ -165,6 +175,8 @@ class RoutingSession
   void RemoveRoute();
   void RemoveRouteImpl();
 
+  bool HasRouteAltitudeImpl() const;
```
kshalnev: private function?

bykoianko: Yes. I'll do that.

map/generate_chart.hpp
```diff
@@ -0,0 +1,38 @@
+#pragma once
```
ygorshenin: The file should be named chart_generator, or not generate_chart.

routing/road_graph_router.cpp
bykoianko: Yes. Is that a problem?
Anyway It's not a big deal. There're very few places where these Set methods are used.

map/generate_chart.hpp
bykoianko: Done

bykoianko:
@ygorshenin @kshalnev PTAL

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+  if (!is_sorted(distanceDataM.cbegin(), distanceDataM.cend()))
+  {
+    LOG(LERROR, ("Route segment distances are not sorted."));
+    return;
+  }
+
+  uniformAltitudeDataM.resize(resultPointCount);
+  double const routeLenM = distanceDataM.back();
+  double const stepLenM = routeLenM / static_cast<double>(resultPointCount - 1);
```
kshalnev: what if resultPointCount == 1? may be then an assert or return nothing then?

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+    if (distanceDataM[prevPointIdx] == distanceDataM[nextPointIdx])
+      return static_cast<double>(altitudeDataM[prevPointIdx]);
+
+    double const k = (altitudeDataM[nextPointIdx] - altitudeDataM[prevPointIdx]) /
+                     (distanceDataM[nextPointIdx] - distanceDataM[prevPointIdx]);
+    return static_cast<double>(altitudeDataM[prevPointIdx]) +
+           k * (distFormStartM - distanceDataM[prevPointIdx]);
+  };
+
+  uniformAltitudeDataM.resize(resultPointCount);
```
kshalnev: uniformAltitudeDataM already has been resized above. this line is redundant.

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+  double const routeLenM = distanceDataM.back();
+  double const stepLenM = routeLenM / static_cast<double>(resultPointCount - 1);
+
+  auto const calculateAltitude = [&](double distFormStartM) {
+    if (distFormStartM <= distanceDataM.front())
+      return static_cast<double>(altitudeDataM.front());
+    if (distFormStartM >= distanceDataM.back())
+      return static_cast<double>(altitudeDataM.back());
+
+    auto const lowerIt = lower_bound(distanceDataM.cbegin(), distanceDataM.cend(), distFormStartM);
```
kshalnev: optional. you can speed up algorithm: since distanceFromStartM is monotonic growing, it should be enough to lower_bound(prevLowerId, distanceDataM.end()), where prevLowerId initially set to distanceDataM.begin().

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+void GenerateYAxisChartData(size_t height, double minMetersPerPxl,
+                            vector<double> const & altitudeDataM, vector<double> & yAxisDataPxl)
+{
+  uint32_t constexpr kHeightIndentPxl = 2.0;
+  yAxisDataPxl.clear();
+  uint32_t heightIndent = kHeightIndentPxl;
+  if (height <= 2 * kHeightIndentPxl)
+  {
+    LOG(LERROR, ("Chart height is less then 2 * kHeightIndentPxl (", 2 * kHeightIndentPxl, ")"));
+    heightIndent = 0;
```
kshalnev: may be return?

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+}
+
+void GenerateChartByPoints(size_t width, size_t height, vector<m2::PointD> const & geometry,
+                           bool day, vector<uint8_t> & frameBuffer)
+{
+  frameBuffer.clear();
+  if (width == 0 || height == 0 || geometry.empty())
+    return;
+
+  agg::rgba8 const kBackgroundColor = agg::rgba8(255, 255, 255, 0);
```
kshalnev: optional. May be add all such hardcoded colors in MapStyle class as getters? then such colors can be easily replaced by data from mapcss or something like that? At least all such data will be collected in one place, instead of spreading them across of project.

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+  using TPixelFormat = agg::pixfmt_custom_blend_rgba<TBlender, agg::rendering_buffer>;
+  using TBaseRenderer = agg::renderer_base<TPixelFormat>;
+  using TPrimitivesRenderer = agg::renderer_primitives<TBaseRenderer>;
+  using TSolidRenderer = agg::renderer_scanline_aa_solid<TBaseRenderer>;
+  using TPath = agg::poly_container_adaptor<vector<m2::PointD>>;
+  using TStroke = agg::conv_stroke<TPath>;
+
+  agg::rendering_buffer renderBuffer;
+  TPixelFormat pixelFormat(renderBuffer, agg::comp_op_src_over);
+  TBaseRenderer m_baseRenderer(pixelFormat);
```
kshalnev: m_baseRenderer - it is name for a class member, right? may be rename to just 'baseRenderer'?

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+  // Polygon under chart line.
+  agg::path_storage underChartGeometryPath;
+  underChartGeometryPath.move_to(geometry.front().x, static_cast<double>(height));
+  for (auto const & p : geometry)
+    underChartGeometryPath.line_to(p.x, p.y);
+  underChartGeometryPath.line_to(geometry.back().x, static_cast<double>(height));
+  underChartGeometryPath.close_polygon();
+
+  agg::conv_curve<agg::path_storage> curve(underChartGeometryPath);
```
kshalnev: optional. no any move semantic in AGG yet? i mean, can extra copying be avoided here  (or here is no copying?)

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+  rasterizer.add_path(stroke);
+  agg::render_scanlines_aa_solid(rasterizer, scanline, m_baseRenderer, kLineColor);
+}
+
+void GenerateChart(size_t width, size_t height, deque<double> const & distanceDataM,
+                   feature::TAltitudes const & altitudeDataM, bool day,
+                   vector<uint8_t> & frameBuffer)
+{
+  frameBuffer.clear();
+  if (altitudeDataM.empty() || distanceDataM.size() != altitudeDataM.size())
```
kshalnev: some of that is checked in NormalizeChartData. may be assert is enough in NormalizeChartData?

kshalnev: GenerateChart looks like a Public API while functions above looks like a private implementation helpers. May be move them into anonymous namespace then?

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+  vector<double> uniformAltitudeDataM;
+  NormalizeChartData(distanceDataM, altitudeDataM, width, uniformAltitudeDataM);
+  if (uniformAltitudeDataM.empty())
+    return;
+
+  vector<double> yAxisDataPxl;
+  GenerateYAxisChartData(height, 1.0 /* minMetersPerPxl */, uniformAltitudeDataM, yAxisDataPxl);
+
+  size_t const uniformAltitudeDataSz = yAxisDataPxl.size();
+  double const oneSegLenPix = static_cast<double>(width) / (uniformAltitudeDataSz - 1);
```
kshalnev: uniformAltitudeDataSz - can it be 1?

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+  vector<m2::PointD> geometry(uniformAltitudeDataSz);
+  for (size_t i = 0; i < uniformAltitudeDataSz; ++i)
+    geometry[i] = m2::PointD(i * oneSegLenPix, yAxisDataPxl[i]);
```
kshalnev: really need vector<m2::PointD>? you already have step - oneSegLenPix and bar heights - yAxisDataPxl. Seems it is enough to render chart and you can get rid of that vector (with allocations, copying, deallocations and other horrible unused things).

routing/route.hpp
```diff
@@ -52,21 +55,10 @@ class Route
...
-  {
-    swap(m_times, v);
-  }
-
-  inline void SetStreetNames(TStreets & v)
-  {
-    swap(m_streets, v);
-  }
-
+  inline void SwapTurnInstructions(TTurns & v) { swap(m_turns, v); }
```
kshalnev: optional. may be SetXXX with move semantic? It looks a bit more regular, imho.

map/chart_generator.cpp
rokuz: No way!
Никакого отношения к стилю карты эти цвета не имеют

kshalnev: Это смотря что понимать под "стилем карты", я понимаю под этим еще и цветовой стиль приложения. здесь как минимум уже есть зависимость цвета линии от day (дневной/ночной стиль карты?) что если к ним добавится завтра еще и набор стилей для слабовидящих людей? еще один bool протаскивать?

PS. предотвращая холивар - я предлагаю подумать как эти цвета могут безболезненно меняться. не сейчас, потом.

map/chart_generator.cpp
ygorshenin: When resultPointCount equals to 1 stepLenM should be set to zero. No need in assertions.

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+}  // namespace
+
+void NormalizeChartData(deque<double> const & distanceDataM,
+                        feature::TAltitudes const & altitudeDataM, size_t resultPointCount,
+                        vector<double> & uniformAltitudeDataM)
+{
+  uniformAltitudeDataM.clear();
+  if (distanceDataM.empty() || resultPointCount == 0 ||
+      distanceDataM.size() != altitudeDataM.size())
+    return;
```
ygorshenin: Curly braces, please.

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+      return static_cast<double>(altitudeDataM.back());
+
+    auto const lowerIt = lower_bound(distanceDataM.cbegin(), distanceDataM.cend(), distFormStartM);
+    if (lowerIt == distanceDataM.cbegin())
+      return static_cast<double>(altitudeDataM.front());
+    if (lowerIt == distanceDataM.cend())
+      return static_cast<double>(altitudeDataM.back());
+
+    size_t const nextPointIdx = distance(distanceDataM.cbegin(), lowerIt);
+    CHECK_LESS(0, nextPointIdx, ());
```
ygorshenin: Yoda condition?

map/chart_generator.cpp
rokuz: Стиль карты - это стиль карты, а не стиль приложения. График - это часть нативного UI для каждой платформы со своей цветовой схемой. Если куда-то эти цвета и выносить, то только в UI часть каждой из платформ.

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+    size_t const prevPointIdx = nextPointIdx - 1;
+
+    if (distanceDataM[prevPointIdx] == distanceDataM[nextPointIdx])
```
ygorshenin: Please, use AlmostEqualAbs instead of direct comparison. No need in linear approximations for short segments, say, shorter than 1e-6 meters.

map/chart_generator.cpp
ygorshenin: I guess it's better to leave this line as is and remove its twin above.

map/chart_generator.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include "geometry/point2d.hpp"
+
+#include "std/deque.hpp"
+#include "std/vector.hpp"
+
+/// \brief fills uniformAltitudeDataM with altitude data which evenly distributed by
+/// |resultPointCount| points. |distanceDataM| and |altitudeDataM| form a curve of route altitude.
+/// This method is used to generalize and evenly distribute points of the chart.
+void NormalizeChartData(deque<double> const & distanceDataM,
```
ygorshenin: Could you please explain why not to use vector?

map/chart_generator.hpp
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+#include "indexer/feature_altitude.hpp"
+
+#include "geometry/point2d.hpp"
+
+#include "std/deque.hpp"
+#include "std/vector.hpp"
+
+/// \brief fills uniformAltitudeDataM with altitude data which evenly distributed by
```
ygorshenin: Please, put all this API to the namespace map.

map/chart_generator.hpp
```diff
@@ -0,0 +1,38 @@
...
+/// |resultPointCount| points. |distanceDataM| and |altitudeDataM| form a curve of route altitude.
+/// This method is used to generalize and evenly distribute points of the chart.
+void NormalizeChartData(deque<double> const & distanceDataM,
+                        feature::TAltitudes const & altitudeDataM, size_t resultPointCount,
```
ygorshenin: #include "std/cstdint.hpp"

map/chart_generator.hpp
```diff
@@ -0,0 +1,38 @@
...
+/// \param yAxisDataPxl Y-axis data of altitude chart in pixels.
+void GenerateYAxisChartData(size_t height, double minMetersPerPxl,
+                            vector<double> const & altitudeDataM, vector<double> & yAxisDataPxl);
+
+/// \brief generates chart image on a canvas with size |width|, |height| with |geometry|.
+/// (0, 0) is a left-top corner. X-axis goes down and Y-axis goes right.
+/// \param width is result image width in pixels.
+/// \param height is result image height in pixels.
+/// \param geometry is points which is used to draw a curve of the chart.
+/// \param frameBuffer is a vector for a result image. It's resized in this method.
```
ygorshenin: Param |day| is missed.

map/chart_generator.cpp
kshalnev: да, пусть это будет "стиль приложения" из UI.

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+  uniformAltitudeDataM.resize(resultPointCount);
+  for (size_t i = 0; i < resultPointCount; ++i)
+    uniformAltitudeDataM[i] = calculateAltitude(static_cast<double>(i) * stepLenM);
+}
+
+void GenerateYAxisChartData(size_t height, double minMetersPerPxl,
+                            vector<double> const & altitudeDataM, vector<double> & yAxisDataPxl)
+{
+  uint32_t constexpr kHeightIndentPxl = 2.0;
+  yAxisDataPxl.clear();
```
ygorshenin: No need in clear as you do resize later.

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+  uniformAltitudeDataM.resize(resultPointCount);
+  for (size_t i = 0; i < resultPointCount; ++i)
+    uniformAltitudeDataM[i] = calculateAltitude(static_cast<double>(i) * stepLenM);
+}
+
+void GenerateYAxisChartData(size_t height, double minMetersPerPxl,
+                            vector<double> const & altitudeDataM, vector<double> & yAxisDataPxl)
+{
+  uint32_t constexpr kHeightIndentPxl = 2.0;
```
ygorshenin: Double to uint32_t constexpr?

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+  yAxisDataPxl.clear();
+  uint32_t heightIndent = kHeightIndentPxl;
+  if (height <= 2 * kHeightIndentPxl)
+  {
+    LOG(LERROR, ("Chart height is less then 2 * kHeightIndentPxl (", 2 * kHeightIndentPxl, ")"));
+    heightIndent = 0;
+  }
+
+  auto const minMaxAltitudeIt = minmax_element(altitudeDataM.begin(), altitudeDataM.end());
```
ygorshenin: altitudeDataM may be empty. Please, write a smoke test on this.

map/chart_generator.cpp
bykoianko: Поправил. Спасибо. Нет нужды в этом методе, если маршрут состоит из одной точки.

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+    return static_cast<double>(altitudeDataM[prevPointIdx]) +
+           k * (distFormStartM - distanceDataM[prevPointIdx]);
+  };
+
+  uniformAltitudeDataM.resize(resultPointCount);
+  for (size_t i = 0; i < resultPointCount; ++i)
+    uniformAltitudeDataM[i] = calculateAltitude(static_cast<double>(i) * stepLenM);
+}
+
+void GenerateYAxisChartData(size_t height, double minMetersPerPxl,
```
ygorshenin: Please, use uint32_t for height.

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+  double const deltaAltM = maxAltM - minAltM;
+  uint32_t const drawHeightPxl = height - 2 * heightIndent;
+  double const meterPerPxl = max(minMetersPerPxl, deltaAltM / static_cast<double>(drawHeightPxl));
+  if (meterPerPxl == 0.0)
+  {
+    LOG(LERROR, ("meterPerPxl == 0.0"));
+    return;
+  }
+
+  size_t const altitudeDataSz = altitudeDataM.size();
```
ygorshenin: s/Sz/Size/g

map/chart_generator.cpp
bykoianko: I removed the first one.

map/chart_generator.cpp
```diff
@@ -0,0 +1,215 @@
...
+  yAxisDataPxl.resize(altitudeDataSz);
+  for (size_t i = 0; i < altitudeDataSz; ++i)
+    yAxisDataPxl[i] = height - heightIndent - (altitudeDataM[i] - minAltM) / meterPerPxl;
+}
+
+void GenerateChartByPoints(size_t width, size_t height, vector<m2::PointD> const & geometry,
+                           bool day, vector<uint8_t> & frameBuffer)
+{
+  frameBuffer.clear();
```
ygorshenin: Why? frameBuffer is resized later.

map/framework.cpp
```diff
@@ -2980,3 +2982,26 @@ bool Framework::OriginalFeatureHasDefaultName(FeatureID const & fid) const
...
+bool Framework::GenerateRouteAltitudeChart(size_t width, size_t height, bool day,
+                                           vector<uint8_t> & imageRGBAData) const
+{
+  feature::TAltitudes altitudes;
+  if (!m_routingSession.GetRouteAltitudes(altitudes))
+    return false;
+  deque<double> segDistanceM;
+  if (!m_routingSession.GetSegDistanceM(segDistanceM) && segDistanceM.empty())
+    return false;
+  segDistanceM.push_front(0.0);
```
ygorshenin: No need in deque. Use vector::insert to the front() here. Amortized complexity O(1).

map/chart_generator.cpp
bykoianko: Probably it's better to draw without indents in this case.

map/chart_generator.cpp
bykoianko: Done

map/chart_generator.cpp
bykoianko: AGG is not a supported library. As far as I'm aware the last commit to it was made at 2005.

map/chart_generator.cpp
bykoianko: The problem is I need to test all of them.

map/chart_generator.cpp
bykoianko: Thanks. Fixed.

map/chart_generator.cpp
bykoianko: Yes. It's a possible solution. But taking into account that after call of NormalizeChartData() uniformAltitudeDataSz does not have a significant size I'd prefer to leave it as is.

routing/route.hpp
bykoianko: Done

map/framework.cpp
bykoianko: Yes. It starts form the first point. (Not from zero one.)

map/chart_generator.cpp
bykoianko: Done

map/chart_generator.cpp
bykoianko: Comment was added.

map/chart_generator.cpp
bykoianko: Done

map/chart_generator.hpp
bykoianko: I need to add one item form the begging in 
```c++
bool Framework::GenerateRouteAltitudeChart(size_t width, size_t height,
                                           vector<uint8_t> & imageRGBAData) const
{
  feature::TAltitudes altitudes;
  if (!m_routingSession.GetRouteAltitudes(altitudes))
    return false;
  deque<double> segDistanceM;
  if (!m_routingSession.GetSegDistanceM(segDistanceM) && segDistanceM.empty())
    return false;
  segDistanceM.push_front(0.0);

  if (altitudes.size() != segDistanceM.size())
  {
    LOG(LERROR, ("The route is in inconsistent state. Size of altitudes is", altitudes.size(),
                 ". Number of segment is", segDistanceM.size()));
    return false;
  }

  bool day = (GetMapStyle() != MapStyleDark);
  GenerateChart(width, height, segDistanceM, altitudes, day, imageRGBAData);
  return true;
}
```

map/chart_generator.hpp
bykoianko: Done

bykoianko: Do we have one? The thing is as far as I'm aware we do not have neither "namespace map" nor "namespace maps"

map/chart_generator.hpp
bykoianko: Done

map/chart_generator.hpp
bykoianko: Added.

map/chart_generator.cpp
bykoianko: I'd prefer to have an empty yAxisDataPxl in case of early return. (Before the resize.)

map/chart_generator.cpp
bykoianko: Fixed

map/chart_generator.cpp
```diff
@@ -0,0 +1,222 @@
...
+  using TSolidRenderer = agg::renderer_scanline_aa_solid<TBaseRenderer>;
+  using TPath = agg::poly_container_adaptor<vector<m2::PointD>>;
+  using TStroke = agg::conv_stroke<TPath>;
+
+  agg::rendering_buffer renderBuffer;
+  TPixelFormat pixelFormat(renderBuffer, agg::comp_op_src_over);
+  TBaseRenderer baseRenderer(pixelFormat);
+
+  frameBuffer.assign(width * kBPP * height, 0);
+  renderBuffer.attach(&frameBuffer[0], static_cast<unsigned int>(width),
```
ygorshenin: s/unsigned int/unsigned/g

map/chart_generator.cpp
ygorshenin: From my point of view it's not a problem when altitudeDataM is empty. But it's a problem when distanceDataM size differs from altitudeDataM size. Is it possible to distinguish between these two cases? Return true in the former case and false in the latter?

map/chart_generator.hpp
ygorshenin: So what? What's the problem to insert an element in a vector? Or, even better, why you can't do following:
```c++
vector<double> segDistanceM;
segDistanceM.push_back(0.0);
if (!m_routingSession.GetSegDistanceM(...) ...)
  ...
```

map/framework.cpp
```diff
@@ -1,5 +1,6 @@
 #include "map/framework.hpp"
 #include "map/ge0_parser.hpp"
+#include "map/chart_generator.hpp"
```
ygorshenin: Sort, please.

map/framework.cpp
```diff
@@ -2981,3 +2983,28 @@ bool Framework::OriginalFeatureHasDefaultName(FeatureID const & fid) const
...
+  segDistanceM.push_front(0.0);
+
+  if (altitudes.size() != segDistanceM.size())
+  {
+    LOG(LERROR, ("The route is in inconsistent state. Size of altitudes is", altitudes.size(),
+                 ". Number of segment is", segDistanceM.size()));
+    return false;
+  }
+
+  bool day = (GetMapStyle() != MapStyleDark);
```
ygorshenin: s/day/lightStyle/g

map/map_tests/chart_generator_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+bool IsColor(vector<uint8_t> const & frameBuffer, size_t startColorIdx, uint8_t expectedR,
+             uint8_t expectedG, uint8_t expectedB, uint8_t expectedA)
+{
+  CHECK_LESS(startColorIdx + kBPP, frameBuffer.size(), ());
+
+  return frameBuffer[startColorIdx] == expectedR && frameBuffer[startColorIdx + 1] == expectedG &&
+         frameBuffer[startColorIdx + 2] == expectedB && frameBuffer[startColorIdx + 3] == expectedA;
+}
+
+UNIT_TEST(NormalizeChartDataZeroTest)
```
ygorshenin: s/NormalizeChartDataZeroTest/NormalizeChartData_TestZeroes/g

map/map_tests/chart_generator_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+  feature::TAltitudes const altitudeDataM = {0, 0, 0};
+
+  vector<double> uniformAltitudeDataM;
+  NormalizeChartData(distanceDataM, altitudeDataM, 2 /* resultPointCount */, uniformAltitudeDataM);
+
+  vector<double> const expectedUniformAltitudeDataM = {0.0, 0.0};
+  TEST_EQUAL(expectedUniformAltitudeDataM, uniformAltitudeDataM, ());
+}
+
+UNIT_TEST(NormalizeChartDataTest)
```
ygorshenin: s/NormalizeChartDataTest/NormalizeChartData_Test/g

map/map_tests/chart_generator_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+  vector<double> uniformAltitudeDataM;
+  NormalizeChartData(distanceDataM, altitudeDataM, 10 /* resultPointCount */, uniformAltitudeDataM);
+
+  vector<double> const expectedUniformAltitudeDataM = {-9.0, -6.0, -3.0, 0.0,  3.0,
+                                                       6.0,  9.0,  12.0, 15.0, 18.0};
+  TEST(AlmostEqualAbs(uniformAltitudeDataM, expectedUniformAltitudeDataM), ());
+}
+
+UNIT_TEST(GenerateYAxisChartDataZeroTest)
```
ygorshenin: I suggest you to use naming rules similar to those I described above.

map/map_tests/chart_generator_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+  GenerateYAxisChartData(100 /* height */, 1.0 /* minMetersPerPxl */, altitudeDataM, yAxisDataPxl);
+  vector<double> expecttedYAxisDataPxl = {96.0, 94.0, 96.0, 98.0, 95.0};
+  TEST(AlmostEqualAbs(yAxisDataPxl, expecttedYAxisDataPxl), ());
+}
+
+UNIT_TEST(GenerateChartByPointsTest)
+{
+  vector<m2::PointD> const geometry = {{0.0, 0.0}, {10.0, 10.0}};
+
+  size_t constexpr width = 40;
```
ygorshenin: For uniformity, why not to declare height as is?

map/chart_generator.cpp
bykoianko: Currently GenerateYAxisChartData is not used the way altitudeDataM can be empty.
```c++
  vector<double> uniformAltitudeDataM;
  NormalizeChartData(distanceDataM, altitudeDataM, width, uniformAltitudeDataM);
  if (uniformAltitudeDataM.empty())
    return;

  vector<double> yAxisDataPxl;
  GenerateYAxisChartData(height, 1.0 /* minMetersPerPxl */, uniformAltitudeDataM, yAxisDataPxl);
```

Anyway I'll add the condition and test on it.

routing/osrm_path_segment_factory.cpp
```diff
@@ -56,13 +57,13 @@ void LoadPathGeometry(buffer_vector<TSeg, 8> const & buffer, size_t startIndex,
     if (startIdx < endIdx)
     {
       for (auto idx = startIdx; idx <= endIdx; ++idx)
-        loadPathGeometry.m_path.push_back(ft.GetPoint(idx));
+        loadPathGeometry.m_path.push_back(routing::Junction(ft.GetPoint(idx), feature::kDefaultAltitudeMeters));
```
ygorshenin: emplace_back?

routing/routing_session.hpp
```diff
@@ -138,6 +148,7 @@ class RoutingSession
   double GetCompletionPercent() const;
 
   void EmitCloseRoutingEvent() const;
+  bool HasRouteAltitudeImpl() const;
```
ygorshenin: To the private section?

routing/routing_session.hpp
```diff
@@ -105,6 +106,15 @@ class RoutingSession
   inline void SetState(State state) { m_state = state; }
 
   Route const & GetRoute() const { return m_route; }
+  /// \returns true if any altitude information along |m_route| is available and
```
ygorshenin: any or all?

routing/routing_session.hpp
```diff
@@ -105,6 +106,15 @@ class RoutingSession
...
+  /// false otherwise.
+  bool HasRouteAltitude() const;
+  /// \brief copies route altitude information to |routeAltitudes| if any is available and
+  /// returns true. If no route altitude information is available returns false.
```
ygorshenin: Am I right that even if some part of path has altitudes and some part has not, the function returns true? What's the purpose to return false? As I remember, when some part of map is not covered by srtm, we use minimum mwm height for this part.

map/chart_generator.cpp
bykoianko: Right. And for width.

map/chart_generator.cpp
bykoianko: Done

map/chart_generator.cpp
bykoianko: I'd prefer to have an empty |frameBuffer| in case of early return (int the beginning).

ygorshenin: Why? It's OK, and, in general, a good behavior when function does not modify it's arguments when fails.

map/framework.cpp
bykoianko: > No need in deque. Use vector::insert to the front() here.

Probably. It's not a big deal to copy the vector.

>  Amortized complexity O(1).

Are you sure? It's from C++11 standard:
"
...
23.3.6.5 vector modifiers [vector.modifiers]
iterator insert(const_iterator position, const T& x);
.....
Complexity: The complexity is linear in the number of elements inserted plus the distance to the end of the vector.
...
"

map/chart_generator.hpp
bykoianko: >  What's the problem to insert an element in a vector?

No problem. It's just linear to the vector size.

> Or, even better, why you can't do following:

It's also a possible solution. But in that case the functions RoutingSession::GetSegDistanceM(), Route::GetSegDistanceM(), FollowedPolyline::GetSegDistanceM() would have a strange meaning in my point of view. They weren't just getters but "adders".

What do you think about that?

map/framework.cpp
ygorshenin: Thanks for the reference from the standard. But when I  said "Amortized complexity" I meant amortized complexity in this case. How much it costs to fill segDistanceM? And how this bound changes when you prepend a single element to the vector? Still O(N), right?

ygorshenin: Do not forget that you're doing a lot of heavy things here, therefore a single prepend does not change the complexity of the code.

map/chart_generator.cpp
bykoianko: Ok. Let's move the check (distanceDataM.size() != altitudeDataM.size()) from framework to GenerateChart(). GenerateChart will return bool. 

Done

map/chart_generator.hpp
ygorshenin: From my point of view it's not a problem when a getter just appends to the provided vector. This property sometimes is very useful, for example when you need to collect results of several calls in the same vector. On the other hand, this approach makes it difficult to throw away current contents of the vector and fill it with results of the latest call. In any case, the behavior should be explained explicitly in the comment to the function.

It's up to you what to do here, but if you want to throw away current contents of the provided vectors, why not to return them by value? In other words, to prevent misunderstanding, I propose you to do following:

Instead of
```c++
void Foo(vector & v);
```

do this:
```c++
vector Foo();
```

No place for misunderstanding, right?

map/framework.cpp
bykoianko: Right. I've found a place where I has to copy the content of the container to be able to use deque:) I guess the best variant to insert {0, 0} in front of the vector.

map/chart_generator.hpp
ygorshenin: We should stop to follow old notation in the module :)

map/framework.cpp
bykoianko: Done

map/chart_generator.cpp
bykoianko: Done

map/framework.cpp
bykoianko: Done. But framework.hpp I left first.

map/framework.cpp
bykoianko: Done

map/map_tests/chart_generator_tests.cpp
bykoianko: I wrote:
NormalizeChartData_SmokeTest
GenerateYAxisChartData_SmokeTest

map/map_tests/chart_generator_tests.cpp
bykoianko: Done

map/map_tests/chart_generator_tests.cpp
bykoianko: Done

map/map_tests/chart_generator_tests.cpp
bykoianko: I saved a lot of lines in these file writing like: 40 /* height */. width is declared here because it is used twice below.

routing/osrm_path_segment_factory.cpp
bykoianko: Done in two places.

routing/routing_session.cpp
```diff
@@ -523,6 +523,31 @@ void RoutingSession::EmitCloseRoutingEvent() const
       alohalytics::Location::FromLatLon(lastGoodPoint.lat, lastGoodPoint.lon));
 }
 
+bool RoutingSession::HasRouteAltitudeImpl() const { return !m_route.GetAltitudes().empty(); }
+bool RoutingSession::HasRouteAltitude() const
```
bykoianko: Done

routing/routing_session.hpp
bykoianko: Done

routing/routing_session.hpp
bykoianko: Just altitude information. It's valid only if all the information is available. If not, there no way to interpret it correctly.

map/framework.cpp
kshalnev: Может быть сделать фунцию getcolor, в которую  спрятать зависимость от light/dark?

routing/routing_session.hpp
bykoianko: At run time no junction has invalid altitude. If the point is not covered by srtm we put default value for altitude at run time. The return value of this method has another meaning. This method return false if there's no navigation route as the time the method is called.

map/chart_generator.cpp
kshalnev: Или можно через ColorScheme класс, поставляемый ядру извне, и служащий источником данных о цветах разных объектов. В любом случае это не тема этого реквеста и моя основная идея не размазывать важные константы по всем файлам.

routing/routing_session.hpp
ygorshenin: So, why not write this:
```c++
If there's no navigation route, the method returns false.
```
in the comment? "If no route altitude information" sounds to generic and can be misinterpreted, as I did.

map/chart_generator.hpp
bykoianko: Thank you for detailed answer. In this case I don't want throw away the current content. I suggest to have a look at the code I'll push in a few minutes. Is it ok with you?

map/chart_generator.hpp
bykoianko: If so I suggest using "maps". The problem is there's something with name map in C++. If I were "namespace map" I would have a compilation error.

map/chart_generator.cpp
bykoianko: Yes. But GenerateChartByPoints returns nothing. I guess it'll be considered as success. So there's an alternative: either returning bool or clearing |frameBuffer|.
What do you think?

routing/routing_session.hpp
bykoianko: Ok. Let's write so.

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+namespace maps
+{
+void NormalizeChartData(vector<double> const & distanceDataM,
+                        feature::TAltitudes const & altitudeDataM, size_t resultPointCount,
+                        vector<double> & uniformAltitudeDataM)
+{
+  double constexpr kEpsilon = 1e-6;
+
+  uniformAltitudeDataM.clear();
+  if (distanceDataM.empty() || resultPointCount == 0 || resultPointCount == 1 ||
```
ygorshenin: The size of uniformAltitudeDataM should be the same as resultPointCount. So it's wrong to return an empty vector when resultPointCount equals to one.

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+      return static_cast<double>(altitudeDataM.back());
+
+    auto const lowerIt = lower_bound(distanceDataM.cbegin(), distanceDataM.cend(), distFormStartM);
+    if (lowerIt == distanceDataM.cbegin())
+      return static_cast<double>(altitudeDataM.front());
+    if (lowerIt == distanceDataM.cend())
+      return static_cast<double>(altitudeDataM.back());
+
+    size_t const nextPointIdx = distance(distanceDataM.cbegin(), lowerIt);
+    CHECK_LESS(0, nextPointIdx, ("distFormStartM is more than 0 but nextPointIdx == 0."));
```
ygorshenin: s/more/greater/

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+{
+  if (altitudeDataM.empty())
+    return;
+
+  uint32_t constexpr kHeightIndentPxl = 2;
+  yAxisDataPxl.clear();
+  uint32_t heightIndent = kHeightIndentPxl;
+  if (height <= 2 * kHeightIndentPxl)
+  {
+    LOG(LERROR, ("Chart height is less then 2 * kHeightIndentPxl (", 2 * kHeightIndentPxl, ")"));
```
ygorshenin: s/then/than/
less or equals

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+  uniformAltitudeDataM.resize(resultPointCount);
+  for (size_t i = 0; i < resultPointCount; ++i)
+    uniformAltitudeDataM[i] = calculateAltitude(static_cast<double>(i) * stepLenM);
+}
+
+void GenerateYAxisChartData(uint32_t height, double minMetersPerPxl,
+                            vector<double> const & altitudeDataM, vector<double> & yAxisDataPxl)
+{
+  if (altitudeDataM.empty())
```
ygorshenin: Need to clear yAxisDataPxl when altitudeDataM is empty.

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+  stroke.width(kLineWidthPxl);
+  stroke.line_cap(agg::round_cap);
+  stroke.line_join(agg::round_join);
+
+  rasterizer.add_path(stroke);
+  agg::render_scanlines_aa_solid(rasterizer, scanline, baseRenderer, kLineColor);
+}
+
+bool GenerateChart(uint32_t width, uint32_t height, vector<double> const & distanceDataM,
+                   feature::TAltitudes const & altitudeDataM, bool day,
```
ygorshenin: s/day/lightTheme/g
?

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+  vector<double> uniformAltitudeDataM;
+  NormalizeChartData(distanceDataM, altitudeDataM, width, uniformAltitudeDataM);
+  if (uniformAltitudeDataM.empty())
+    return true;
+
+  vector<double> yAxisDataPxl;
+  GenerateYAxisChartData(height, 1.0 /* minMetersPerPxl */, uniformAltitudeDataM, yAxisDataPxl);
+
+  size_t const uniformAltitudeDataSz = yAxisDataPxl.size();
+  if (uniformAltitudeDataSz <= 1)
```
ygorshenin: Frame buffer should not be empty if uniformAltitudeDataSz equals to 1, right?

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+  vector<double> uniformAltitudeDataM;
+  NormalizeChartData(distanceDataM, altitudeDataM, width, uniformAltitudeDataM);
+  if (uniformAltitudeDataM.empty())
+    return true;
+
+  vector<double> yAxisDataPxl;
+  GenerateYAxisChartData(height, 1.0 /* minMetersPerPxl */, uniformAltitudeDataM, yAxisDataPxl);
+
+  size_t const uniformAltitudeDataSz = yAxisDataPxl.size();
```
ygorshenin: s/Sz/Size/g

map/framework.cpp
```diff
@@ -2981,3 +2982,20 @@ bool Framework::OriginalFeatureHasDefaultName(FeatureID const & fid) const
...
+{
+  feature::TAltitudes altitudes;
+  if (!m_routingSession.GetRouteAltitudes(altitudes))
+    return false;
+  vector<double> segDistanceM;
+  if (!m_routingSession.GetSegDistanceM(segDistanceM) && segDistanceM.empty())
+    return false;
+  segDistanceM.insert(segDistanceM.begin(), 0.0);
+
+  bool dayLight = (GetMapStyle() != MapStyleDark);
```
ygorshenin: s/dayLight/lightTheme/g

map/chart_generator.cpp
bykoianko: I rewrote NormalizeChartData. It returns false in case of error and does not change uniformAltitudeDataM.

map/chart_generator.cpp
bykoianko: Done

map/chart_generator.cpp
bykoianko: Done. 
less or equal

map/chart_generator.cpp
bykoianko: yes

map/chart_generator.cpp
bykoianko: Done

map/chart_generator.cpp
bykoianko: Now I return false if uniformAltitudeDataSz is less than one. And I do not touch frame buffer in that case.

map/chart_generator.cpp
bykoianko: Done

map/framework.cpp
bykoianko: Done

map/chart_generator.cpp
ygorshenin: But it's not an error when resultPointCount is zero or one.

map/framework.cpp
bykoianko: bool Framework::IsLightMapTheme() const; 
Done

map/chart_generator.hpp
bykoianko: Done

routing/route.hpp
```diff
@@ -52,21 +55,10 @@ class Route
...
-  {
-    swap(m_times, v);
-  }
-
-  inline void SetStreetNames(TStreets & v)
-  {
-    swap(m_streets, v);
-  }
-
+  inline void SetTurnInstructions(TTurns &&v) { m_turns = move(v); }
```
kshalnev: '&& v' - space before 'v'

routing/routing_session.hpp
kshalnev: please, do it :)

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+namespace maps
+{
+bool NormalizeChartData(vector<double> const & distanceDataM,
+                        feature::TAltitudes const & altitudeDataM, size_t resultPointCount,
+                        vector<double> & uniformAltitudeDataM)
+{
+  double constexpr kEpsilon = 1e-6;
+
+  if (distanceDataM.empty() || resultPointCount == 0 || resultPointCount == 1 ||
```
kshalnev: may be resultPointCount <= 1

ygorshenin: As we discussed, it's not a problem that resultPointCount is zero or one, and the method should return true. Note that by the method contract in the absence of errors the size of uniformAltitudeDataM should be equal to resultPointCount.

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  if (!is_sorted(distanceDataM.cbegin(), distanceDataM.cend()))
+  {
+    LOG(LERROR, ("Route segment distances are not sorted."));
+    return false;
+  }
+
+  double const routeLenM = distanceDataM.back();
+  double const stepLenM = routeLenM / static_cast<double>(resultPointCount - 1);
```
ygorshenin: double const stepLenM = resultPointCount <= 1 ? 0.0 : routeLenM / (resultPointCount - 1);

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+                           bool lightTheme, vector<uint8_t> & frameBuffer)
+{
+  frameBuffer.clear();
+  if (width == 0 || height == 0 || geometry.empty())
+    return;
+
+  agg::rgba8 const kBackgroundColor = agg::rgba8(255, 255, 255, 0);
+  agg::rgba8 const kLineColor =
+      lightTheme ? agg::rgba8(30, 150, 240, 255) : agg::rgba8(255, 230, 140, 255);
+  agg::rgba8 const kCurveColor = lightTheme ? agg::rgba8(30, 150, 240, 20) : agg::rgba8(255, 230, 140, 20);
```
kshalnev: предлагаю сделать функции GetLineColor, GetCurveColor следующего вида:
color GetXxxColor()
{
switch (MapStyle.GetCurrentStyle())
{
...
}
}
тогда не надо будет делать костыль во Framework - IsLightTheme.

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+  double const stepLenM = routeLenM / static_cast<double>(resultPointCount - 1);
+
+  auto const calculateAltitude = [&](double distFormStartM) {
+    if (distFormStartM <= distanceDataM.front())
+      return static_cast<double>(altitudeDataM.front());
+    if (distFormStartM >= distanceDataM.back())
+      return static_cast<double>(altitudeDataM.back());
+
+    auto const lowerIt = lower_bound(distanceDataM.cbegin(), distanceDataM.cend(), distFormStartM);
+    if (lowerIt == distanceDataM.cbegin())
```
ygorshenin: It's not possible.

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+      return static_cast<double>(altitudeDataM.front());
+    if (lowerIt == distanceDataM.cend())
```
ygorshenin: It's not possible.

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+      return static_cast<double>(altitudeDataM.back());
+
+    size_t const nextPointIdx = distance(distanceDataM.cbegin(), lowerIt);
```
ygorshenin: As lowerIt does not equal to cbegin, nextPointIdx can't be zero. ASSERT is enough.

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+bool NormalizeChartData(vector<double> const & distanceDataM,
+                        feature::TAltitudes const & altitudeDataM, size_t resultPointCount,
+                        vector<double> & uniformAltitudeDataM)
+{
+  double constexpr kEpsilon = 1e-6;
+
+  if (distanceDataM.empty() || resultPointCount == 0 || resultPointCount == 1 ||
+      distanceDataM.size() != altitudeDataM.size())
+  {
+    return false;
```
ygorshenin: Error log here too?

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+void GenerateYAxisChartData(uint32_t height, double minMetersPerPxl,
+                            vector<double> const & altitudeDataM, vector<double> & yAxisDataPxl)
+{
+  yAxisDataPxl.clear();
+  if (altitudeDataM.empty())
+    return;
+
+  uint32_t constexpr kHeightIndentPxl = 2;
+  uint32_t heightIndent = kHeightIndentPxl;
+  if (height <= 2 * kHeightIndentPxl)
```
ygorshenin: s/kHeightIndenPxl/heightIndent/

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+    LOG(LERROR, ("Chart height is less or equal than 2 * kHeightIndentPxl (", 2 * kHeightIndentPxl, ")"));
+    heightIndent = 0;
+  }
+
+  auto const minMaxAltitudeIt = minmax_element(altitudeDataM.begin(), altitudeDataM.end());
+  double const minAltM = *minMaxAltitudeIt.first;
+  double const maxAltM = *minMaxAltitudeIt.second;
+  double const deltaAltM = maxAltM - minAltM;
+  uint32_t const drawHeightPxl = height - 2 * heightIndent;
+  double const meterPerPxl = max(minMetersPerPxl, deltaAltM / static_cast<double>(drawHeightPxl));
```
ygorshenin: s/meterPerPxl/metersPerPxl/g

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+                   vector<uint8_t> & frameBuffer)
+{
+  if (distanceDataM.size() != altitudeDataM.size())
+  {
+    LOG(LERROR, ("The route is in inconsistent state. Size of altitudes is", altitudeDataM.size(),
+                 ". Number of segment is", distanceDataM.size()));
+    return false;
+  }
+
+  if (altitudeDataM.empty())
```
ygorshenin: What's wrong here? In this case you need to clear the buffer and return true.

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+  vector<double> uniformAltitudeDataM;
+  if (!NormalizeChartData(distanceDataM, altitudeDataM, width, uniformAltitudeDataM))
+    return false;
+
+  vector<double> yAxisDataPxl;
+  GenerateYAxisChartData(height, 1.0 /* minMetersPerPxl */, uniformAltitudeDataM, yAxisDataPxl);
+
+  size_t const uniformAltitudeDataSize = yAxisDataPxl.size();
+  if (uniformAltitudeDataSize <= 1)
```
ygorshenin: Same here - why do you think that zero or one are error cases?

map/framework.cpp
```diff
@@ -2981,3 +2987,20 @@ bool Framework::OriginalFeatureHasDefaultName(FeatureID const & fid) const
...
+bool Framework::HasRouteAltitude() const { return m_routingSession.HasRouteAltitude(); }
+
+bool Framework::GenerateRouteAltitudeChart(uint32_t width, uint32_t height,
+                                           vector<uint8_t> & imageRGBAData) const
+{
+  feature::TAltitudes altitudes;
+  if (!m_routingSession.GetRouteAltitudes(altitudes))
+    return false;
+  vector<double> segDistanceM;
+  if (!m_routingSession.GetSegDistanceM(segDistanceM) && segDistanceM.empty())
```
```c++
if (!m_routingSession.GetSegDistanceM(segDistanceM))
  return false;
```
I.e. no need to check that segDistanceM is empty.

map/map_tests/chart_generator_tests.cpp
```diff
@@ -0,0 +1,125 @@
...
+UNIT_TEST(GenerateYAxisChartData_EmptyAltitudeDataTest)
+{
+  vector<double> const altitudeDataM = {};
+  vector<double> yAxisDataPxl;
+
+  maps::GenerateYAxisChartData(30 /* height */, 1.0 /* minMetersPerPxl */, altitudeDataM, yAxisDataPxl);
+  TEST(yAxisDataPxl.empty(), ());
+}
+
+UNIT_TEST(GenerateYAxisChartDataTest)
```
ygorshenin: s/GenerateYAxisChartDataTest/GenerateYAxisChartData_Test/g

routing/routing_session.cpp
```diff
@@ -523,6 +523,36 @@ void RoutingSession::EmitCloseRoutingEvent() const
       alohalytics::Location::FromLatLon(lastGoodPoint.lat, lastGoodPoint.lon));
 }
 
+bool RoutingSession::HasRouteAltitudeImpl() const
+{
+  return !m_route.GetAltitudes().empty()
```
ygorshenin: The first condition is redundant.

map/chart_generator.cpp
```diff
@@ -0,0 +1,236 @@
...
+    return false;
+
+  double const oneSegLenPix = static_cast<double>(width) / (uniformAltitudeDataSize - 1);
+  vector<m2::PointD> geometry(uniformAltitudeDataSize);
+  for (size_t i = 0; i < uniformAltitudeDataSize; ++i)
+    geometry[i] = m2::PointD(i * oneSegLenPix, yAxisDataPxl[i]);
+
+  frameBuffer.clear();
```
kshalnev: GenerateChartByPoints уже чистит массив. frameBuffer.clear() можно убрать.

map/chart_generator.cpp
kshalnev: why? distanceDataM.cbegin() equals to 0.0, first value of distFormStartM is also 0.0

ygorshenin: Please, take a look at the couple of ifs at the beginning of the function.

routing/route.hpp
bykoianko: Fixed

map/chart_generator.cpp
bykoianko: 1. It's been rewritten in the RP. Please have a look.

> Note that by the method contract in the absence of errors the size of uniformAltitudeDataM should be equal to resultPointCount.

2. Unfortunately it's not so. There's following case:
```
NormalizeChartData({} /* distanceDataM */, {} /* altitudeDataM */, 10 /* resultPointCount */, uniformAltitudeDataM)
```
If altitudeDataM/distanceDataM are empty there no information to fill uniformAltitudeDataM. I leave it empty in this case.

map/chart_generator.cpp
bykoianko: yes

map/chart_generator.cpp
bykoianko: Да. Давай так сделаем.

map/chart_generator.cpp
bykoianko: Yes. Let's remove it.

map/chart_generator.cpp
bykoianko: Let's remove it as well

map/chart_generator.cpp
bykoianko: Ok

map/chart_generator.cpp
bykoianko: If distanceDataM.size() != altitudeDataM.size() yes. Error log. This method has been rewritten. Please have a look.

map/chart_generator.cpp
bykoianko: No. There's been |heightIndent|. I'll rename heightIndent to heightIndentPxl

map/chart_generator.cpp
bykoianko: Done

map/chart_generator.cpp
bykoianko: Yes. All these method has been rewritten. Now it works the way you describe. There's a test on it(GenerateChart_NoPointsTest)

map/chart_generator.cpp
bykoianko: Yes. This method has been rewritten. (GenerateChart_OnePointTest)

map/framework.cpp
bykoianko: Yes

map/map_tests/chart_generator_tests.cpp
bykoianko: Fixed

routing/routing_session.cpp
bykoianko: Yes for now

routing/routing_session.hpp
bykoianko: It is private:)
```
private:
  struct DoReadyCallback
  {
    RoutingSession & m_rs;
    TReadyCallback m_callback;
    threads::Mutex & m_routeSessionMutexInner;

    DoReadyCallback(RoutingSession & rs, TReadyCallback const & cb,
                    threads::Mutex & routeSessionMutex)
        : m_rs(rs), m_callback(cb), m_routeSessionMutexInner(routeSessionMutex)
    {
    }

    void operator()(Route & route, IRouter::ResultCode e);
  };

  void AssignRoute(Route & route, IRouter::ResultCode e);

  /// Returns a nearest speed camera record on your way and distance to it.
  /// Returns kInvalidSpeedCameraDistance if there is no cameras on your way.
  double GetDistanceToCurrentCamM(SpeedCameraRestriction & camera, Index const & index);

  /// RemoveRoute removes m_route and resets route attributes (m_state, m_lastDistance, m_moveAwayCounter).
  void RemoveRoute();
  void RemoveRouteImpl();

  bool HasRouteAltitudeImpl() const;
```

bykoianko:
@ygorshenin @kshalnev PTAL

rokuz:
LGTM

map/chart_generator.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+  return true;
+}
+
+bool GenerateYAxisChartData(uint32_t height, double minMetersPerPxl,
+                            vector<double> const & altitudeDataM, vector<double> & yAxisDataPxl)
+{
+  yAxisDataPxl.clear();
+  if (altitudeDataM.empty())
+    return true;
```
ygorshenin: Please, move yAxisDataPxl.clear() to the block where you return true. Otherwise it's possible to return false but modify the result argument.

kshalnev:
LGTM

map/chart_generator.cpp
bykoianko: Right!

map/chart_generator.cpp
bykoianko: Да. Ты прав. Поправил.

ygorshenin:
LGTM

map/chart_generator.cpp
```diff
@@ -0,0 +1,276 @@
...
+  double const deltaAltM = maxAltM - minAltM;
+  uint32_t const drawHeightPxl = height - 2 * heightIndentPxl;
+  double const metersPerPxl = max(minMetersPerPxl, deltaAltM / static_cast<double>(drawHeightPxl));
+  if (metersPerPxl == 0.0)
+  {
+    LOG(LERROR, ("metersPerPxl == 0.0"));
+    return false;
+  }
+
+  size_t const altitudeDataSz = altitudeDataM.size();
```
ygorshenin: s/Sz/Size/g

bykoianko: Done

bykoianko:
Всем спасибо за ревью!

PR #4196, opened 2016-09-01
[android] initializing alohalytics on startup

milchakov:
Фикс креша на старте.
@ygorshenin 
@yunikkk

### DIFF
--- a/android/src/com/mapswithme/maps/MwmApplication.java
+++ b/android/src/com/mapswithme/maps/MwmApplication.java
@@ -107,6 +107,9 @@ public class MwmApplication extends Application
     initHockeyApp();
     initCrashlytics();
     initPushWoosh();
+    // Workaround to initialize alohalytics for transferring http requests.
+    // Http wrapper should be cut out from the alohalytics in the future.
+    Statistics s = Statistics.INSTANCE;
 
     String settingsPath = getSettingsPath();
     new File(settingsPath).mkdirs();


android/src/com/mapswithme/maps/MwmApplication.java
```diff
@@ -107,6 +107,9 @@ public void onCreate()
     initHockeyApp();
     initCrashlytics();
     initPushWoosh();
+    // Workaround to initialize alohalytics for transferring http requests.
+    // Http wrapper should be cut out from the alohalytics in the future.
+    Statistics s = Statistics.INSTANCE;
```
ygorshenin: OK as a workaround. But need to extract http code into an individual unit and initialize it separately.

ygorshenin:
LGTM, but please, move out http code from alohalytics.

android/src/com/mapswithme/maps/MwmApplication.java
yunikkk: +100500 for extracting http module.

android/src/com/mapswithme/maps/MwmApplication.java
```diff
@@ -107,6 +107,9 @@ public void onCreate()
     initHockeyApp();
     initCrashlytics();
     initPushWoosh();
+    // Workaround to initialize alohalytics for transferring http requests.
```
yunikkk: Leave TODO here, it will help to remember about the case.

yunikkk:
LGTM, with or without TODO

milchakov:
Я буду выпиливать сразу после своей текущей задачи, вмержите плиз

PR #4239, opened 2016-09-08
[DO NOT MERGE] Bg locale

therearesomewhocallmetim:
Updated twine, regenerated the string resources from the old strings.txt and got a lot of modifications.


therearesomewhocallmetim:
Should I glue together the commits where I remove the old twine and add the new one?

PR #4270, opened 2016-09-14
Passing min and max route altitude. 

bykoianko:
Передача клиенту минимальной и максимальной высоты по проложенному маршруту.

@ygorshenin @igrechuhin @alexzatsepin PTAL

https://jira.mail.ru/browse/MAPSME-2387

### DIFF
--- a/android/jni/com/mapswithme/maps/Framework.cpp
+++ b/android/jni/com/mapswithme/maps/Framework.cpp
@@ -895,17 +895,36 @@ Java_com_mapswithme_maps_Framework_nativeGetRouteFollowingInfo(JNIEnv * env, jcl
 }
 
 JNIEXPORT jintArray JNICALL
-Java_com_mapswithme_maps_Framework_nativeGenerateRouteAltitudeChartBits(JNIEnv * env, jclass, jint width, jint height)
+Java_com_mapswithme_maps_Framework_nativeGenerateRouteAltitudeChartBits(JNIEnv * env, jclass, jint width, jint height, jobject routeAltitudeLimits)
 {
   ::Framework * fr = frm();
   ASSERT(fr, ());
 
   vector<uint8_t> imageRGBAData;
-  if (!fr->GenerateRouteAltitudeChart(width, height, imageRGBAData))
+  int32_t minRouteAltitude = 0;
+  int32_t maxRouteAltitude = 0;
+  measurement_utils::Units units = measurement_utils::Units::Metric;
+  if (!fr->GenerateRouteAltitudeChart(width, height, imageRGBAData, minRouteAltitude, maxRouteAltitude, units))
   {
     LOG(LWARNING, ("Can't generate route altitude image."));
     return nullptr;
   }
+
+  // Passing route limits.
+  jclass const routeAltitudeLimitsClass = env->GetObjectClass(routeAltitudeLimits);
+  ASSERT(routeAltitudeLimitsClass, ());
+
+  static jfieldID const minRouteAltitudeField = env->GetFieldID(routeAltitudeLimitsClass, "minRouteAltitude", "I");
+  ASSERT(minRouteAltitudeField, ());
+  env->SetIntField(routeAltitudeLimits, minRouteAltitudeField, minRouteAltitude);
+
+  static jfieldID const maxRouteAltitudeField = env->GetFieldID(routeAltitudeLimitsClass, "maxRouteAltitude", "I");
+  ASSERT(maxRouteAltitudeField, ());
+  env->SetIntField(routeAltitudeLimits, maxRouteAltitudeField, maxRouteAltitude);
+
+  static jfieldID const isMetricUnitsField = env->GetFieldID(routeAltitudeLimitsClass, "isMetricUnits", "Z");
+  ASSERT(isMetricUnitsField, ());
+  env->SetBooleanField(routeAltitudeLimits, isMetricUnitsField, units == measurement_utils::Units::Metric);
 
   size_t const imageRGBADataSize = imageRGBAData.size();
   ASSERT_NOT_EQUAL(imageRGBADataSize, 0, ("GenerateRouteAltitudeChart returns true but the vector with altitude image bits is empty."));
--- a/android/src/com/mapswithme/maps/Framework.java
+++ b/android/src/com/mapswithme/maps/Framework.java
@@ -61,6 +61,13 @@ public class Framework
     public boolean buildings;
   }
 
+  public static class RouteAltitudeLimits
+  {
+    public int minRouteAltitude;
+    public int maxRouteAltitude;
+    public boolean isMetricUnits;
+  }
+
   // this class is just bridge between Java and C++ worlds, we must not create it
   private Framework() {}
 
@@ -81,7 +88,8 @@ public class Framework
     if (width <= 0 || height <= 0)
       return null;
 
-    final int[] altitudeChartBits = Framework.nativeGenerateRouteAltitudeChartBits(width, height);
+    RouteAltitudeLimits routeAltitudeLimits = new RouteAltitudeLimits();
+    final int[] altitudeChartBits = Framework.nativeGenerateRouteAltitudeChartBits(width, height, routeAltitudeLimits);
     if (altitudeChartBits == null)
       return null;
 
@@ -165,7 +173,7 @@ public class Framework
   public static native RoutingInfo nativeGetRouteFollowingInfo();
 
   @Nullable
-  public static native final int[] nativeGenerateRouteAltitudeChartBits(int width, int height);
+  public static native final int[] nativeGenerateRouteAltitudeChartBits(int width, int height, RouteAltitudeLimits routeAltitudeLimits);
 
   // When an end user is going to a turn he gets sound turn instructions.
   // If C++ part wants the client to pronounce an instruction nativeGenerateTurnNotifications returns
--- a/iphone/Maps/Classes/Routing/MWMRouter.mm
+++ b/iphone/Maps/Classes/Routing/MWMRouter.mm
@@ -288,8 +288,14 @@ bool isMarkerPoint(MWMRoutePoint const & point) { return point.IsValid() && !poi
     if (!imageData)
     {
       vector<uint8_t> imageRGBAData;
-      if (!GetFramework().GenerateRouteAltitudeChart(width, height, imageRGBAData))
+      int32_t minRouteAltitude = 0;
+      int32_t maxRouteAltitude = 0;
+      measurement_utils::Units units = measurement_utils::Units::Metric;
+      if (!GetFramework().GenerateRouteAltitudeChart(width, height, imageRGBAData,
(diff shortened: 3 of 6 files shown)


map/framework.cpp
```diff
@@ -2997,6 +2999,29 @@ bool Framework::GenerateRouteAltitudeChart(uint32_t width, uint32_t height,
...
+
+  if (!maps::GenerateChart(width, height, segDistance, altitudes, GetMapStyle(), imageRGBAData))
+    return false;
+
+  auto const minMaxIt = minmax_element(altitudes.cbegin(), altitudes.cend());
+  feature::TAltitude const minRouteAltitudeM = *minMaxIt.first;
+  feature::TAltitude const maxRouteAltitudeM = *minMaxIt.second;
+
+  altitudeUnits = measurement_utils::Units::Metric;
+  UNUSED_VALUE(settings::Get(settings::kMeasurementUnits, altitudeUnits));
```
```c++
if (!settings::Get(settings::kMeasurementUnits, altitudeUnits))
  altitudeUnits = measurement_utils::Units::Metric;
```

bykoianko: Done

bykoianko:
@ygorshenin @igrechuhin @alexzatsepin PTAL

igrechuhin:
LGTM

map/framework.hpp
```diff
@@ -729,13 +729,21 @@ class Framework : public search::ViewportSearchCallback::Delegate,
   /// false otherwise.
   bool HasRouteAltitude() const;
   /// \brief Generates 4 bytes per point image (RGBA) and put the data to |imageRGBAData|.
+  /// \param width is width of chart shall be generated in pixels.
+  /// \param height is height of chart shall be generated in pixels.
+  /// \param imageRGBAData is bits of result image in RGBA.
+  /// \param minRouteAltitude is min altitude along the route in altitudeUnits.
```
kshalnev: optional. let add that is output parameter. it may be obvious from the code, but would be better to add to comment, i think.

android/jni/com/mapswithme/maps/Framework.cpp
```diff
@@ -895,18 +895,37 @@ Java_com_mapswithme_maps_Framework_nativeGetRouteFollowingInfo(JNIEnv * env, jcl
...
   {
     LOG(LWARNING, ("Can't generate route altitude image."));
     return nullptr;
   }
 
+  // Passing route limits.
+  jclass const routeAltitudeLimitsClass = env->GetObjectClass(routeAltitudeLimits);
+  ASSERT(routeAltitudeLimitsClass, ());
+
+  static jfieldID const minRouteAltitudeField = env->GetFieldID(routeAltitudeLimitsClass, "minRouteAltitude", "I");
```
ygorshenin: Is it good to store static field id here? There's no guarantees that env is the same between calls, isn't it?

alexzatsepin:
LGTM

kshalnev:
может быть сделать это свойствами объекта "маршрут"? там, мне кажется, логичнее иметь эти поля.

bykoianko:
> может быть сделать это свойствами объекта "маршрут"? там, мне кажется, логичнее иметь эти поля.

Ты про класс Route в ядре? Сейчас эти св-ва (минимальная и максимальная высота + единицы измерения) создаются при генерации диаграммы высот и сразу (в том же методе) уходят в клиент. Когда и если потребуется их можно добавить в Route. Но пока мне кажется не стоит хранить лишние состояния в классе Route.

map/framework.hpp
bykoianko: Are use sure it's worth doing so? I guess passing a param by non-const ref implies that it's an output param. Is not it so?

android/jni/com/mapswithme/maps/Framework.cpp
bykoianko: On the one hand methods GetFieldID and GetMethodID are slow. ("In the worst case it has to process the strings, and lookup the fieldID in the corresponding class data."). Because of it in our project we always use static or global for GetFieldID and very ofter for GetMethodID. @yunikkk insists on it.

On the other hand according "to the specification, the field ID is guaranteed to be valid only as long as the class is loaded." (routeAltitudeLimitsClass) "So, if for some reason the class gets unloaded, and then loaded again, a former field ID is probably invalid. You can work around that by storing the jclass reference as a global reference somewhere, thus preventing the GC from unloading the class."

According to 
https://rkennke.wordpress.com/2007/07/24/efficient-jni-programming-ii-field-and-method-access/
we cannot just assign result of GetFieldID/GetMethodID to static value and reuse it laster.

@yunikkk Could I ask you please to comment it?

kshalnev:
ну сейчас это выглядит - "нарисовать диаграмму и попутно посчитать кое-какую статистику для раута, а другая статистика для раута доступна в другом месте". я предлагаю объединить статистику в одном месте и не смешивать её с функцией "нарисовать диаграмму". это опционально, просто мое видение.

map/framework.hpp
kshalnev: non-constant reference also means in-out params. sometimes it is obvious what is where, but i would prefer obvious declaration in comment or in name. it is optional, just my opinion.

igrechuhin:
Test

android/jni/com/mapswithme/maps/Framework.cpp
```diff
@@ -895,18 +895,37 @@ Java_com_mapswithme_maps_Framework_nativeGetRouteFollowingInfo(JNIEnv * env, jcl
 }
 
 JNIEXPORT jintArray JNICALL
-Java_com_mapswithme_maps_Framework_nativeGenerateRouteAltitudeChartBits(JNIEnv * env, jclass, jint width, jint height)
+Java_com_mapswithme_maps_Framework_nativeGenerateRouteAltitudeChartBits(JNIEnv * env, jclass, jint width, jint height, jobject routeAltitudeLimits)
```
igrechuhin: Test

igrechuhin: test1

android/src/com/mapswithme/maps/Framework.java
```diff
@@ -61,6 +61,13 @@
     public boolean buildings;
   }
 
+  public static class RouteAltitudeLimits
+  {
+    public int minRouteAltitude;
+    public int maxRouteAltitude;
```
igrechuhin: test2

android/jni/com/mapswithme/maps/Framework.cpp
yunikkk: To @ygorshenin we doesn't worry about env, cause resulting jfieldID will be equal in any case.

To @bykoianko in fact Android doesn't unload classes, so it's valid to cache them. You can check http://android-developers.blogspot.com.by/2011/11/jni-local-reference-changes-in-ics.html for details (..."In a system without class unloading (like Android), you can treat jfieldID and jmethodID as if they were global."...)

yunikkk: However, there were rumors on last google IO that N release WILL unload classes, so we probably should reconsider our approach in future.

iphone/Maps/Classes/Routing/MWMRouter.mm
```diff
@@ -288,8 +288,14 @@ - (void)routeAltitudeImageForSize:(CGSize)size completion:(MWMImageBlock)block
     if (!imageData)
     {
       vector<uint8_t> imageRGBAData;
-      if (!GetFramework().GenerateRouteAltitudeChart(width, height, imageRGBAData))
+      int32_t minRouteAltitude = 0;
+      int32_t maxRouteAltitude = 0;
+      measurement_utils::Units units = measurement_utils::Units::Metric;
+      if (!GetFramework().GenerateRouteAltitudeChart(width, height, imageRGBAData,
+                                                     minRouteAltitude, maxRouteAltitude, units))
+      {
```
yunikkk: Лишние скобки?

bykoianko: Мы обычно так пишем, когда выражение в if переносится на новую строчку.

bykoianko:
@ygorshenin

android/src/com/mapswithme/maps/Framework.java
```diff
@@ -61,6 +61,13 @@
     public boolean buildings;
   }
 
+  public static class RouteAltitudeLimits
+  {
+    public int minRouteAltitude;
```
ygorshenin: Why do we need this at all? I don't see places where this class is used. Moreover, it seems quite strange to me that the method named GenerateRouteAltitudeChart returns min and max altitudes of a route. min and max altitudes are properties of a current route, not a chart.

bykoianko: > Why do we need this at all? I don't see places where this class is used. 

I'll be used in Android code by @alexzatsepin . He's waiting this PR to use it. The same situation with iOS code.

> Moreover, it seems quite strange to me that the method named GenerateRouteAltitudeChart returns min and max altitudes of a route. min and max altitudes are properties of a current route, not a chart.

There're the following options:
1. Add a special method for getting RouteAltitudeLimits;
2. Add this info RoutingInfo and return it with nativeGetRouteFollowingInfo();
3. Return altitude limits in method which generates route altitude diagram.

As for 1 method for returning limits and for generating diagram could return different result (for two different routes) because the route could be deleted/rebuilt;

As for 2 there's one more disadvantage: RoutingInfo contains significant amount info about following along route;

The third variant (implemented now): the method (nativeGenerateRouteAltitudeChartBits) is developed to be used by client to get info for altitude diagram. So it's useful for clients to get the info together. Moreover if everything is returned in one method we can guarantee that this information is for the same route.

Probably it's worth renaming it to nativeGenerateRouteAltitudeInfo or something like that.

ygorshenin: What's the problem to get current Route, and create a couple of methods, like:
```c++
GenerateChart(Route const & route, ...)
GetMinMaxAltitude(Route const & route, ...)
```
and then call both methods on the same route instance?

bykoianko: > What's the problem to get current Route, and create a couple of methods, like:
GenerateChart(Route const & route, ...)
GetMinMaxAltitude(Route const & route, ...)

Saying methods, which class are you talking about? Framework? RoutingSession? Route? We can lock the current route only within RoutingSession methods. Or as an alternative we can copy instance or the current route (RoutingSession::m_route) and then work with it.

ygorshenin: So, please, implement functionality to lock current route, or to get current routing session, and work with it please.

bykoianko: I had a look at m_routingSession.m_route. It's quite possible to wrap it in shared_ptr and use it as we agreed. But it takes some time. Artem ask to merge it ASAP. So let's merge it as is and implement share_prt<Route> in separate PR.

burivuh: Guys, let's do it in the master after creating new release branch. Volodya, please setup a task for this.

PR #4300, opened 2016-09-16
Adding BicycleTest on a route from mail.ru office to ul. simanova5k2 which does not pass now.

bykoianko:
Добавил тест на вело маршрут, который сейчас не прокладывается.
Притом, я отдельно проверил, что маршрут не прокладывается ни при помощи AStarBidirectionalRoutingAlgorithm, ни при помощи AStarRoutingAlgorithm.

@ygorshenin @mpimenov PTAL

### DIFF
--- a/routing/routing_benchmarks/bicycle_routing_tests.cpp
+++ b/routing/routing_benchmarks/bicycle_routing_tests.cpp
@@ -44,4 +44,11 @@ UNIT_CLASS_TEST(BicycleTest, Smoke)
   m2::PointD const final = MercatorBounds::FromLatLon(55.79369, 37.56054);
   TestRouters(start, final);
 }
+
+UNIT_CLASS_TEST(BicycleTest, MailRuOfficeToUlSimanova5K2)
+{
+  m2::PointD const start = MercatorBounds::FromLatLon(55.79828, 37.53710);
+  m2::PointD const final = MercatorBounds::FromLatLon(55.79956, 37.54115);
+  TestRouters(start, final);
+}
 }  // namespace


mpimenov:
LGTM

routing/routing_benchmarks/bicycle_routing_tests.cpp
```diff
@@ -44,4 +44,11 @@ UNIT_CLASS_TEST(BicycleTest, Smoke)
   m2::PointD const final = MercatorBounds::FromLatLon(55.79369, 37.56054);
   TestRouters(start, final);
 }
+
+UNIT_CLASS_TEST(BicycleTest, MailRuOfficeToUlSimanova5K2)
```
ygorshenin: Please, rename to Test1, or Moscow_Small1.
The reason is that MailRuOfficeToUlSimanova5K2 is a BAD address specification. Lat-lons is the test body a GOOD, but the name of the test is useless.

ygorshenin:
Both bicycle tests pass on my machine, moreover, bicycle routing works fine between points provided by you. May be something wrong with your environment?

routing/routing_benchmarks/bicycle_routing_tests.cpp
bykoianko: I agree MailRuOfficeToUlSimanova5K2 is a bad address specification. I suggest using name specification which I use in routing_integration_tests: https://github.com/mapsme/omim/pull/4307

bykoianko:
> Both bicycle tests pass on my machine, moreover, bicycle routing works fine between points provided by you. May be something wrong with your environment?

I don't think so. The thing is I reproduced this behavior on my iPhone, Android and my Mac. Which map are you using? (I'm using 11.09.2016 / Current master). On the other had I cannot reproduce it with 11.09.2016/ release-64.

PR #4307, opened 2016-09-19
Renaming a test.

bykoianko:
@ygorshenin PTAL

### DIFF
--- a/routing/routing_benchmarks/bicycle_routing_tests.cpp
+++ b/routing/routing_benchmarks/bicycle_routing_tests.cpp
@@ -45,7 +45,7 @@ UNIT_CLASS_TEST(BicycleTest, Smoke)
   TestRouters(start, final);
 }
 
-UNIT_CLASS_TEST(BicycleTest, MailRuOfficeToUlSimanova5K2)
+UNIT_CLASS_TEST(BicycleTest, RussiaMoscow_Test1)
 {
   m2::PointD const start = MercatorBounds::FromLatLon(55.79828, 37.53710);
   m2::PointD const final = MercatorBounds::FromLatLon(55.79956, 37.54115);


routing/routing_benchmarks/bicycle_routing_tests.cpp
```diff
@@ -45,7 +45,7 @@ UNIT_CLASS_TEST(BicycleTest, Smoke)
   TestRouters(start, final);
 }
 
-UNIT_CLASS_TEST(BicycleTest, MailRuOfficeToUlSimanova5K2)
```
ygorshenin: s/RussiaMoscow.../RussiaMoscow_Test1/

bykoianko: As I wrote answering your comment in 4300 now we use in routing_integration_tests test names with full address: <Country><City><Street>... . Doesn't it give us less info than <Country><Cicy>_TestXX?

bykoianko: So I suggest using the same notation hear in routing_benchmarks.

ygorshenin: It does not give any useful info.

bykoianko: It does give. Just have a look at osrm_turn_test.cpp. There're about 30 tests there. I wrote them more than one year ago but even now I can recall the test cases and theirs places by theirs names and theirs code for most of them. I can use lat lon coords for recovering the place as well but having such descriptive names I can take a short cut. Do you really think that address or short description is not useful at all? As for me it looks like a short comment for test which helps recall the test case quickly.

bykoianko:
@ygorshenin PTAL

routing/routing_benchmarks/bicycle_routing_tests.cpp
ygorshenin: It doesn't. Yesterday I tried to reproduce the problem in the second case, on map. I tapped on the Mail.ru office, then I tapped on the "Domumetro" hotel. And the bicycle router was able to construct a route between these two points. On the other hand, when I tried to construct a route between two points from the test body (exact lat-lons), the bicycle router failed. So, my point is that these text descriptions are not helpful, they're misleading. Because in any case, when you try to reproduce a problem in a failing test, you should use lat lons from the test body, not names from the test name. Please, take a look at pedestrian tests in the nearby file - all tests are named in a simple manner, on the other hand if you are interested in how these routes look like, all what you need is just take exact coords from test bodies and feed them to a geocoder.

ygorshenin: Our search engine is able to search by lat-lons, routers expect start and final points as a couple of doubles, not text descriptions.

ygorshenin: To sum up, I'm against such notation from osrm integration tests.

bykoianko:
@ygorshenin PTAL&merge

ygorshenin:
LGTM, thanks!

PR #4318, opened 2016-09-19
Crash fix in CheckUTurnOnRoute

bykoianko:
Крешфис при построении маршрута на карте Санкт-Петербурга. (В CheckUTurnOnRoute)
https://jira.mail.ru/browse/MAPSME-1716

@ygorshenin @mpimenov PTAL

### DIFF
--- a/routing/routing_integration_tests/osrm_route_test.cpp
+++ b/routing/routing_integration_tests/osrm_route_test.cpp
@@ -281,4 +281,22 @@ namespace
     TEST_EQUAL(result, IRouter::NoError, ());
     TEST_LESS(route.GetTotalTimeSec(), numeric_limits<uint32_t>::max() / 2, ());
   }
+
+// There are road ids in osrm which don't have appropriate features ids in mwm.
+// When the route goes through such osrm id a code line with LOG(LERROR, ... is executed:
+// on route reconstruction stage. As a result some item of |segments| vector could have an empty |m_path|.
+// This test shows such case. It's commented because if to uncomment it debug version of
+// routing_integration_tests would crash.
+//  UNIT_TEST(RussiaSpbPloschadBekhterevaToKomendantskiyProspekt)
+//  {
+//    TRouteResult const routeResult = integration::CalculateRoute(
+//        integration::GetOsrmComponents(), MercatorBounds::FromLatLon(59.90126, 30.39970), {0., 0.},
+//        MercatorBounds::FromLatLon(60.02499, 30.23889));
+
+//    Route const & route = *routeResult.first;
+//    IRouter::ResultCode const result = routeResult.second;
+//    TEST_EQUAL(result, IRouter::NoError, ());
+
+//    integration::TestRouteTime(route, 1364.0);
+//  }
 }  // namespace
--- a/routing/routing_integration_tests/osrm_route_test.cpp
+++ b/routing/routing_integration_tests/osrm_route_test.cpp
@@ -283,11 +283,9 @@ namespace
   }
 
   // There are road ids in osrm which don't have appropriate features ids in mwm.
-  // When the route goes through such osrm id a code line with LOG(LERROR, ... is executed:
-  // on route reconstruction stage. As a result some item of |segments| vector could have an empty
-  // |m_path|.
-  // This test shows such case. It's commented because if to uncomment it debug version of
-  // routing_integration_tests would crash.
+  // When the route goes through such osrm id a code line with LOG(LERROR, ...) is executed
+  // on route reconstruction stage. As a result some items of |segments| vector could have an empty
+  // |m_path|. This test shows such case.
   //  UNIT_TEST(RussiaSpbPloschadBekhterevaToKomendantskiyProspekt)
   //  {
   //    TRouteResult const routeResult = integration::CalculateRoute(
--- a/routing/routing_integration_tests/osrm_route_test.cpp
+++ b/routing/routing_integration_tests/osrm_route_test.cpp
@@ -282,21 +282,23 @@ namespace
     TEST_LESS(route.GetTotalTimeSec(), numeric_limits<uint32_t>::max() / 2, ());
   }
 
-// There are road ids in osrm which don't have appropriate features ids in mwm.
-// When the route goes through such osrm id a code line with LOG(LERROR, ... is executed:
-// on route reconstruction stage. As a result some item of |segments| vector could have an empty |m_path|.
-// This test shows such case. It's commented because if to uncomment it debug version of
-// routing_integration_tests would crash.
-//  UNIT_TEST(RussiaSpbPloschadBekhterevaToKomendantskiyProspekt)
-//  {
-//    TRouteResult const routeResult = integration::CalculateRoute(
-//        integration::GetOsrmComponents(), MercatorBounds::FromLatLon(59.90126, 30.39970), {0., 0.},
-//        MercatorBounds::FromLatLon(60.02499, 30.23889));
+  // There are road ids in osrm which don't have appropriate features ids in mwm.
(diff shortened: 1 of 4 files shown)


mpimenov:
Интеграционный тест ты не стал добавлять в итоге?

routing/turns_generator.cpp
```diff
@@ -668,6 +668,8 @@ size_t CheckUTurnOnRoute(TUnpackedPathSegments const & segments,
   auto const & masterSegment = segments[currentSegment - 1];
   if (masterSegment.m_path.size() < 2)
     return 0;
+  if (segments[currentSegment].m_path.empty())
```
ygorshenin: Could you please explain why do we need this check?

bykoianko: There's an issue in our system. There are road ids in osrm which don't have appropriate ids in mwm. When the route goes through such osrm id we get     
```
LOG(LERROR, ("Can't unpack geometry for map:", mapping.GetCountryName(), " node: ",
                 osrmPathSegment.node)); 
```
on route reconstruction stage. As a result some item of |segments| vector could have an empty |m_path|. It leads to crashes on release version.

Probably the best solution is to have for every osrm id an appropriate road id in mwm. As far as I remember Lev has investigated the problem and told me that there're some cases when osrm edge cannot have a corresponding road id in mwm. Anyway it's a topic for further investigation.

For the time being our code should rely that some orsm edges doesn't have a corresponding features. (At least in release version.)

bykoianko:
@ygorshenin @mpimenov PTAL

routing/turns_generator.cpp
kshalnev: A few lines below:
...
auto const & checkedSegment = segments[currentSegment + i];
...
if (path[path.size() - 2] == checkedSegment.m_path[1])
...
Why you don't check that next segment's m_path is not empty before access 1st element?
May be need to add similar check?

kshalnev: Small note: CheckUTurnOnRoute - does it return 1-based index of segment and 0 is some non-invalid value? Would be good to have short description about the function.

bykoianko: @kshalnev 
1. fixed, 2. done.

bykoianko:
@mpimenov @ygorshenin @kshalnev PTAL

routing/turns_generator.cpp
```diff
@@ -668,12 +668,17 @@ size_t CheckUTurnOnRoute(TUnpackedPathSegments const & segments,
...
+    return 0;
   // Roundabout is not the UTurn.
   if (masterSegment.m_onRoundabout)
     return 0;
   for (size_t i = 0; i < kUTurnLookAhead && i + currentSegment < segments.size(); ++i)
   {
     auto const & checkedSegment = segments[currentSegment + i];
+    if (checkedSegment.m_path.empty())
```
ygorshenin: What if checkedSegment.m_path size is less than two? Below, I see ```checkdSegment.m_path[1]```.

routing/turns_generator.hpp
```diff
@@ -130,9 +130,10 @@ TurnDirection GetRoundaboutDirection(bool isIngoingEdgeRoundabout, bool isOutgoi
 void GetTurnDirection(IRoutingResult const & result, turns::TurnInfo & turnInfo, TurnItem & turn);
 
 /*!
- * \brief Finds an UTurn that starts from current segment and returns how many segments it lasts.
- * Returns 0 if there is no UTurn.
- * Warning! currentSegment must be greater than 0.
+ * \brief Finds an uturn that starts from master segment and returns how many segments it lasts.
```
mpimenov: U-turn

routing/turns_generator.cpp
bykoianko: Right. The same thing with size of |segments[currentSegment].m_path|.
Fixed.

routing/turns_generator.hpp
bykoianko: Yes. Fixed.

kshalnev:
LGTM

bykoianko:
@ygorshenin @mpimenov PTAL

mpimenov:
Думаю, стоит добавить тест хотя бы в закомментированном виде, чтобы он не потерялся.

routing/turns_generator.cpp
```diff
@@ -668,12 +668,17 @@ size_t CheckUTurnOnRoute(TUnpackedPathSegments const & segments,
   auto const & masterSegment = segments[currentSegment - 1];
   if (masterSegment.m_path.size() < 2)
     return 0;
+  if (segments[currentSegment].m_path.size() < 2)
```
ygorshenin: This check will be duplicated on the first iteration of the loop below. Do we need it here?

bykoianko: Let's remove this check.

bykoianko:
> Думаю, стоит добавить тест хотя бы в закомментированном виде, чтобы он не потерялся.

Давай добавим закомментированным.

bykoianko:
@ygorshenin @mpimenov PTAL

routing/routing_tests/turns_generator_test.cpp
```diff
@@ -4,8 +4,9 @@
 #include "routing/turns.hpp"
 #include "routing/turns_generator.hpp"
 
-#include "geometry/mercator.hpp"
+#include "indexer/ftypes_matcher.hpp"
```
ygorshenin: Why do you need this here?

routing/routing_integration_tests/osrm_route_test.cpp
```diff
@@ -281,4 +281,22 @@ namespace
     TEST_EQUAL(result, IRouter::NoError, ());
     TEST_LESS(route.GetTotalTimeSec(), numeric_limits<uint32_t>::max() / 2, ());
   }
+
+// There are road ids in osrm which don't have appropriate features ids in mwm.
+// When the route goes through such osrm id a code line with LOG(LERROR, ... is executed:
```
mpimenov: Скобка после многоточия?

routing/routing_integration_tests/osrm_route_test.cpp
```diff
@@ -281,4 +281,22 @@ namespace
...
+// on route reconstruction stage. As a result some item of |segments| vector could have an empty |m_path|.
```
mpimenov: s/item/items/

routing/routing_tests/turns_generator_test.cpp
bykoianko: Because ftypes::HighwayClass::Trunk defined in ftypes_matcher.hpp is used in this file below.

routing/routing_integration_tests/osrm_route_test.cpp
```diff
@@ -281,4 +281,22 @@ namespace
...
+// This test shows such case. It's commented because if to uncomment it debug version of
```
mpimenov: Не надо. И так понятно, что мы комментируем те тесты, которые не проходят.

routing/routing_integration_tests/osrm_route_test.cpp
bykoianko: Да

ygorshenin:
LGTM

routing/routing_integration_tests/osrm_route_test.cpp
bykoianko: Fixed

routing/routing_integration_tests/osrm_route_test.cpp
bykoianko: Ok

bykoianko:
@mpimenov PTAL&merge

mpimenov:
LGTM

bykoianko:
Всем спасибо за ревью!

PR #4341, opened 2016-09-21
Mapsme 2399 save altitude chart state on tablet

alexzatsepin:


alexzatsepin:
@yunikkk @trashkalmar PTAL

trashkalmar:
LGTM

PR #4375, opened 2016-09-23
[Android] EAT display in navigation mode change when switching between landscape/portrait.

fijemax:
When i change the EAT information to display the arrival hour and i switch the ui into landscape/portait mode, the EAT information go back to display the remaining time.

Example : 
device-2016-09-23-164426

[Only under Android]

### DIFF
--- a/android/src/com/mapswithme/maps/MwmActivity.java
+++ b/android/src/com/mapswithme/maps/MwmActivity.java
@@ -124,6 +124,7 @@ public class MwmActivity extends BaseMwmFragmentActivity
   private PlacePageView mPlacePage;
 
   private RoutingPlanInplaceController mRoutingPlanInplaceController;
+  @Nullable
   private NavigationController mNavigationController;
 
   private MainMenu mMainMenu;
@@ -768,6 +769,9 @@ public class MwmActivity extends BaseMwmFragmentActivity
         fragment.saveAltitudeChartState(outState);
     }
 
+    if (mNavigationController != null)
+      mNavigationController.onSaveState(outState);
+
     RoutingController.get().onSaveState();
     super.onSaveInstanceState(outState);
   }
@@ -782,6 +786,9 @@ public class MwmActivity extends BaseMwmFragmentActivity
 
     if (!mIsFragmentContainer && RoutingController.get().isPlanning())
       mRoutingPlanInplaceController.restoreState(savedInstanceState);
+
+    if (mNavigationController != null)
+      mNavigationController.onRestoreState(savedInstanceState);
   }
 
   @Override
--- a/android/src/com/mapswithme/maps/routing/NavigationController.java
+++ b/android/src/com/mapswithme/maps/routing/NavigationController.java
@@ -4,6 +4,8 @@ import android.app.Activity;
 import android.content.Intent;
 import android.location.Location;
 import android.os.Build;
+import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.text.TextUtils;
 import android.util.Pair;
 import android.view.View;
@@ -31,6 +33,8 @@ import com.mapswithme.util.statistics.Statistics;
 
 public class NavigationController
 {
+  private static final String STATE_SHOW_TIME_LEFT = "ShowTimeLeft";
+
   private final View mFrame;
   private final View mBottomFrame;
   private final NavMenu mNavMenu;
@@ -281,4 +285,15 @@ public class NavigationController
   {
     return mNavMenu;
   }
+
+  public void onSaveState(@NonNull Bundle outState)
+  {
+    outState.putBoolean(STATE_SHOW_TIME_LEFT, mShowTimeLeft);
+  }
+
+  public void onRestoreState(@NonNull Bundle savedInstanceState)
+  {
+    mShowTimeLeft = savedInstanceState.getBoolean(STATE_SHOW_TIME_LEFT);
+  }
+
 }


burivuh:
@alexzatsepin and @yunikkk please look at this PR

yunikkk:
LGTM

Zverik:
LGTM

android/src/com/mapswithme/maps/routing/NavigationController.java
```diff
@@ -59,7 +59,7 @@
 
   private final SearchWheel mSearchWheel;
 
-  private boolean mShowTimeLeft = true;
+  private static boolean mShowTimeLeft = true;
```
alexzatsepin: I'm really concerned about this fix. If problem was appeared during the rotation screen then I would add additional methods in this class such as saveState/restorState and call it from MwmActivity which is owner of this controller. I think it would be much better solution then declaring the variable static. @fijemax  what do you think (maybe I'm wrong)?

fijemax: Hi @alexzatsepin , I totally  agree with you. What do you think about my last change ?

android/src/com/mapswithme/maps/MwmActivity.java
```diff
@@ -768,6 +769,11 @@ protected void onSaveInstanceState(Bundle outState)
         fragment.saveAltitudeChartState(outState);
     }
 
+    if(mNavigationController != null)
+    {
+      outState.putBoolean(STATE_SHOW_TIME_LEFT, mNavigationController.saveState());
```
alexzatsepin: @fijemax  please take a look at the 762 line `mRoutingPlanInplaceController.onSavedState(outState)`, i.e. just pass the whole bundle, `outState`, to the `mNavigationController.saveState(@NonNulle Bundle outState)`  to let the navigation controller to save its state as it wants.

Also, please rename the `mNavigationController.saveState()` method to `mNavigationController.onSaveState(@NonNull Bundle outState)` so that to be consistent with other code design/style.

android/src/com/mapswithme/maps/MwmActivity.java
```diff
@@ -782,6 +788,11 @@ protected void onRestoreInstanceState(@NonNull Bundle savedInstanceState)
 
     if (!mIsFragmentContainer && RoutingController.get().isPlanning())
       mRoutingPlanInplaceController.restoreState(savedInstanceState);
+
+    if(mNavigationController != null)
+    {
+      mNavigationController.restoreState(savedInstanceState.getBoolean(STATE_SHOW_TIME_LEFT));
```
alexzatsepin: The same here. Please pass the whole state bundle to the `mRoutingPlanInplaceController.restoreState(@NonNull Bundle savedState)`

android/src/com/mapswithme/maps/MwmActivity.java
```diff
@@ -113,6 +113,7 @@
   // Instance state
   private static final String STATE_PP_OPENED = "PpOpened";
   private static final String STATE_MAP_OBJECT = "MapObject";
+  private static final String STATE_SHOW_TIME_LEFT = "ShowTimeLeft";
```
alexzatsepin: When you pass the whole bundle state (see my notes below) to the mNavigationController class you will be able to move this constant to NavigationController class (for example, take a look at the `RoutingPlanInplaceController `)

android/src/com/mapswithme/maps/routing/NavigationController.java
```diff
@@ -281,4 +281,15 @@ public NavMenu getNavMenu()
...
+
+  public boolean saveState()
+  {
+    return mShowTimeLeft;
+  }
+
+  public void restoreState(boolean showTimeLeft)
+  {
+    mShowTimeLeft = showTimeLeft;
+  }
```
alexzatsepin: Please change the signature and name of methods according my comments in the MwmActivity class.

android/src/com/mapswithme/maps/MwmActivity.java
```diff
@@ -768,6 +769,11 @@ protected void onSaveInstanceState(Bundle outState)
         fragment.saveAltitudeChartState(outState);
     }
 
+    if(mNavigationController != null)
```
alexzatsepin: Please download the out code style scheme https://www.dropbox.com/s/b6xbx5ajryh7rhx/maps.me.xml?dl=0 and apply it to your Android Studio. You just need to place this xml file into this directory /Users/<your_user>/Library/Preferences/AndroidStudio2.2/codestyles/maps.me.xml (if your OS is MacOs, if not - please google it) . After that you should restart your Android studio and choose this scheme here http://clip2net.com/s/3CLs8ZA 

Of course, it's not a requirement and you can fix formatting manually, but probably you find it convenient.

P.S: white space are missed. And we don't put brackets with if condition when the body is just single line (take a look at the same place in our code please)

android/src/com/mapswithme/maps/MwmActivity.java
```diff
@@ -782,6 +788,11 @@ protected void onRestoreInstanceState(@NonNull Bundle savedInstanceState)
 
     if (!mIsFragmentContainer && RoutingController.get().isPlanning())
       mRoutingPlanInplaceController.restoreState(savedInstanceState);
+
+    if(mNavigationController != null)
```
alexzatsepin: here the same, white spaces are missed and brackets should be removed

android/src/com/mapswithme/maps/routing/NavigationController.java
alexzatsepin: Hey @fijemax , now it's much better. There are some notes and I put all them in comments, please read it.

android/src/com/mapswithme/maps/MwmActivity.java
fijemax: Thank you for the xml style file.

fijemax:
Hi, @alexzatsepin, thanks for your reviews.
I changed respecting all your notes.
I don't use another version control system, my fork is on github too and i use git.
My fork : https://github.com/Mapotempo/omim
My branch : https://github.com/Mapotempo/omim/tree/mt-ui-remaining-time

android/src/com/mapswithme/maps/MwmActivity.java
alexzatsepin: the same is here. White space is missed.

android/src/com/mapswithme/maps/MwmActivity.java
alexzatsepin: Very good. Please select this code  and format it, **only code which written by you** not whole file, because the white space between work 'if' and '(' is still missed. Fix it please.

alexzatsepin:
Also, please add the @Nullable annotation to the MwmActivity.mNavigationController as well

fijemax:
Sorry, i forgot to format.
It's done.

alexzatsepin:
LGTM. @fijemax one advice: when you add the changes to existing pull request leave the comment please like "@reviewer_name PTAL" , i.e. @alexzatsepin PTAL . It automatically triggers the letter to my mailbox and my reaction will be faster. Great Job!

alexzatsepin:
@fijemax I see you left comment at the same time as me, so it's okay. Any comment (not especially with PTAL abr) will attend my attention to pull request. Just leave the comment after some changes made

fijemax:
Ok I would think for the next time. Thank you.

PR #4429, opened 2016-10-03
fix tile text disappearing in DRAW_TILE_NET mode

Dushistov:
Point<float> not init any fields in default constructor,
so TextViewParams::m_primaryOffset can have any value
(depending on previous stack content), which leads to
captions disappearing in DRAW_TILE_NET debug mode.


Zverik:
@rokuz посмотри плз

PR #4430, opened 2016-10-03
[android] Added the missed http link into the sharing message

alexzatsepin:


android/src/com/mapswithme/util/sharing/MapObjectShareable.java
```diff
@@ -18,26 +18,31 @@
     super(context);
 
     final Activity activity = getActivity();
-    final String ge0Url = Framework.nativeGetGe0Url(mapObject.getLat(), mapObject.getLon(), mapObject.getScale(), mapObject.getTitle());
+    final double lan = mapObject.getLat();
```
trashkalmar: `lat`

alexzatsepin: fixed

PR #4459, opened 2016-10-09
fix build with clang/libcxx on linux

### LINKED ISSUE #3807: compilation failure: illegal max(float,double) in drape_frontend/text_layout.cpp
I managed to natively build **MAPS.ME** executable on a Gentoo/x86_64 desktop.

I had to solve some problems one of which is described below.

In line 563 of **drape_frontend/text_layout.cpp** expression `max(floor(pathLength / minPeriodSize), 1.0)` is evaluated. This expression takes max(float,double) which is not supported by C++11 dialect. The line does not compile with clang 3.8.0-r3.

I fixed the issue by converting 1.0 to float:
```
sed -re 's:1\.0[)]:1.0f):' -i drape_frontend/text_layout.cpp
```
I suggest that MAPS.ME team fix **text_layout.cpp**.

Dushistov:
This closes #3807 issue. Because of clang/libcxx 3.8 have overload of
`floor` function in math.h for `float` and `double` we got compilation
error, but obvious fix:
1.0 -> 1.0f
not works, because of at least on Android we have no `floor(float)`
overload. So to fix build force usage of floor(double) which exists on every platform.

### DIFF
--- a/drape_frontend/text_layout.cpp
+++ b/drape_frontend/text_layout.cpp
@@ -561,7 +561,7 @@ void PathTextLayout::CalculatePositions(vector<float> & offsets, float splineLen
   }
   else
   {
-    double const textCount = max(floor(pathLength / minPeriodSize), 1.0);
+    double const textCount = max(floor(static_cast<double>(pathLength / minPeriodSize)), 1.0);
     double const glbTextLen = splineLength / textCount;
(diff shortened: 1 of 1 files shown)


Zverik:
LGTM
@syershov check plz

rokuz:
LGTM

syershov:
LGTM

PR #4467, opened 2016-10-10
French spelling correction

fijemax:
French spelling correction


mpimenov:
I'm fine with the change but our current state is apparentely broken.
@therearesomewhocallmetim We have not run the script after the latest PR, have we?

therearesomewhocallmetim:
We haven't.

PR #4492, opened 2016-10-12
[search] JNI layer for HotelsFilter.

ygorshenin:
@mpimenov @alexzatsepin PTAL

### DIFF
--- a/android/jni/com/mapswithme/maps/SearchEngine.cpp
+++ b/android/jni/com/mapswithme/maps/SearchEngine.cpp
@@ -135,10 +135,8 @@ private:
 
     switch (field)
     {
-    case FIELD_RATING:
-      return BuildRatingOp(env, op, filter);
-    case FIELD_PRICE_RATE:
-      return BuildPriceRateOp(env, op, filter);
+    case FIELD_RATING: return BuildRatingOp(env, op, filter);
+    case FIELD_PRICE_RATE: return BuildPriceRateOp(env, op, filter);
     }
 
     LOG(LERROR, ("Unknown field:", field));
@@ -153,11 +151,11 @@ private:
 
     switch (op)
     {
-      case OP_LT: return Lt<Rating>(rating);
-      case OP_LE: return Le<Rating>(rating);
-      case OP_GT: return Gt<Rating>(rating);
-      case OP_GE: return Ge<Rating>(rating);
-      case OP_EQ: return Eq<Rating>(rating);
+    case OP_LT: return Lt<Rating>(rating);
+    case OP_LE: return Le<Rating>(rating);
+    case OP_GT: return Gt<Rating>(rating);
+    case OP_GE: return Ge<Rating>(rating);
+    case OP_EQ: return Eq<Rating>(rating);
     }
 
     LOG(LERROR, ("Unknown op:", op));
@@ -172,11 +170,11 @@ private:
 
     switch (op)
     {
-      case OP_LT: return Lt<PriceRate>(priceRate);
-      case OP_LE: return Le<PriceRate>(priceRate);
-      case OP_GT: return Gt<PriceRate>(priceRate);
-      case OP_GE: return Ge<PriceRate>(priceRate);
-      case OP_EQ: return Eq<PriceRate>(priceRate);
+    case OP_LT: return Lt<PriceRate>(priceRate);
+    case OP_LE: return Le<PriceRate>(priceRate);
+    case OP_GT: return Gt<PriceRate>(priceRate);
+    case OP_GE: return Ge<PriceRate>(priceRate);
+    case OP_EQ: return Eq<PriceRate>(priceRate);
     }
 
     LOG(LERROR, ("Unknown op:", op));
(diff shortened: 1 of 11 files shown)


syershov:
DUDE JTTP

android/jni/com/mapswithme/maps/SearchEngine.cpp
```diff
@@ -180,16 +364,19 @@ extern "C"
...
   JNIEXPORT jboolean JNICALL
   Java_com_mapswithme_maps_search_SearchEngine_nativeRunSearch(JNIEnv * env, jclass clazz, jbyteArray bytes, jstring lang,
-                                                               jlong timestamp, jboolean hasPosition, jdouble lat, jdouble lon)
+                                                               jlong timestamp, jboolean hasPosition, jdouble lat, jdouble lon, jobject hotelsFilter)
   {
     search::EverywhereSearchParams params;
     params.m_query = jni::ToNativeString(env, bytes);
     params.m_inputLocale = ReplaceDeprecatedLanguageCode(jni::ToNativeString(env, lang));
     params.m_onResults = bind(&OnResults, _1, timestamp, false, hasPosition, lat, lon);
+    params.m_hotelsFilter = g_hotelsFilterBuilder.Build(env, hotelsFilter);
```
alexzatsepin: The `hotelsFilter` is a local reference, and this reference is valid only duration the method call according to JNI specification - http://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/design.html#wp1242 . That's why, I'm pretty sure this reference should be converted to global one first (`env->NewGlobalRef(env, hotelsFilter)`) and then it can be passed in to the core code. Also, a good practise is to delete this global reference when it's no needed anymore (env->DeleteGlobalRef(g_hotelsFilter)) or if we know that this object should be in memory during application live we can skip deletion I guess.

Probably I missed something and wrong about invalid reference to hotelsFilter, but it looks to me so so far. Let me know if I'm wrong

alexzatsepin: hm..or this reference doesn't go to core code and should live only during jni native call ?

alexzatsepin: I see that it uses to get values filter fields . Probably I was wrong and local visibility is ok for this reference

ygorshenin: Yes, hotelsFilter is not passed to the core. HotelsFilterBuilder uses it just to build C++ representation for the filter.

android/src/com/mapswithme/maps/search/HotelsFilter.java
```diff
@@ -0,0 +1,90 @@
+package com.mapswithme.maps.search;
+
+import android.support.annotation.NonNull;
+
+public class HotelsFilter
```
ygorshenin: Yes, we can. But I saw a lot of places in our Java code where enumerations are implemented just like a set of named constants, not as enum.

ygorshenin: Do you insist on switching to enums here?

alexzatsepin: No, I don't. To be honest, there is advice from google about "Avoid using enums, use the constants instead", because enums are the classes and can consume memory. But it maybe reasonable for last times, when devices had a 16 mb of RAM. Also, they suggest to use "Typedef Annotations" like a trade-off for using the "Enums". You can try to use it, but it's up to you - https://developer.android.com/studio/write/annotations.html (Typedef Annotations)

android/src/com/mapswithme/maps/search/HotelsFilter.java
```diff
@@ -0,0 +1,90 @@
...
+    {
+      super(TYPE_AND);
+      mLhs = lhs;
+      mRhs = rhs;
+    }
+  }
+
+  public static class Or extends HotelsFilter
+  {
+    @NonNull public final HotelsFilter mLhs;
```
alexzatsepin: Could you put nullness annotation above of field (on new string) please:

```
@NonNull
public final HotelsFilter mLhs;
```
? and do this for rest nullnes annotations in this file?

alexzatsepin:
LGTM (if you fix nullness annotations position it would be great)

android/src/com/mapswithme/maps/search/HotelsFilter.java
ygorshenin: Okay, let's leave these constants as is.

android/src/com/mapswithme/maps/search/HotelsFilter.java
ygorshenin: No problem, also, I can add a corresponding entry to .clang-format.

ygorshenin:
@mpimenov PTAL

android/src/com/mapswithme/maps/search/HotelsFilter.java
alexzatsepin: Great! Thanks

mpimenov:
LGTM

ygorshenin:
@syershov PTAL please.

syershov:
LGTM

PR #4596, opened 2016-10-31
[android] Fix profile verified changes text color.

goblinr:
Fix profile verified changes text color.

@alexzatsepin 
@mpimenov


mpimenov:
LGTM

PR #4602, opened 2016-10-31
[android] clearing search results when the navigation session is closed.

goblinr:
Closing the interactive search when the navigation menu is hidden.

@alexzatsepin 
@mpimenov


mpimenov:
LGTM

PR #4664, opened 2016-11-06
[styles] Use native icons for shop=* in 16, 17 levels.

vng:
For better check what happened, see drules_proto*.txt diff.


bykoianko:
LGTM
@igortomko PTAL

igortomko:
LGTM

PR #4673, opened 2016-11-07
[android] Fixed the random opening/closing the search bar during navi…

alexzatsepin:
…gation

@mpimenov @milchakov  PTAL please


milchakov:
LGTM

android/src/com/mapswithme/maps/routing/SearchWheel.java
```diff
@@ -36,6 +36,10 @@
     @Override
     public void run()
     {
+      //if the search bar is already closed, i.e. nothing should be done here
```
mpimenov: ```// If the search bar is already closed, i.e. nothing should be done here.```

mpimenov:
LGTM after fix

PR #4688, opened 2016-11-09
[android] Removed elevation from toolbar

goblinr:
Removed elevation from toolbar(http://jira.65apps.com/browse/MAP-74).

@alexzatsepin 
@mpimenov


mpimenov:
LGTM

PR #4694, opened 2016-11-10
[DNM] [android] Updated NDK to 12

alexzatsepin:
@therearesomewhocallmetim @rokuz @syershov @bykoianko

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -220,7 +220,7 @@ android {
   splits.abi {
     enable true
     reset()
-    include 'x86', 'armeabi-v7a'
+    include 'armeabi-v7a'
     universalApk true
   }
 
@@ -325,7 +325,7 @@ project.ext.NDK_BUILD = android.getNdkDirectory().toString() + '/ndk-build'
 if (System.properties['os.name'].toLowerCase().contains('windows'))
   project.ext.NDK_BUILD += ".cmd"
 
-def archs = ['x86', 'armeabi-v7a-hard']
+def archs = ['armeabi-v7a']
 def buildTypes = [[ndkType: 'release', cppType: "production", flags: propReleaseNdkFlags], [ndkType: 'debug', cppType: "debug", flags: propDebugNdkFlags]]
 
 buildTypes.each { type ->
--- a/android/jni/Application.mk
+++ b/android/jni/Application.mk
@@ -1,4 +1,4 @@
-NDK_TOOLCHAIN_VERSION := clang3.6
+NDK_TOOLCHAIN_VERSION := clang
 APP_PLATFORM := android-15
 APP_STL := c++_static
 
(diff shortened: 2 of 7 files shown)


android/jni/Application.mk
```diff
@@ -8,7 +8,7 @@ APP_STL := c++_static
 APP_CPPFLAGS += -Wno-deprecated-register
 
 ifeq (x$(NDK_ABI_TO_BUILD), x)
-  APP_ABI := armeabi-v7a-hard x86
+  APP_ABI := armeabi-v7a
```
Dushistov: Why remove `x86` here?

android/build.gradle
```diff
@@ -220,7 +220,7 @@ android {
   splits.abi {
     enable true
     reset()
-    include 'x86', 'armeabi-v7a'
+    include 'armeabi-v7a'
```
Dushistov: And why here remove x86?

alexzatsepin: For the same reason. It's temporary. First I wanted to update NDK at least for arm architecture to get it (https://developer.android.com/studio/projects/add-native-code.html ) work

Dushistov:
@alexzatsepin Not get it. I have similar patch, but for NDK 13. It has several additional changes in compare with this pull request and related to diff 12..13, but not many.

And it compiles fine for arm and for x86, also it works in emulator/x86. 

So what you want to achive, ability to debug C++ code in Android Studio?

>Or probably we will remove x86 completely if the arm app works fine on x86 in emulation mode

qemu that emulates arm on x86 is so slow, any news that anybody speedup qemu of the order of magnitude?

alexzatsepin:
> So what you want to achive, ability to debug C++ code in Android Studio?

Exactly. Plus we need to update NDK to newer version anyway.

`qemu that emulates arm on x86 is so slow, any news that anybody speedup qemu of the order of magnitude?`

Removing the x86 is definitely the last action which we will apply. Of course, we should try to keep x86. And the support of emulators is the really important thing, because as you said, the arm emulator is still slow.

I've never heard about 'qemu', if you clarify it for me I will be appreciate.

Dushistov:
@alexzatsepin 

>I've never heard about 'qemu', if you clarify it for me I will be appreciate.

Android emulator based on qemu opensource project: http://wiki.qemu.org/Main_Page
For example you can look at `Android/Sdk/tools/qemu` or run emulator and find "*qemu*" process in `top`.

tools/mkspecs/android-g++/qmake.conf
```diff
@@ -63,7 +63,7 @@ contains(ANDROID_TARGET_ARCH, arm) {
     QMAKE_CFLAGS += -march=armv7-a -mfloat-abi=softfp -mfpu=vfpv3-d16
     QMAKE_LFLAGS += -march=armv7-a -Wl,--fix-cortex-a8
   } else {
-    contains(ANDROID_TARGET_ABI, armeabi-v7a-hard) {
```
Dushistov: I compare your patch with mine line by line, and
at now it looks strange:
contains(ANDROID_TARGET_ABI, armeabi-v7a) {
...
  } else {
    contains(ANDROID_TARGET_ABI, armeabi-v7a) {
  ... 
   } else {
...

why two times handle `contains(ANDROID_TARGET_ABI, armeabi-v7a) ` ?

alexzatsepin:
Updated to ndk 13 already

PR #4768, opened 2016-11-21
[strings] Fixed Danish translation on [book_button]

bennybechdk:


Zverik:
Are you sure?

bennybechdk:
I am sure (Danish is my native language).

Bog (in danish) is a "bunch of papers for reading" (also the english word book).
To be convinced you can se the same use of book (instead of bog) in  [sharing_booking]

Thanks for your concern

Zverik:
Okay, thanks.

PR #4772, opened 2016-11-21
Do not select streets on tap.

darina:
https://jira.mail.ru/browse/MAPSME-3033

### DIFF
--- a/drape/overlay_tree.cpp
+++ b/drape/overlay_tree.cpp
@@ -395,7 +395,7 @@ void OverlayTree::Select(m2::PointD const & glbPoint, TOverlayContainer & result
 
   for (auto const & handle : m_handlesCache)
   {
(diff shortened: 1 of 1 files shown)


razor74:
This is a problem and aditionally " do not select big poligons on a tap...

drape/overlay_tree.cpp
```diff
@@ -395,7 +395,7 @@ void OverlayTree::Select(m2::PointD const & glbPoint, TOverlayContainer & result
 
   for (auto const & handle : m_handlesCache)
   {
-    if (rect.IsPointInside(handle->GetPivot(screen, false)))
+    if (!handle->HasDynamicAttributes() && rect.IsPointInside(handle->GetPivot(screen, false)))
```
darina: Yes, it’s ok. Select(by Point) is used for finding z coordinate at that point and here we need to check all the overlays.

PR #4792, opened 2016-11-23
[3party] Opening hours compile speedup

syershov:


ygorshenin:
LGTM, thanks for doing that!

PR #4870, opened 2016-12-05
Implemented read-write mutex

rokuz:


base/rw_mutex.hpp
```diff
@@ -0,0 +1,65 @@
...
+public:
+  RWMutex() : m_counter(0) {}
+  void LockRead()
+  {
+    std::lock_guard<mutex> lockWriter(m_noWritersMutex);
+    {
+      std::lock_guard<mutex> lockCounter(m_counterMutex);
+      size_t const prevCounter = m_counter;
+      m_counter++;
+      if (prevCounter == 0)
```
ygorshenin: Imagine that two readers calls LockRead, and it's possible to determine that, say, the first reader calls m_noReadersMutex.lock(). Then, the first reader calls UnlockRead, but m_noReadersMutex is still locked by the first thread (because the first reader called m_noReadersMutex.lock). Then, the second reader calls UnlockRead, so, the second thread calls m_noReadersMutex.unlock(). In accordance with cpp documentation, this is UB (http://en.cppreference.com/w/cpp/thread/mutex/unlock).

ygorshenin: Is it possible to implement RWMutex via semaphores, like Preshing does? (https://github.com/preshing/cpp11-on-multicore/tree/master/common, http://preshing.com/20150316/semaphores-are-surprisingly-versatile/)

PR #4895, opened 2016-12-07
Teardown traffic manager before drape engine destruction.

darina:


darina:
@syershov @rokuz PTAL

rokuz:
LGTM

PR #4911, opened 2016-12-08
Fixed map tests crash on traffic manager destruction.

darina:


vega626 (approved):
Put in my PayPal account

PR #4932, opened 2016-12-09
[build] C++ core issue workaround.

ygorshenin:
http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1397


ygorshenin:
@syershov @therearesomewhocallmetim PTAL

PR #4968, opened 2016-12-12
Route traffic information.

bykoianko:
Сохранение в классе Route информации о трафике.

@rokuz @darina @ygorshenin @mpimenov PTAL

### DIFF
--- a/routing/bicycle_directions.cpp
+++ b/routing/bicycle_directions.cpp
@@ -1,8 +1,11 @@
 #include "routing/bicycle_directions.hpp"
 #include "routing/car_model.hpp"
+#include "routing/road_point.hpp"
 #include "routing/router_delegate.hpp"
 #include "routing/routing_result_graph.hpp"
 #include "routing/turns_generator.hpp"
+
+#include "traffic/traffic_info.hpp"
 
 #include "indexer/ftypes_matcher.hpp"
 #include "indexer/index.hpp"
@@ -14,6 +17,7 @@ namespace
 {
 using namespace routing;
 using namespace routing::turns;
+using namespace traffic;
 
 class RoutingResult : public IRoutingResult
 {
@@ -83,6 +87,7 @@ BicycleDirectionsEngine::BicycleDirectionsEngine(Index const & index) : m_index(
 void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction> const & path,
                                        Route::TTimes & times, Route::TTurns & turns,
                                        vector<Junction> & routeGeometry,
+                                       vector<traffic::TrafficInfo::RoadSegmentId> & routeSegs,
                                        my::Cancellable const & cancellable)
 {
   times.clear();
@@ -90,6 +95,7 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
   routeGeometry.clear();
   m_adjacentEdges.clear();
   m_pathSegments.clear();
+  routeSegs.clear();
 
   size_t const pathSize = path.size();
   if (pathSize == 0)
@@ -158,8 +164,17 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
     }
 
     LoadedPathSegment pathSegment;
+    // @TODO(bykoianko) This place should be fixed. Putting |prevJunction| and |currJunction|
+    // for every route edge leads that all route points are duplicated. It's because
+    // prevJunction == path[i - 1] and currJunction == path[i].
     if (inFeatureId.IsValid())
       LoadPathGeometry(inFeatureId, {prevJunction, currJunction}, pathSegment);
+    pathSegment.m_routeSegs = {
+      {inFeatureId.m_index,
+       static_cast<uint16_t>(routeEdges[i - 1].GetSegId()),
+       routeEdges[i - 1].IsForward() ? TrafficInfo::RoadSegmentId::kForwardDirection
+                                     : TrafficInfo::RoadSegmentId::kReverseDirection
+      }};
 
     m_adjacentEdges.insert(make_pair(inFeatureId.m_index, move(adjacentEdges)));
     m_pathSegments.push_back(move(pathSegment));
@@ -170,7 +185,13 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
   Route::TTimes turnAnnotationTimes;
   Route::TStreets streetNames;
 
-  MakeTurnAnnotation(resultGraph, delegate, routeGeometry, turns, turnAnnotationTimes, streetNames);
+  MakeTurnAnnotation(resultGraph, delegate, routeGeometry, turns, turnAnnotationTimes,
+                     streetNames, routeSegs);
+  // @TODO(bykoianko) The invariant below it's an issue but now it's so and it should be checked.
+  // The problem is every edge is added as a pair of point to route geometry.
+  // So all the points except for beginning and ending are duplicated. It should
+  // be fixed in the future.
+  ASSERT_EQUAL(routeGeometry.size(), 2 * routeSegs.size(), ());
 }
 
 Index::FeaturesLoaderGuard & BicycleDirectionsEngine::GetLoader(MwmSet::MwmId const & id)
--- a/routing/bicycle_directions.cpp
+++ b/routing/bicycle_directions.cpp
@@ -87,7 +87,7 @@ BicycleDirectionsEngine::BicycleDirectionsEngine(Index const & index) : m_index(
 void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction> const & path,
                                        Route::TTimes & times, Route::TTurns & turns,
                                        vector<Junction> & routeGeometry,
-                                       vector<traffic::TrafficInfo::RoadSegmentId> & routeSegs,
+                                       vector<TrafficInfo::RoadSegmentId> & trafficSegs,
                                        my::Cancellable const & cancellable)
 {
   times.clear();
@@ -95,7 +95,7 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
   routeGeometry.clear();
   m_adjacentEdges.clear();
   m_pathSegments.clear();
-  routeSegs.clear();
+  trafficSegs.clear();
 
   size_t const pathSize = path.size();
   if (pathSize == 0)
@@ -132,6 +132,9 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
   m_adjacentEdges.insert(make_pair(0, AdjacentEdges(0)));
   for (size_t i = 1; i < pathSize; ++i)
   {
+    if (cancellable.IsCancelled())
+      return;
+
     Junction const & prevJunction = path[i - 1];
     Junction const & currJunction = path[i];
     IRoadGraph::TEdgeVector outgoingEdges, ingoingEdges;
@@ -169,7 +172,7 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
     // prevJunction == path[i - 1] and currJunction == path[i].
     if (inFeatureId.IsValid())
       LoadPathGeometry(inFeatureId, {prevJunction, currJunction}, pathSegment);
-    pathSegment.m_routeSegs = {
+    pathSegment.m_trafficSegs = {
       {inFeatureId.m_index,
        static_cast<uint16_t>(routeEdges[i - 1].GetSegId()),
        routeEdges[i - 1].IsForward() ? TrafficInfo::RoadSegmentId::kForwardDirection
@@ -186,12 +189,17 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
   Route::TStreets streetNames;
 
   MakeTurnAnnotation(resultGraph, delegate, routeGeometry, turns, turnAnnotationTimes,
-                     streetNames, routeSegs);
+                     streetNames, trafficSegs);
   // @TODO(bykoianko) The invariant below it's an issue but now it's so and it should be checked.
-  // The problem is every edge is added as a pair of point to route geometry.
+  // The problem is every edge is added as a pair of points to route geometry.
   // So all the points except for beginning and ending are duplicated. It should
   // be fixed in the future.
-  ASSERT_EQUAL(routeGeometry.size(), 2 * routeSegs.size(), ());
+
+  // Note 1. Accoding to current implementation all internal points are duplicated for
+  // all A* routes.
+  // Note 2. Number of |trafficSegs| should be equal to number of |routeEdges|.
+  ASSERT_EQUAL(routeGeometry.size(), 2 * (pathSize - 2) + 2, ());
+  ASSERT_EQUAL(trafficSegs.size(), pathSize - 1, ());
 }
 
 Index::FeaturesLoaderGuard & BicycleDirectionsEngine::GetLoader(MwmSet::MwmId const & id)
--- a/routing/bicycle_directions.cpp
+++ b/routing/bicycle_directions.cpp
@@ -190,6 +190,7 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
 
   MakeTurnAnnotation(resultGraph, delegate, routeGeometry, turns, turnAnnotationTimes,
                      streetNames, trafficSegs);
+
   // @TODO(bykoianko) The invariant below it's an issue but now it's so and it should be checked.
   // The problem is every edge is added as a pair of points to route geometry.
   // So all the points except for beginning and ending are duplicated. It should
--- a/routing/car_router.cpp
+++ b/routing/car_router.cpp
@@ -7,7 +7,10 @@
 #include "routing/osrm_helpers.hpp"
 #include "routing/osrm_path_segment_factory.hpp"
 #include "routing/road_graph_router.hpp"
+#include "routing/road_point.hpp"
 #include "routing/turns_generator.hpp"
+
+#include "traffic/traffic_info.hpp"
 
 #include "platform/country_file.hpp"
 #include "platform/platform.hpp"
@@ -204,6 +207,7 @@ IRouter::ResultCode FindSingleOsrmRoute(FeatureGraphNode const & source,
   Route::TTurns turns;
   Route::TTimes times;
   Route::TStreets streets;
+  vector<traffic::TrafficInfo::RoadSegmentId> routeSegs;
 
   LOG(LINFO, ("OSRM route from", MercatorBounds::ToLatLon(source.segmentPoint), "to",
               MercatorBounds::ToLatLon(target.segmentPoint)));
@@ -214,7 +218,7 @@ IRouter::ResultCode FindSingleOsrmRoute(FeatureGraphNode const & source,
 
   OSRMRoutingResult routingResult(index, *mapping, rawRoutingResult);
   routing::IRouter::ResultCode const result =
-      MakeTurnAnnotation(routingResult, delegate, geometry, turns, times, streets);
+      MakeTurnAnnotation(routingResult, delegate, geometry, turns, times, streets, routeSegs);
   if (result != routing::IRouter::NoError)
   {
     LOG(LWARNING, ("Can't load road path data from disk for", mapping->GetCountryName(),
@@ -225,6 +229,7 @@ IRouter::ResultCode FindSingleOsrmRoute(FeatureGraphNode const & source,
   route.SetTurnInstructions(move(turns));
   route.SetSectionTimes(move(times));
   route.SetStreetNames(move(streets));
+  route.SetTraffic({} /* No traffic info in case of OSRM */);
 
   vector<m2::PointD> mwmPoints;
   JunctionsToPoints(geometry, mwmPoints);
--- a/routing/car_router.cpp
+++ b/routing/car_router.cpp
@@ -207,7 +207,7 @@ IRouter::ResultCode FindSingleOsrmRoute(FeatureGraphNode const & source,
   Route::TTurns turns;
   Route::TTimes times;
   Route::TStreets streets;
-  vector<traffic::TrafficInfo::RoadSegmentId> routeSegs;
+  vector<traffic::TrafficInfo::RoadSegmentId> trafficSegs;
 
   LOG(LINFO, ("OSRM route from", MercatorBounds::ToLatLon(source.segmentPoint), "to",
               MercatorBounds::ToLatLon(target.segmentPoint)));
@@ -218,7 +218,7 @@ IRouter::ResultCode FindSingleOsrmRoute(FeatureGraphNode const & source,
 
   OSRMRoutingResult routingResult(index, *mapping, rawRoutingResult);
   routing::IRouter::ResultCode const result =
-      MakeTurnAnnotation(routingResult, delegate, geometry, turns, times, streets, routeSegs);
+      MakeTurnAnnotation(routingResult, delegate, geometry, turns, times, streets, trafficSegs);
   if (result != routing::IRouter::NoError)
   {
     LOG(LWARNING, ("Can't load road path data from disk for", mapping->GetCountryName(),
@@ -276,8 +276,14 @@ bool CarRouter::FindRouteMSMT(TFeatureGraphNodeVec const & sources,
   {
     for (auto const & sourceEdge : sources)
     {
-      if (FindSingleRouteDispatcher(sourceEdge, targetEdge, delegate, mapping, route) == NoError)
-        return true;
+      IRouter::ResultCode const code =
+          FindSingleRouteDispatcher(sourceEdge, targetEdge, delegate, mapping, route);
+      switch (code)
+      {
+      case NoError: return true;
+      case Cancelled: return false;
+      default: continue;
+      }
     }
   }
   return false;
--- a/routing/directions_engine.hpp
+++ b/routing/directions_engine.hpp
@@ -1,7 +1,10 @@
 #pragma once
 
 #include "routing/road_graph.hpp"
+#include "routing/road_point.hpp"
 #include "routing/route.hpp"
+
+#include "traffic/traffic_info.hpp"
 
 #include "base/cancellable.hpp"
 
@@ -16,7 +19,9 @@ public:
 
   virtual void Generate(IRoadGraph const & graph, vector<Junction> const & path,
                         Route::TTimes & times, Route::TTurns & turns,
-                        vector<Junction> & routeGeometry, my::Cancellable const & cancellable) = 0;
+                        vector<Junction> & routeGeometry,
+                        vector<traffic::TrafficInfo::RoadSegmentId> & routeSegs,
+                        my::Cancellable const & cancellable) = 0;
 
 protected:
   /// \brief constructs route based on |graph| and |path|. Fills |routeEdges| with the route.
--- a/routing/directions_engine.hpp
+++ b/routing/directions_engine.hpp
@@ -20,7 +20,7 @@ public:
   virtual void Generate(IRoadGraph const & graph, vector<Junction> const & path,
                         Route::TTimes & times, Route::TTurns & turns,
                         vector<Junction> & routeGeometry,
-                        vector<traffic::TrafficInfo::RoadSegmentId> & routeSegs,
+                        vector<traffic::TrafficInfo::RoadSegmentId> & trafficSegs,
                         my::Cancellable const & cancellable) = 0;
 
 protected:
--- a/routing/pedestrian_directions.cpp
+++ b/routing/pedestrian_directions.cpp
@@ -35,6 +35,7 @@ PedestrianDirectionsEngine::PedestrianDirectionsEngine()
 void PedestrianDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction> const & path,
                                           Route::TTimes & times, Route::TTurns & turns,
                                           vector<Junction> & routeGeometry,
+                                          vector<traffic::TrafficInfo::RoadSegmentId> &,
                                           my::Cancellable const & cancellable)
 {
   times.clear();
--- a/routing/pedestrian_directions.cpp
+++ b/routing/pedestrian_directions.cpp
@@ -35,7 +35,7 @@ PedestrianDirectionsEngine::PedestrianDirectionsEngine()
 void PedestrianDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction> const & path,
                                           Route::TTimes & times, Route::TTurns & turns,
                                           vector<Junction> & routeGeometry,
-                                          vector<traffic::TrafficInfo::RoadSegmentId> &,
+                                          vector<traffic::TrafficInfo::RoadSegmentId> & /* trafficSegs */,
                                           my::Cancellable const & cancellable)
 {
   times.clear();
--- a/routing/pedestrian_directions.hpp
+++ b/routing/pedestrian_directions.hpp
@@ -13,6 +13,7 @@ public:
   // IDirectionsEngine override:
   void Generate(IRoadGraph const & graph, vector<Junction> const & path, Route::TTimes & times,
                 Route::TTurns & turns, vector<Junction> & routeGeometry,
+                vector<traffic::TrafficInfo::RoadSegmentId> &,
                 my::Cancellable const & cancellable) override;
 
 private:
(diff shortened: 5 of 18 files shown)


routing/bicycle_directions.cpp
```diff
@@ -170,7 +185,13 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
   Route::TTimes turnAnnotationTimes;
   Route::TStreets streetNames;
 
-  MakeTurnAnnotation(resultGraph, delegate, routeGeometry, turns, turnAnnotationTimes, streetNames);
+  MakeTurnAnnotation(resultGraph, delegate, routeGeometry, turns, turnAnnotationTimes,
+                     streetNames, routeSegs);
+  // @TODO(bykoianko) The invariant below it's an issue but now it's so and it should be checked.
+  // The problem is every edge is added as a pair of point to route geometry.
```
ygorshenin: paor of points

routing/bicycle_directions.cpp
```diff
@@ -170,7 +185,13 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
...
+  // So all the points except for beginning and ending are duplicated. It should
```
ygorshenin: Is it a real problem?

routing/bicycle_directions.cpp
```diff
@@ -170,7 +185,13 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
...
+  // be fixed in the future.
+  ASSERT_EQUAL(routeGeometry.size(), 2 * routeSegs.size(), ());
```
ygorshenin: Are you sure? Number of points should be equal to number of segments + 1, isn't it?

routing/pedestrian_directions.cpp
```diff
@@ -35,6 +35,7 @@ PedestrianDirectionsEngine::PedestrianDirectionsEngine()
 void PedestrianDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction> const & path,
                                           Route::TTimes & times, Route::TTurns & turns,
                                           vector<Junction> & routeGeometry,
+                                          vector<traffic::TrafficInfo::RoadSegmentId> &,
```
ygorshenin: If you don't need this parameter, at least, put it into ```/* */```.

routing/pedestrian_directions.hpp
```diff
@@ -13,6 +13,7 @@ class PedestrianDirectionsEngine : public IDirectionsEngine
   // IDirectionsEngine override:
   void Generate(IRoadGraph const & graph, vector<Junction> const & path, Route::TTimes & times,
                 Route::TTurns & turns, vector<Junction> & routeGeometry,
+                vector<traffic::TrafficInfo::RoadSegmentId> &,
```
ygorshenin: If you don't need this parameter, at least, put it into ```/* */```.

routing/route.cpp
```diff
@@ -15,6 +17,8 @@
 
 namespace routing
 {
+using namespace traffic;
```
ygorshenin: Please, move this out of namespace routing.

routing/route.cpp
```diff
@@ -354,6 +359,42 @@ void Route::Update()
...
+  {
+    m_traffic.assign(route.GetTraffic().cbegin(), route.GetTraffic().cend());
+    return;
+  }
+
+  // Note. At this point the last item of |m_poly| should be removed.
+  // So the size of |m_traffic| should be equal to size of |m_poly.
+  if (GetTraffic().empty())
+  {
+    CHECK_GREATER_OR_EQUAL(m_poly.GetPolyline().GetSize(), 1, ());
```
ygorshenin: Why? If start and end points of route coincide, the size of poyline is one. From my point of view, it's not a problem.

routing/route.cpp
```diff
@@ -354,6 +359,42 @@ void Route::Update()
...
+    m_traffic.resize(m_poly.GetPolyline().GetSize(), SpeedGroup::Unknown);
```
ygorshenin: Why? Traffic is on segments, not on points, therefore the size of traffic should be ```m_poly.GetPolyline().GetSize() - 1```, isn't it?

routing/route.cpp
```diff
@@ -354,6 +359,42 @@ void Route::Update()
...
+  }
+
+  CHECK_EQUAL(GetTraffic().size(), m_poly.GetPolyline().GetSize(), ());
+
+  if (route.GetTraffic().empty())
+  {
+    CHECK_GREATER_OR_EQUAL(route.m_poly.GetPolyline().GetSize(), 1, ());
+    m_traffic.insert(m_traffic.end(),
+                     route.m_poly.GetPolyline().GetSize() - 1 /* segment number is less by one */,
```
ygorshenin: Not a "segment number". "Segment number" is a number of a particular segment. Please, use "number of segments".

routing/route.cpp
ygorshenin: The method is called "Append", but here I see that you actually assign something to traffic. Is it OK that the m_traffic may not be empty before this line of code?

routing/routing_helpers.cpp
```diff
@@ -35,5 +42,34 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
   route.SetTurnInstructions(move(turnsDir));
   route.SetStreetNames(move(streetNames));
   route.SetAltitudes(move(altitudes));
+
+  vector<traffic::SpeedGroup> traffic;
+  traffic.reserve(routeSegs.size());
```
ygorshenin: So, what's the point to reserve space but possibly don't use it?

routing/routing_helpers.cpp
```diff
@@ -35,5 +42,34 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
...
+      auto const it = trafficColoring->find(rp);
+      SpeedGroup segTraffic =  (it == trafficColoring->cend()) ? SpeedGroup::Unknown
+                                                               : it->second;
+      // @TODO It's written to compensate an error. The problem is in case of any routing except for osrm
+      // all route points except for begining and ending are duplicated.
+      // See a TODO in BicycleDirectionsEngine::Generate() for details.
+      // At this moment the route looks like:
+      // 0----1    4----5
+      //      2----3    6---7
+      // This route composes of 4 edges and there're 4 items in routeSegs.
```
ygorshenin: consists of 4 edges

routing/routing_helpers.cpp
```diff
@@ -35,5 +42,34 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
...
+      // But it has 8 items in routeGeometry.
+      // So for segments 0-1 and 1-2 let's set routeSegs[0]
+      // for segments 2-3 and 3-4 let's set routeSegs[1]
+      // for segments 4-5 and 5-7 let's set routeSegs[2]
+      // and for segment 6-7 let's set routeSegs[3]
+      traffic.insert(traffic.end(), {segTraffic, segTraffic});
```
ygorshenin: push_back/emplace_back are worse than insert at the end?

routing/turns_generator.cpp
```diff
@@ -271,6 +273,7 @@ IRouter::ResultCode MakeTurnAnnotation(turns::IRoutingResult const & result,
   // Annotate turns.
   size_t skipTurnSegments = 0;
   auto const & loadedSegments = result.GetSegments();
+  routeSegs.reserve(loadedSegments.size());
```
ygorshenin: Is it guaranteed that in average the size of loadedSegmentIt->m_routeSegs is 1?

routing/bicycle_directions.cpp
bykoianko: pair of points. fixed.

routing/bicycle_directions.cpp
bykoianko: It depends on what do you mean by saying an edge.

At this moment the route for all routing except for osrm looks like:
```
 0----1    4----5
      2----3    6---7
```
This route composes of 4 connected routing::Edge(s). That means that point 1 and 2, 3 and 4 or 5 and 6 have the same coords. So I put 4 items in routeSegs. More then that the route comes from the router as 5 point route. But then in Bicycle(Pedestrian)DirectionsEngine::Generate() the path converted to 8-point route with 3 zero length segments.

I guess the route in this example should contain only 5 points and it's an issue that after some manipulations it turns to 8 point route. I consider this turning as an issue.

I try to git rid of this effect in the PR but face some further issues. I guess it should be refactored when we try to get rid of ReconstractRoute/ReconstructPath/RedressRoute and so on. This problems are close to each other.

routing/bicycle_directions.cpp
bykoianko: It has some disadvantages:
- Difficulty in understanding: (а) if after route building a route has n points after route reconstruction (including turn generation) it has (n-1) * 2 point; (b) OSRM route don't have such duplication effect;
- Efficiency. The route contains more points then necessary;

So in my opinion it's better to fix.

routing/pedestrian_directions.cpp
bykoianko: Done in two places.

routing/pedestrian_directions.hpp
bykoianko: Done

routing/route.cpp
bykoianko: Done

routing/route.cpp
```diff
@@ -354,6 +359,42 @@ void Route::Update()
...
+    return;
+
+  if (!IsValid())
+  {
+    m_traffic.assign(route.GetTraffic().cbegin(), route.GetTraffic().cend());
+    return;
+  }
+
+  // Note. At this point the last item of |m_poly| should be removed.
+  // So the size of |m_traffic| should be equal to size of |m_poly.
```
bykoianko: Done

routing/route.cpp
bykoianko: Such cases are processed above in this method. 

```c++
void Route::AppendTraffic(Route const & route)
{
  CHECK(route.IsValid(), ());

  if (GetTraffic().empty() && route.GetTraffic().empty())
    return;

  if (!IsValid())
  {
    m_traffic.assign(route.GetTraffic().cbegin(), route.GetTraffic().cend());
    return;
  }

  // Note. At this point the last item of |m_poly| should be removed.
  // So the size of |m_traffic| should be equal to size of |m_poly|.
  if (GetTraffic().empty())
  {
    CHECK_GREATER_OR_EQUAL(m_poly.GetPolyline().GetSize(), 1, ());
    m_traffic.resize(m_poly.GetPolyline().GetSize(), SpeedGroup::Unknown);
```

If IsValid() returns false there's a special branch for it. And ```bool IsValid() const { return (m_poly.GetPolyline().GetSize() > 1); }```.

routing/route.cpp
bykoianko: > Why? Traffic is on segments, not on points, therefore the size of traffic should be m_poly.GetPolyline().GetSize() - 1, isn't it?

Please see my note above in the code AppendTraffic(). The thing is AppendTraffic() is called form Route::AppendRoute() and at this moment last point of m_poly.GetPolyline() has been removed in Route::AppendRoute().

> The method is called "Append", but here I see that you actually assign something to traffic.

Please see code below, after the line:
```c++
...
  if (route.GetTraffic().empty())
  {
    CHECK_GREATER_OR_EQUAL(route.m_poly.GetPolyline().GetSize(), 1, ());
    m_traffic.insert(m_traffic.end(),
                     route.m_poly.GetPolyline().GetSize() - 1 /* segment number is less by one */,
                     SpeedGroup::Unknown);
  }
  else
  {
    m_traffic.insert(m_traffic.end(), route.GetTraffic().cbegin(), route.GetTraffic().cend());
  }
```

> Is it OK that the m_traffic may not be empty before this line of code?

Ok, m_traffic could be empty (if there's no traffic information for this kind of route) or the size of |m_traffic| should be one less then |m_poly.GetPolyline().GetSize()|. There's a tricky when we need to append the route with empty |m_traffic| to the route with full |m_traffic|. It could happens in case of cross mwm car routing form A* route to osrm route. In that case the resulting route should have size |m_traffic| one less then |m_poly.GetPolyline().GetSize()|. In that case for segments which don't have any traffic information we set SpeedGroup::Unknown.

routing/route.cpp
bykoianko: Right!

routing/routing_helpers.cpp
bykoianko: Yes. I'm moving the line below, to reserve only if we need it.

routing/routing_helpers.cpp
bykoianko: Yes

routing/routing_helpers.cpp
bykoianko: Here I'd rather use insert because to has interval/initializer_list signature. Do you think using push_back twice is better here?

bykoianko:
DUDE JTTP

bykoianko:
@rokuz @darina @ygorshenin @mpimenov PTAL

rokuz:
LGTM

routing/bicycle_directions.cpp
```diff
@@ -170,7 +185,13 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
...
   Route::TStreets streetNames;
 
-  MakeTurnAnnotation(resultGraph, delegate, routeGeometry, turns, turnAnnotationTimes, streetNames);
+  MakeTurnAnnotation(resultGraph, delegate, routeGeometry, turns, turnAnnotationTimes,
+                     streetNames, routeSegs);
+  // @TODO(bykoianko) The invariant below it's an issue but now it's so and it should be checked.
+  // The problem is every edge is added as a pair of points to route geometry.
+  // So all the points except for beginning and ending are duplicated. It should
+  // be fixed in the future.
+  ASSERT_EQUAL(routeGeometry.size(), 2 * routeSegs.size(), ());
```
ygorshenin: Please, replace this ASSERTion by two:
```c++
ASSERT_EQUAL(routeGeometry.size(), 2 * (pathSize - 2) + 2, ());
ASSERT_EQUAL(routeSegs.size(), pathSize - 1, ());
```
with the comment explaining that all internal points of path are duplicated in routeGeometry, and that number of routeSegs should be equal to number of edges in the path.

routing/car_router.cpp
```diff
@@ -225,6 +229,7 @@ IRouter::ResultCode FindSingleOsrmRoute(FeatureGraphNode const & source,
   route.SetTurnInstructions(move(turns));
   route.SetSectionTimes(move(times));
   route.SetStreetNames(move(streets));
+  route.SetTraffic({} /* No traffic info in case of OSRM */);
```
ygorshenin: Is it ok that in route the invariant you described above is violated? Why not to create a traffic list of Unknowns of size geometry.size() / 2?

routing/directions_engine.hpp
```diff
@@ -16,7 +19,9 @@ class IDirectionsEngine
 
   virtual void Generate(IRoadGraph const & graph, vector<Junction> const & path,
                         Route::TTimes & times, Route::TTurns & turns,
-                        vector<Junction> & routeGeometry, my::Cancellable const & cancellable) = 0;
+                        vector<Junction> & routeGeometry,
+                        vector<traffic::TrafficInfo::RoadSegmentId> & routeSegs,
```
ygorshenin: s/routeSegs/trafficSegs/g ?

routing/route.cpp
```diff
@@ -354,6 +359,42 @@ void Route::Update()
...
+  {
+    CHECK_GREATER_OR_EQUAL(route.m_poly.GetPolyline().GetSize(), 1, ());
+    m_traffic.insert(m_traffic.end(),
+                     route.m_poly.GetPolyline().GetSize() - 1 /* number of segments is less by one */,
+                     SpeedGroup::Unknown);
+  }
+  else
+  {
+    m_traffic.insert(m_traffic.end(), route.GetTraffic().cbegin(), route.GetTraffic().cend());
+  }
```
ygorshenin: Is it possible to put assertions here?

ygorshenin: About invariant you described above?

routing/routing_helpers.cpp
ygorshenin: Leave as is.

routing/routing_helpers.cpp
```diff
@@ -35,5 +42,34 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
   route.SetTurnInstructions(move(turnsDir));
   route.SetStreetNames(move(streetNames));
   route.SetAltitudes(move(altitudes));
+
+  vector<traffic::SpeedGroup> traffic;
+  if (trafficColoring)
+  {
+    traffic.reserve(routeSegs.size());
```
ygorshenin: 2 * routeSegs.size() ?

routing/routing_helpers.cpp
```diff
@@ -35,5 +42,34 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
...
+      //      2----3    6---7
+      // This route consists of 4 edges and there're 4 items in routeSegs.
+      // But it has 8 items in routeGeometry.
+      // So for segments 0-1 and 1-2 let's set routeSegs[0]
+      // for segments 2-3 and 3-4 let's set routeSegs[1]
+      // for segments 4-5 and 5-7 let's set routeSegs[2]
+      // and for segment 6-7 let's set routeSegs[3]
+      traffic.insert(traffic.end(), {segTraffic, segTraffic});
+    }
+    if (!traffic.empty())
```
ygorshenin: Why?

darina:
LGTM

routing/bicycle_directions.cpp
bykoianko: Done

routing/car_router.cpp
bykoianko: The route invariant is not violated here. Route may have either empty |m_traffic| or the size of |m_traffic| should be m_poly.GetPolyline().GetSize()-1. This invariant is checked in Route::AppendRoute(Route const & route)/Route::AppendTraffic(Route const & route).

Why do we need to support an empty |m_traffic| vector? Some routes (osrm/pedestrian..) don't have any traffic info at all.

routing/route.cpp
bykoianko: Yes. It's possible. More then this invariant is checked after call AppendTraffic() in AppendRoute()
```
  AppendTraffic(route);

  m_poly.Append(route.m_poly);
  if (!GetTraffic().empty())
    CHECK_EQUAL(GetTraffic().size() + 1, m_poly.GetPolyline().GetSize(), ());
```
To be on the safe side let's CHECK it here at the end of AppendTraffic one more time. I has to leave the CHECK in AppendRoute because we may leave AppendTraffic earlier (in one of early returns.)

routing/routing_helpers.cpp
bykoianko: BicycleDirectionsEngine::Generate() may fill with empty vectors:
|times|, |turns|, |routeGeometry| and |trafficSegs|. This params form an invalid route. 

It's wrong working with routes which have beginning and ending on the same segment. It's better to fix but I guess it's a question of separate PR.

Now I:
- write todo about the issue;
- check (with if) if routeSegs is empty instead of checking |traffic|;
- put CHECK that |traffic| is not empty:
```
  if (engine)
    engine->Generate(graph, path, times, turnsDir, junctions, routeSegs, cancellable);

  // @TODO(bykoianko) If the start and the finish of a route lies on the same road segemnt
  // engine->Generate() fills with empty vectors |times|, |turnsDir|, |junctions| and |routeSegs|.
  // It's not correct and should be fixed. It's necessary to work corrrectly with such routes.

  vector<m2::PointD> routeGeometry;
  JunctionsToPoints(junctions, routeGeometry);
  feature::TAltitudes altitudes;
  JunctionsToAltitudes(junctions, altitudes);

  route.SetGeometry(routeGeometry.begin(), routeGeometry.end());
  route.SetSectionTimes(move(times));
  route.SetTurnInstructions(move(turnsDir));
  route.SetStreetNames(move(streetNames));
  route.SetAltitudes(move(altitudes));

  vector<traffic::SpeedGroup> traffic;
  if (trafficColoring && !routeSegs.empty())
  {
    traffic.reserve(2 * routeSegs.size());
    for (TrafficInfo::RoadSegmentId const & rp : routeSegs)
    {
      auto const it = trafficColoring->find(rp);
      SpeedGroup segTraffic =  (it == trafficColoring->cend()) ? SpeedGroup::Unknown
                                                               : it->second;
      // @TODO It's written to compensate an error. The problem is in case of any routing except for osrm
      // all route points except for begining and ending are duplicated.
      // See a TODO in BicycleDirectionsEngine::Generate() for details.
      // At this moment the route looks like:
      // 0----1    4----5
      //      2----3    6---7
      // This route consists of 4 edges and there're 4 items in routeSegs.
      // But it has 8 items in routeGeometry.
      // So for segments 0-1 and 1-2 let's set routeSegs[0]
      // for segments 2-3 and 3-4 let's set routeSegs[1]
      // for segments 4-5 and 5-7 let's set routeSegs[2]
      // and for segment 6-7 let's set routeSegs[3]
      traffic.insert(traffic.end(), {segTraffic, segTraffic});
    }
    CHECK(!traffic.empty(), ());
    traffic.pop_back();
  }

  route.SetTraffic(move(traffic));
}
```

bykoianko:
@ygorshenin PTAL

routing/pedestrian_directions.cpp
```diff
@@ -35,6 +35,7 @@ PedestrianDirectionsEngine::PedestrianDirectionsEngine()
 void PedestrianDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction> const & path,
                                           Route::TTimes & times, Route::TTurns & turns,
                                           vector<Junction> & routeGeometry,
+                                          vector<traffic::TrafficInfo::RoadSegmentId> & /* routeSegs */,
```
ygorshenin: s/routeSegs/trafficSegs/g

routing/pedestrian_directions.hpp
```diff
@@ -13,6 +13,7 @@ class PedestrianDirectionsEngine : public IDirectionsEngine
   // IDirectionsEngine override:
   void Generate(IRoadGraph const & graph, vector<Junction> const & path, Route::TTimes & times,
                 Route::TTurns & turns, vector<Junction> & routeGeometry,
+                vector<traffic::TrafficInfo::RoadSegmentId> & /* routeSegs */,
```
ygorshenin: s/routeSegs/trafficSegs/g

routing/route.cpp
```diff
@@ -408,13 +449,16 @@ void Route::AppendRoute(Route const & route)
     m_times.push_back(move(t));
   }
 
+  AppendTraffic(route);
+
   m_poly.Append(route.m_poly);
+  if (!GetTraffic().empty())
```
ygorshenin: Okay, leave as is.

routing/turns_generator.hpp
```diff
@@ -9,7 +9,10 @@
 #include "routing/turns.hpp"
 #include "routing/turn_candidate.hpp"
 
+#include "traffic/traffic_info.hpp"
+
 #include "std/function.hpp"
+#include "std/shared_ptr.hpp"
```
ygorshenin: Why do we need shared pointers here?

routing/bicycle_directions.cpp
```diff
@@ -83,13 +87,15 @@ BicycleDirectionsEngine::BicycleDirectionsEngine(Index const & index) : m_index(
 void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction> const & path,
                                        Route::TTimes & times, Route::TTurns & turns,
                                        vector<Junction> & routeGeometry,
+                                       vector<traffic::TrafficInfo::RoadSegmentId> & trafficSegs,
```
mpimenov: No need in ```traffic::```

routing/pedestrian_directions.cpp
bykoianko: Yes. If to look for routeSegs there're a lot which should be called trafficSegs . I'll rename them.

routing/bicycle_directions.cpp
```diff
@@ -170,7 +185,18 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
...
+                     streetNames, trafficSegs);
+  // @TODO(bykoianko) The invariant below it's an issue but now it's so and it should be checked.
+  // The problem is every edge is added as a pair of points to route geometry.
+  // So all the points except for beginning and ending are duplicated. It should
+  // be fixed in the future.
+
+  // Note 1. Accoding to current implementation all internal points are duplicated for
+  // all A* routes.
+  // Note 2. Number of |trafficSegs| should be equal to number of |routeEdges|.
+  ASSERT_EQUAL(routeGeometry.size(), 2 * (pathSize - 2) + 2, ());
```
mpimenov: ```2 * (pathSize - 1)``` ?

routing/pedestrian_directions.hpp
bykoianko: Done

routing/bicycle_directions.cpp
```diff
@@ -158,8 +164,17 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
...
+    // for every route edge leads that all route points are duplicated. It's because
+    // prevJunction == path[i - 1] and currJunction == path[i].
     if (inFeatureId.IsValid())
       LoadPathGeometry(inFeatureId, {prevJunction, currJunction}, pathSegment);
+    pathSegment.m_trafficSegs = {
+      {inFeatureId.m_index,
+       static_cast<uint16_t>(routeEdges[i - 1].GetSegId()),
+       routeEdges[i - 1].IsForward() ? TrafficInfo::RoadSegmentId::kForwardDirection
+                                     : TrafficInfo::RoadSegmentId::kReverseDirection
+      }};
```
kshalnev: May be check |cancellable| in these loops? At least in the outer.

routing/routing_helpers.cpp
```diff
@@ -35,5 +46,34 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
   route.SetTurnInstructions(move(turnsDir));
   route.SetStreetNames(move(streetNames));
   route.SetAltitudes(move(altitudes));
+
+  vector<traffic::SpeedGroup> traffic;
+  if (trafficColoring && !routeSegs.empty())
+  {
+    traffic.reserve(2 * routeSegs.size());
+    for (TrafficInfo::RoadSegmentId const & rp : routeSegs)
```
mpimenov: s/rp/seg/ ?

routing/bicycle_directions.cpp
ygorshenin: I guess ```2 * (n - 2) + 2``` is more readable, as it means that ```each non-end point is duplicated, end points are left as is```, what is the same as what is specified in the comment above.

routing/bicycle_directions.cpp
kshalnev: "traffic::" can be removed

routing/route.cpp
```diff
@@ -354,6 +359,43 @@ void Route::Update()
...
+void Route::AppendTraffic(Route const & route)
+{
+  CHECK(route.IsValid(), ());
+
+  if (GetTraffic().empty() && route.GetTraffic().empty())
+    return;
+
+  if (!IsValid())
+  {
+    m_traffic.assign(route.GetTraffic().cbegin(), route.GetTraffic().cend());
```
kshalnev: you can rewrite the line a bit shorter and clearer:
m_traffic = route.GetTraffic();

routing/bicycle_directions.cpp
bykoianko: Removed

routing/routing_helpers.cpp
bykoianko: Yes. Thanks.

routing/bicycle_directions.cpp
bykoianko: Removed

routing/route.cpp
bykoianko: Yes. Right!

routing/bicycle_directions.cpp
bykoianko: I added code which I wrote above.

routing/routing_helpers.hpp
```diff
@@ -23,5 +26,6 @@ bool IsRoad(TTypes const & types)
 }
 
 void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
+                      shared_ptr<traffic::TrafficInfo::Coloring> trafficColoring,
```
kshalnev: may be const &? here is only using, not owning.

bykoianko:
@ygorshenin @mpimenov @kshalnev PTAL

bykoianko:
DUDE JTTP

ygorshenin:
LGTM

mpimenov:
LGTM

routing/bicycle_directions.cpp
```diff
@@ -170,7 +188,18 @@ void BicycleDirectionsEngine::Generate(IRoadGraph const & graph, vector<Junction
   Route::TTimes turnAnnotationTimes;
   Route::TStreets streetNames;
 
-  MakeTurnAnnotation(resultGraph, delegate, routeGeometry, turns, turnAnnotationTimes, streetNames);
+  MakeTurnAnnotation(resultGraph, delegate, routeGeometry, turns, turnAnnotationTimes,
+                     streetNames, trafficSegs);
+  // @TODO(bykoianko) The invariant below it's an issue but now it's so and it should be checked.
```
bykoianko: Done

ygorshenin:
LGTM

bykoianko:
Я дописываю комментарий еще один. Пока прошу не мержить.

bykoianko:
@kshalnev PTAL

routing/routing_helpers.cpp
```diff
@@ -35,5 +46,34 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
...
+      // This route consists of 4 edges and there're 4 items in trafficSegs.
+      // But it has 8 items in routeGeometry.
+      // So for segments 0-1 and 1-2 let's set trafficSegs[0]
+      // for segments 2-3 and 3-4 let's set trafficSegs[1]
+      // for segments 4-5 and 5-7 let's set trafficSegs[2]
+      // and for segment 6-7 let's set trafficSegs[3]
+      traffic.insert(traffic.end(), {segTraffic, segTraffic});
+    }
+    CHECK(!traffic.empty(), ());
+    traffic.pop_back();
```
kshalnev: let add comment why pop_back

routing/routing_helpers.cpp
```diff
@@ -35,5 +46,34 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
...
   route.SetAltitudes(move(altitudes));
+
+  vector<traffic::SpeedGroup> traffic;
+  if (trafficColoring && !trafficSegs.empty())
+  {
+    traffic.reserve(2 * trafficSegs.size());
+    for (TrafficInfo::RoadSegmentId const & seg : trafficSegs)
+    {
+      auto const it = trafficColoring->find(seg);
+      SpeedGroup segTraffic =  (it == trafficColoring->cend()) ? SpeedGroup::Unknown
```
kshalnev: const

routing/routing_helpers.cpp
```diff
@@ -35,5 +46,34 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
...
+      // all route points except for begining and ending are duplicated.
+      // See a TODO in BicycleDirectionsEngine::Generate() for details.
+      // At this moment the route looks like:
+      // 0----1    4----5
+      //      2----3    6---7
+      // This route consists of 4 edges and there're 4 items in trafficSegs.
+      // But it has 8 items in routeGeometry.
+      // So for segments 0-1 and 1-2 let's set trafficSegs[0]
+      // for segments 2-3 and 3-4 let's set trafficSegs[1]
+      // for segments 4-5 and 5-7 let's set trafficSegs[2]
```
kshalnev: 'and 5-6'?

routing/routing_helpers.cpp
bykoianko: Please see the comment above about it.

routing/routing_helpers.cpp
```diff
@@ -35,5 +46,34 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
...
+      // and for segment 6-7 let's set trafficSegs[3]
+      traffic.insert(traffic.end(), {segTraffic, segTraffic});
```
bykoianko: emplace_back does not have interval/initializer_list version which is used here.

routing/routing_helpers.cpp
bykoianko: yes

routing/routing_helpers.cpp
bykoianko: Yes

kshalnev:
LGTM

bykoianko:
Всем спасибо за ревью!

PR #4998, opened 2016-12-15
[RELEASE ONLY] Applying restrictions line-point-line at runtime on index graph.

bykoianko:
cherry-pick PR на мастре https://github.com/mapsme/omim/pull/4817, который пока не ревью.


bykoianko:
DUDE JTTP

PR #5009, opened 2016-12-15
[Release crash fix] Correct processing fake feature of IndexGraphStarter.

bykoianko:
@ygorshenin @dobriy-eeh PTAL

### DIFF
--- a/routing/index_graph_starter.cpp
+++ b/routing/index_graph_starter.cpp
@@ -42,6 +42,7 @@ m2::PointD const & IndexGraphStarter::GetPoint(Joint::Id jointId)
 
 DirectedEdge const & IndexGraphStarter::FindFakeEdge(uint32_t fakeFeatureId)
 {
+  ASSERT(IsFakeFeature(fakeFeatureId), ("Feature", fakeFeatureId, "is not a fake one"));
   for (DirectedEdge const & e : m_fakeZeroLengthEdges)
   {
     ASSERT_EQUAL(GetPoint(e.GetFrom()), GetPoint(e.GetTo()), ());
--- a/routing/index_graph_starter.cpp
+++ b/routing/index_graph_starter.cpp
@@ -40,20 +40,34 @@ m2::PointD const & IndexGraphStarter::GetPoint(Joint::Id jointId)
   return m_graph.GetPoint(jointId);
 }
 
+DirectedEdge const & IndexGraphStarter::FindFakeEdge(uint32_t fakeFeatureId)
+{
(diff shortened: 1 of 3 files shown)


routing/index_graph_starter.cpp
```diff
@@ -40,20 +40,34 @@ m2::PointD const & IndexGraphStarter::GetPoint(Joint::Id jointId)
...
-    if (e.GetFeatureId() == rp.GetFeatureId())
-      return GetPoint(e.GetFrom());
+    if (e.GetFeatureId() == fakeFeatureId)
+      return e;
   }
-  CHECK(false,
-        (rp, "is a point of a fake feature but it's not contained in |m_fakeZeroLengthEdges|"));
-  return m_graph.GetPoint(rp);
+  CHECK(false, ("Fake feature:", fakeFeatureId, "has to be contained in |m_fakeZeroLengthEdges|"));
+  return m_fakeZeroLengthEdges.front();
```
ygorshenin: Better to return default-constructed DirectedEdge, if it's possible.

bykoianko: Yes. But in this case there's a warning about returning local value by const ref if to do so.

bykoianko: No problem in returning DirectedEdge by value except for little number of copying a small object (DirectedEdge) in seldom cases. On the other hand it's little cost we have to pay for returning default-constructed DirectedEdge. As for me the both variants are ok. Do you prefer returning a default - constructed DirectedEdge by value?

ygorshenin: It's up to you.

LGTM in any case.

routing/index_graph_starter.cpp
```diff
@@ -40,20 +40,34 @@ m2::PointD const & IndexGraphStarter::GetPoint(Joint::Id jointId)
...
+{
+  if (!IsFakeFeature(rp.GetFeatureId()))
+    return m_graph.GetPoint(rp);
+
+  // Fake edges have zero length so coords of "point from" and "point to" are equal.
+  return GetPoint(FindFakeEdge(rp.GetFeatureId()).GetFrom());
+}
+
+RoadGeometry IndexGraphStarter::GetFakeRoadGeometry(uint32_t fakeFeatureId)
+{
```
dobriy-eeh: ASSERT( IsFakeFeature(fakeFeatureId), ("Feature" + fakeFeatureId "is not fake" ));
?

bykoianko: FindFakeEdge вызывается от сюда.

dobriy-eeh:
LGTM

PR #5011, opened 2016-12-15
Preventing cyclical calling of building route in case of some errors.

bykoianko:
@dobriy-eeh @ygorshenin PTAL

### DIFF
--- a/routing/car_router.cpp
+++ b/routing/car_router.cpp
@@ -282,7 +282,18 @@ bool CarRouter::FindRouteMSMT(TFeatureGraphNodeVec const & sources,
       {
       case NoError: return true;
       case Cancelled: return false;
-      default: continue;
+      case NoCurrentPosition:
+        LOG(LERROR, ("NoCurrentPosition routing result returned by FindSingleRouteDispatcher()"));
+        return false;
+      case InconsistentMWMandRoute: return false;
+      case RouteFileNotExist: return false;
+      case StartPointNotFound: continue;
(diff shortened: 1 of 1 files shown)


dobriy-eeh:
LGTM

routing/car_router.cpp
```diff
@@ -282,7 +282,16 @@ bool CarRouter::FindRouteMSMT(TFeatureGraphNodeVec const & sources,
       {
       case NoError: return true;
       case Cancelled: return false;
-      default: continue;
+      case NoCurrentPosition: return false;
```
ygorshenin: But sources are specified as a parameter, right?

bykoianko: There's only one place where something is called:
```c++
void Framework::BuildRoute(m2::PointD const & finish, uint32_t timeoutSec)
{
  ASSERT_THREAD_CHECKER(m_threadChecker, ("BuildRoute"));
  ASSERT(m_drapeEngine != nullptr, ());

  m2::PointD start;
  if (!m_drapeEngine->GetMyPosition(start))
  {
    CallRouteBuilded(IRouter::NoCurrentPosition, storage::TCountriesVec());
    return;
  }
  BuildRoute(start, finish, false /* isP2P */, timeoutSec);
}
```
So, I guess it's possible to put an ASSERT here.

bykoianko:
@ygorshenin PTAL

ygorshenin:
LGTM

PR #5053, opened 2016-12-19
Fixes crash on startup of cmake-built ios application.

therearesomewhocallmetim:


therearesomewhocallmetim:
@VladiMihaylenko PTAL

PR #5059, opened 2016-12-19
Correction of subtracted distance for weight calculation in osrm_helpers.

fijemax:
The subtracted distance is wrong.
If segment is ordered it's transparent because `splittedSegment.m_pointEnd` and `seg.m_pointEnd` are same but if segment is reverse the wrong part of last segment is subtracted.


mapsmetest:

bykoianko:
@fijemax Thanks for the PR, but we're moving from osrm to our implementation of A* routing algorithm by some reasons. So, there's no routing/osrm_helpers.cpp file right now. I've removed from the project.

PR #5068, opened 2016-12-20
[routing] Crashfix. Processing errors while turn generation.

bykoianko:
Обработка ошибок возникающих в IDirectionsEngine::Generate().

https://fabric.io/mapsme/android/apps/com.mapswithme.maps.pro/issues/5858f58b0aeb16625b352b6d

@dobriy-eeh @ygorshenin @mpimenov PTAL

### DIFF
--- a/routing/routing_helpers.cpp
+++ b/routing/routing_helpers.cpp
@@ -7,7 +7,7 @@ namespace routing
 {
 using namespace traffic;
 
-void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
+void ReconstructRoute(IDirectionsEngine & engine, IRoadGraph const & graph,
                       shared_ptr<TrafficInfo::Coloring> const & trafficColoring,
                       my::Cancellable const & cancellable, m2::PointD const & start,
                       m2::PointD const & finish, vector<Junction> & path, Route & route)
@@ -30,8 +30,7 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
   // @TODO(bykoianko) streetNames is not filled in Generate(). It should be done.
   Route::TStreets streetNames;
   vector<TrafficInfo::RoadSegmentId> trafficSegs;
-  if (engine)
-    engine->Generate(graph, path, times, turnsDir, junctions, trafficSegs, cancellable);
+  engine.Generate(graph, path, times, turnsDir, junctions, trafficSegs, cancellable);
 
   if (cancellable.IsCancelled())
     return;
--- a/routing/routing_helpers.cpp
+++ b/routing/routing_helpers.cpp
@@ -33,6 +33,16 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
   if (engine)
     engine->Generate(graph, path, times, turnsDir, junctions, trafficSegs, cancellable);
 
+  if (cancellable.IsCancelled())
+    return;
+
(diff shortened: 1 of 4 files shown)


routing/routing_helpers.cpp
```diff
@@ -33,6 +33,16 @@ void ReconstructRoute(IDirectionsEngine * engine, IRoadGraph const & graph,
   if (engine)
     engine->Generate(graph, path, times, turnsDir, junctions, trafficSegs, cancellable);
 
+  if (cancellable.IsCancelled())
+    return;
+
+  // In case of any errors in IDirectionsEngine::Generate() |junctions| is empty.
+  if (junctions.empty())
```
ygorshenin: What errors may happen in IDirectionsEngine::Generate?

ygorshenin: It does not return anything, isn't it? How you derived that in case of errors in directions engine junctions is empty, if we assume that directions engine works without errors?

dobriy-eeh:
LGTM

routing/routing_helpers.cpp
bykoianko: If it comes to crash in fabric (https://fabric.io/mapsme/android/apps/com.mapswithme.maps.pro/issues/5858f58b0aeb16625b352b6d)

The most likely routing was canceled while reconstructing path/generating turns. I check it two line above.

But as for your question, in BicycleDirectionsEngine::Generate() and PedestrianDirectionsEngine::Generate() may be:
1. If path.size() == 1 => junctions is empty;
2. If ReconstructPath() returns false => junctions is empty;

ygorshenin: But it's not an error, right?

bykoianko: In case path.size() == 1, it's not an error. But in case of ReconstructPath() it is an error.
Do you think it's better to remove assert (LOG(LERROR,...) from here?

bykoianko:
@ygorshenin PTAL

routing/routing_helpers.cpp
kshalnev: but junctions will be empty if engine is null.

bykoianko: @kshalnev Yes, you're right. But I'd prefer to go without doing it right now. I guess it's a question of different PR to another branch.

kshalnev: without 'engine != nullptr' fix is not correct, replacing pointer by reference - is up to you.

ygorshenin:
LGTM

routing/single_mwm_router.cpp
```diff
@@ -264,7 +264,8 @@ bool SingleMwmRouter::BuildRoute(MwmSet::MwmId const & mwmId, vector<Joint::Id>
   shared_ptr<traffic::TrafficInfo::Coloring> trafficColoring = m_trafficCache.GetTrafficInfo(mwmId);
 
   auto const numJunctions = junctions.size();
-  ReconstructRoute(m_directionsEngine.get(), m_roadGraph, trafficColoring, delegate, start, finish,
+  CHECK(m_directionsEngine, ());
```
ygorshenin: For uniformity with ctor, replace CHECK by ASSERT.

bykoianko: I'd prefer to use CHECK in such cases to be sure that the app is terminated and write to log the line before * m_directionsEngine.

bykoianko:
DUDE JTTP

PR #5077, opened 2016-12-21
[search] Support hyphen in the Wi-Fi category.

mpimenov:
@ygorshenin PTAL


ygorshenin:
Tests?

data/categories.txt
```diff
@@ -5565,35 +5565,35 @@ sk:Oceán
 sw:Bahari
 
 @wifi
-en:WiFi
-ru:WiFi
+en:WiFi|Wi-Fi
+ru:WiFi|Wi-Fi
 ar:وايفاي
```
jancborchardt: So will it also work for Arabic, Japanese and Chinese, if this is not fixed? Or is that what is done in the `android/res/values/strings.xml` part? :)

indexer/categories_holder.cpp
```diff
@@ -225,9 +225,11 @@ bool CategoriesHolder::ValidKeyToken(StringT const & s)
     return true;
 
   /// @todo We need to have global stop words array for the most used languages.
-  for (char const * token : {"a", "z", "s", "d", "di", "de", "le", "wi", "fi", "ra", "ao"})
```
ygorshenin: ROFL

ygorshenin:
LGTM

ygorshenin:
@therearesomewhocallmetim PTAL

data/categories.txt
mpimenov: @jancborchardt I don't think we need to change the synonyms for these languages in this PR. I'm not sure if we have to change them at all.

mpimenov:
PTAL

therearesomewhocallmetim:
LGTM

mpimenov:
PTAL & merge

mpimenov:
PTAL

jancborchardt:
Awesome, thanks a lot! :)

PR #5116, opened 2016-12-30
[search][android] Fixed downloader callback behavior.

ygorshenin:
Fixed downloader callback behavior - for consistency with other search callbacks results are accumulated in the results vector.

@alexzatsepin @igrechuhin PTAL


ygorshenin:
@mpimenov FYI

PR #5139, opened 2017-01-09
[android] Navigation buttons margings on small screens

goblinr:
http://jira.65apps.com/browse/MAP-100

@alexzatsepin 
@mpimenov


mpimenov:
LGTM

PR #5171, opened 2017-01-12
[android] Bookmark edit refactoring

goblinr:
Modify bookmark only when user pressed "save" button.

@alexzatsepin 
@mpimenov


mpimenov:
LGTM

PR #5226, opened 2017-01-18
[location] [ios] Removed location reporting in background.

igrechuhin:


syershov:
Нужно ли сделать копию на релиз?

PR #5250, opened 2017-01-20
[android] Extracted all UI logic from the LocationHelper class to Mwm…

alexzatsepin:
…Activity


alexzatsepin:
@milchakov @rokuz PTAL

PR #5275, opened 2017-01-24
[android] Move "large font" setting from misc to map settings

goblinr:
Move "large font" setting from misc to map settings(https://jira.mail.ru/browse/MAPSME-3453).

@alexzatsepin 
@mpimenov


mpimenov:
LGTM

PR #5292, opened 2017-01-27
[search] TF-IDF for localities.

ygorshenin:
@mpimenov @bykoianko PTAL

### DIFF
--- a/search/geocoder.cpp
+++ b/search/geocoder.cpp
@@ -641,48 +641,9 @@ void Geocoder::FillLocalityCandidates(BaseContext const & ctx, CBV const & filte
                                       size_t const maxNumLocalities,
                                       vector<Locality> & preLocalities)
 {
-  preLocalities.clear();
-
-  for (size_t startToken = 0; startToken < ctx.m_numTokens; ++startToken)
-  {
-    CBV intersection = filter.Intersect(ctx.m_features[startToken]);
-    if (intersection.IsEmpty())
-      continue;
-
-    CBV unfilteredIntersection = ctx.m_features[startToken];
-
-    for (size_t endToken = startToken + 1; endToken <= ctx.m_numTokens; ++endToken)
-    {
-      // Skip locality candidates that match only numbers.
-      if (!m_params.IsNumberTokens(startToken, endToken))
-      {
-        intersection.ForEach([&](uint32_t featureId)
-                             {
-                               Locality l;
-                               l.m_countryId = m_context->GetId();
-                               l.m_featureId = featureId;
-                               l.m_startToken = startToken;
-                               l.m_endToken = endToken;
-                               l.m_prob = static_cast<double>(intersection.PopCount()) /
-                                          static_cast<double>(unfilteredIntersection.PopCount());
-                               preLocalities.push_back(l);
-                             });
-      }
-
-      if (endToken < ctx.m_numTokens)
-      {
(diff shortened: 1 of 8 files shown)


search/geocoder.cpp
```diff
@@ -745,7 +706,7 @@ void Geocoder::FillLocalitiesTable(BaseContext const & ctx)
 
 #if defined(DEBUG)
         ft.GetName(StringUtf8Multilang::kDefaultCode, city.m_defaultName);
-        LOG(LDEBUG, ("City =", city.m_defaultName, radius));
+        LOG(LINFO, ("City =", city.m_defaultName, "radius =", radius, "tfidf =", l.m_tfidf));
```
bykoianko: LDEBUG?

search/locality_scorer.cpp
```diff
@@ -55,44 +109,34 @@ void LocalityScorer::GetTopLocalities(size_t limit, vector<Geocoder::Locality> &
...
+    return lhs.m_rank > rhs.m_rank;
+  });
+
+  std::vector<ExLocality> rs;
+  std::unordered_set<uint32_t> ids;
+
+  for (size_t i = 0; i < ls.size() && rs.size() < limit; ++i)
+  {
+    bool const inserted = ids.insert(ls[i].GetId()).second;
+    if (inserted)
```
bykoianko: if (ids.insert(ls[i].GetId()).second) ?

search/geocoder.cpp
ygorshenin: Why? This code is compiled only when if defined(DEBUG), so there's no reason in LDEBUG.

search/locality_scorer.cpp
ygorshenin: This is the same, but more readable from my point of view.

search/geocoder.cpp
bykoianko: Yes, that's right. But are you sure that you want to have |LINFO| under #ifdef DEBUG?
At the first glance it'll be print in release version. I don't insist. It's just for your information.

search/locality_scorer.cpp
bykoianko: Ok.

search/geocoder.cpp
ygorshenin: Sorry? This code is not compiled in the release version.

search/search_tests/locality_scorer_test.cpp
```diff
@@ -20,80 +25,77 @@ using namespace strings;
...
-  size_t const numTokens = params.GetNumTokens();
+    m_names[featureId].push_back(name);
+  }
 
-  for (size_t startToken = 0; startToken != numTokens; ++startToken)
+  void GetTopLocalities(size_t limit)
   {
-    for (size_t endToken = startToken + 1; endToken <= numTokens; ++endToken)
+    BaseContext ctx;
+    ctx.m_usedTokens.assign(m_params.GetNumTokens(), false);
```
bykoianko: Please add inline comment for false

ygorshenin: Could you please suggest the best comment?

bykoianko: Ok. Let's leave as is.

ygorshenin:
The benefits from this request are quite arguable, we decided to temporarily freeze it.

PR #5402, opened 2017-02-10
[android] Added milliseconds to log record

alexzatsepin:
https://jira.mail.ru/browse/MAPSME-3831
@mpimenov PTAL

### DIFF
--- a/android/src/com/mapswithme/util/log/FileLoggerStrategy.java
+++ b/android/src/com/mapswithme/util/log/FileLoggerStrategy.java
@@ -18,6 +18,7 @@ import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.Locale;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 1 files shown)


android/src/com/mapswithme/util/log/FileLoggerStrategy.java
```diff
@@ -18,6 +18,7 @@
 import java.io.FileWriter;
 import java.io.IOException;
 import java.text.DateFormat;
+import java.text.SimpleDateFormat;
```
mpimenov: Sort imports?

mpimenov:
Why do we need such precision?

android/src/com/mapswithme/util/log/FileLoggerStrategy.java
alexzatsepin: Imports are managed by IDE, manual sorting is no needed. And it looks sorted, doesn't it?

alexzatsepin:
This precision is needed to make our QA able to test the parameter 'Geo location request interval' which depends on my position mode. For example, in car routing that parameter is 500 ms.

android/src/com/mapswithme/util/log/FileLoggerStrategy.java
mpimenov: Oh, I missed it's text*, and the next is util*.

PR #5445, opened 2017-02-17
Upgraded stb library

rokuz:


syershov:
Please remove data/gps_track.dat from request.

PR #5465, opened 2017-02-22
[search] A compiler for faster processing of hotel filters.

ygorshenin:
@mpimenov PTAL

### DIFF
--- a/base/stl_add.hpp
+++ b/base/stl_add.hpp
@@ -173,6 +173,11 @@ struct IdFunctor
   }
 };
 
+template <typename T>
+struct Void
+{
+};
+
 template <class T> struct EqualFunctor
 {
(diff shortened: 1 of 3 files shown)


mpimenov:
s/VM/Virtual machine for filters/
or
s/VM/A compiler for filters/

mpimenov:
Do you have any benchmarks supporting this PR?

ygorshenin:
No, I have no benchmarks.

mpimenov:
Could you please provide some benchmarks?
Note that our testing facility measures time so a sufficiently large test should suffice.

ygorshenin:
Seems that on moderate-size requests on real maps this approach is 1.5 times slower than the previous one. Let's close the PR.

PR #5475, opened 2017-02-27
[search] Use MemTrie for street synonyms.

ygorshenin:
@mpimenov @bykoianko PTAL


indexer/search_string_utils.cpp
```diff
@@ -94,11 +96,67 @@ char const * kStreetTokensSeparator = "\t -,.";
...
+
+      created = true;
+      m_subtrees.emplace_back(c, make_unique<SubTree>());
+      return *m_subtrees.back().second;
+    }
+
+    void Clear() { m_subtrees.clear(); }
+
+  private:
+    buffer_vector<pair<Char, std::unique_ptr<SubTree>>, 8> m_subtrees;
```
bykoianko: Do you really need buffer_vector in this case? vector?

ygorshenin: Yes, I need. Please, take a look at the old implementation of this class - it used buffer_vector in nodes, as the branching factor is quite low, therefore it's ok to pay for unused space.

base/mem_trie.hpp
```diff
@@ -157,10 +212,9 @@ class MemTrie
     auto const * cur = &m_root;
     for (auto const & c : key)
     {
-      auto const it = cur->m_moves.find(c);
-      if (it == cur->m_moves.end())
-        return nullptr;
-      cur = it->second.get();
+      cur = cur->m_moves.GetSubTree(c);
```
mpimenov: s/SubTree/Subtree/g

indexer/search_string_utils.cpp
```diff
@@ -94,11 +96,67 @@ char const * kStreetTokensSeparator = "\t -,.";
...
-  using TStrings = search::StringSet<UniChar, 8>;
+  struct BooleanSum
+  {
+    void Add(bool value) { m_value = m_value || value; }
+
+    template <typename ToDo>
+    void ForEach(ToDo && toDo) const
+    {
+      if (m_value)
+        toDo(m_value);
```
mpimenov: So this is toDo(true)?
I think that either the check should be moved to the client (as in ```toDo(m_value)```) or the argument should be removed.

indexer/search_string_utils.cpp
```diff
@@ -94,11 +96,67 @@ char const * kStreetTokensSeparator = "\t -,.";
...
+    }
+
+    void Clear() { m_value = false; }
+
+    bool m_value = false;
+  };
+
+  template <typename Char, typename SubTree>
+  class Moves
```
mpimenov: Probably renaming this to BufferVectorMoves and moving closer to MapMoves is cleaner.

ygorshenin: This is quite custom implementation of branching, I really don't want to expose it out of this translation unit.

indexer/search_string_utils.cpp
ygorshenin: Unfortunately I don't see an easy way to get rid of the argument.

ygorshenin: Moreover, without value you won't be able to distinguish between internal nodes and nodes that mean end of some string.

base/mem_trie.hpp
ygorshenin: Done

ygorshenin:
@mpimenov @bykoianko PTAL

indexer/search_string_utils.cpp
```diff
@@ -168,24 +225,32 @@ class StreetsSynonymsHolder
     for (auto const * s : affics)
     {
       UniString const us = NormalizeAndSimplifyString(s);
-      m_strings.Add(us.begin(), us.end());
+      m_strings.Add(us, true);
```
kshalnev: add comment /* who is true */?

base/mem_trie.hpp
```diff
@@ -5,15 +5,76 @@
...
+  std::map<Char, std::unique_ptr<Subtree>> m_subtrees;
+};
+
+template <typename Value>
+struct VectorValues
+{
+  template <typename V>
+  void Add(V && v)
+  {
+    m_values.push_back(std::forward<V>(v));
```
kshalnev: emplace_back?

indexer/search_string_utils.cpp
mpimenov: How does the value help you? It is always true as far as toDo is concerned.

base/mem_trie.hpp
```diff
@@ -170,8 +224,7 @@ class MemTrie
   template <typename ToDo>
   void ForEachInNode(Node const & node, String const & prefix, ToDo && toDo) const
   {
-    for (auto const & value : node.m_values)
-      toDo(prefix, value);
+    node.m_values.ForEach(std::bind(std::forward<ToDo>(toDo), prefix, std::placeholders::_1));
```
kshalnev: ref(prefix)? чтоб не копировать?

indexer/search_string_utils.cpp
ygorshenin: Please, take a look at the current version of this function. If is removed.

indexer/search_string_utils.cpp
ygorshenin: Done

base/mem_trie.hpp
ygorshenin: Done

base/mem_trie.hpp
ygorshenin: Errr, unfortunately we don't use cpp14, that's why I can't use auto in lambdas. But in the case of bind, thanks, it's a right thing to use ref here.

ygorshenin:
@bykoianko @kshalnev PTAL

kshalnev:
LGTM

PR #5488, opened 2017-02-28
[build] Use system-provided boost instead of 3party in cmake.

ygorshenin:
@milchakov @syershov PTAL


therearesomewhocallmetim:
LGTM

PR #5543, opened 2017-03-09
Edit FAQ - add <wbr> tags to paths for no horizontall scroll

Alexuser1:
Edit FAQ - add <wbr> tags to paths for no horizontall scroll


Zverik:
OK

PR #5555, opened 2017-03-10
Http client fix encoding for osrm

milchakov:


milchakov:
@bykoianko @syershov PTAL

platform/http_client.cpp
```diff
@@ -10,7 +10,12 @@ namespace platform
 {
 HttpClient::HttpClient(string const & url) : m_urlRequested(url)
 {
+// Http client for Linux is not supported gzip encoding.
```
mpimenov: does not support?

milchakov: а, да

bykoianko:
LGTM

PR #5594, opened 2017-03-14
Added debug commands for changing map style.

darina:


map/framework.cpp
```diff
@@ -1394,6 +1394,9 @@ string Framework::GetCountryName(m2::PointD const & pt) const
 
 bool Framework::Search(search::SearchParams const & params)
 {
+  if (ParseDrapeDebugCommand(params.m_query))
```
ygorshenin: Could you please group this check with ParseEditorDebugCommand()? For example, you may create a helper function, say:
```c++
// Returns true when |command| is a debug command.
bool ParseDebugCommand(std::string const & command)
...
```
and call it there.

map/framework.cpp
```diff
@@ -2868,6 +2871,29 @@ uint32_t GetBestType(FeatureType const & ft)
...
+  else if (query == "?light" || query == "mapstyle:light")
+    desiredStyle = MapStyleClear;
+  else if (query == "?vdark" || query == "mapstyle:vehicle_dark")
+    desiredStyle = MapStyleVehicleDark;
+  else if (query == "?vlight" || query == "mapstyle:vehicle_light")
+    desiredStyle = MapStyleVehicleClear;
+  else
+    return false;
+
+#ifdef OMIM_OS_ANDROID
```
```c++
#if defined(OMIM_OS_ANDROID)
```
is the same, but more flexible as you may combine it later with other checks.

map/framework.hpp
```diff
@@ -620,6 +620,7 @@ class Framework : public search::ViewportSearchCallback::Delegate,
   //void GetLocality(m2::PointD const & pt, search::AddressInfo & info) const;
   /// @returns true if command was handled by editor.
   bool ParseEditorDebugCommand(search::SearchParams const & params);
+  bool ParseDrapeDebugCommand(std::string const & query);
```
ygorshenin: Comment to the method, please.

map/framework.cpp
darina: done

map/framework.hpp
darina: done

PR #5657, opened 2017-03-21
[feature_list] Update feature_list for new feeds

Zverik:
Немного изменений в feature_list, которые давно использовал, но не доходили руки сделать PR.

### DIFF
--- a/feature_list/feature_list.cpp
+++ b/feature_list/feature_list.cpp
@@ -23,6 +23,7 @@
 #include "storage/index.hpp"
 #include "storage/storage.hpp"
 
+#include "std/algorithm.hpp"
 #include "std/iostream.hpp"
 
 class ClosestPoint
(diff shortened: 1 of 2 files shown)


feature_list/feature_list.cpp
```diff
@@ -83,6 +84,25 @@ string GetReadableType(FeatureType const & f)
   return result == 0 ? string() : classif().GetReadableObjectName(result);
 }
 
+string GetWheelchairType(FeatureType const & f)
```
ygorshenin: Please, put this in the anonymous namespace.

Zverik: What is the point? Feature_list is a standalone program, not a library. This is one of the functions related to its operation. Should I put all the functions inside an anonymous namespace? Is that a commonly-used practice?

ygorshenin: Ah, see now. As this is not a library, okay.

Meanwhile, it's a good practice to put all what you don't want to expose outside to the anonymous namespaces, even in main files.

PR #5690, opened 2017-03-24
Added ICU library sources.

darina:
* Added i18n and common ICU libraries sources.
* Added ICU .dat package with base data and transliterators data.
http://source.icu-project.org/repos/icu/tags/release-57-1/icu4c/
http://apps.icu-project.org/datacustom/


Zverik:
LGTM задним числом

PR #5696, opened 2017-03-27
[search] Fixed some warnings.

mpimenov:


search/geocoder.cpp
```diff
@@ -1002,7 +1005,9 @@ void Geocoder::MatchPOIsAndBuildings(BaseContext & ctx, size_t curToken)
     InitLayer(SearchModel::SEARCH_TYPE_BUILDING, m_postcodes.m_tokenRange, layer);
 
     vector<uint32_t> features;
-    m_postcodes.m_features.ForEach(MakeBackInsertFunctor(features));
+    m_postcodes.m_features.ForEach([&features](uint64_t bit) {
+	features.push_back(static_cast<uint32_t>(bit));
```
syershov: base::asserted_cast ?

mpimenov: done

CMakeLists.txt
```diff
@@ -240,6 +240,12 @@ add_compile_options(
   "-std=c++11"
 )
 
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(
```
ygorshenin: Why not on all compilers?

mpimenov: As far as I understand, gcc does not support this option. I could not find it here:
https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html

PR #5780, opened 2017-04-06
Fixed uri tests.

VladiMihaylenko:
https://tools.ietf.org/html/rfc3986#page-11

6.2.2.1. Case Normalization

When a URI uses components of the generic syntax, the component syntax equivalence rules always apply; namely, that the scheme and host are case-insensitive and therefore should be normalized to lowercase. For example, the URI is equivalent to http://www.example.com/.

The other generic syntax components are assumed to be case-sensitive unless specifically defined otherwise by the scheme (see Section 6.2.3).


VladiMihaylenko:
DUDE JTTP

map/map_tests/mwm_url_tests.cpp
```diff
@@ -207,12 +209,12 @@ UNIT_TEST(MapApiPointNameOverwritten)
   ApiTest api("mapswithme://map?ll=1,2&n=A&N=B");
   TEST(api.IsValid(), ());
   TEST_EQUAL(api.GetPointCount(), 1, ());
-  TEST(api.TestName(0, "B"), ());
+  TEST(api.TestName(0, "A"), ());
```
mpimenov: Теперь не соответствует названию теста.
Добавь сюда же ещё один c n=A&n=B.

VladiMihaylenko: ok

PR #5784, opened 2017-04-06
Generator test fix.

bykoianko:
@dobriy-eeh @ygorshenin @mpimenov PTAL

### DIFF
--- a/generator/generator_tests/restriction_collector_test.cpp
+++ b/generator/generator_tests/restriction_collector_test.cpp
@@ -30,8 +30,7 @@ string const kRestrictionTestDir = "test-restrictions";
 
 UNIT_TEST(RestrictionTest_ValidCase)
 {
-  RestrictionCollector restrictionCollector("" /* restrictionPath */,
-                                            "" /* osmIdsToFeatureIdsPath */);
+  RestrictionCollector restrictionCollector;
   // Adding feature ids.
   restrictionCollector.AddFeatureId(30 /* featureId */, osm::Id::Way(3));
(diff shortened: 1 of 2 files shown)


bykoianko:
DUDE JTTP

generator/generator_tests/restriction_collector_test.cpp
```diff
@@ -30,8 +30,7 @@ string const kRestrictionTestDir = "test-restrictions";
 
 UNIT_TEST(RestrictionTest_ValidCase)
 {
-  RestrictionCollector restrictionCollector("" /* restrictionPath */,
-                                            "" /* osmIdsToFeatureIdsPath */);
+  RestrictionCollector restrictionCollector;
```
ygorshenin: Please, explain how this helps?

ygorshenin: I don't see the difference.

bykoianko: The problem was that in tests ctor RestrictionCollector::RestrictionCollector() was called with empty strings as parameter. Its params are paths to files. So inside the ctor an exception was thrown. This exception was caught in a catcher.
Before todays merges the catcher wrote a warning message to log.
After todays merge the catcher writes error message to log and debug version of the tests stop passing.

To fix it (and to stop parsing empty file path) I add a default ctor and use it in tests for the time being. On the other hand the default ctor could be used if we need to create an instance of RestrictionCollector and add restriction to in by calling methods then. Because of it I didn't add a comment that default ctor should be used for testing only.

ygorshenin: Okay, see now, thanks!

PR #5910, opened 2017-04-25
Add UA and VIE voice instructions for routing

Alexuser1:
subj


Zverik:
Note that this PR will have to be redone after #6290.

Zverik:
Added in #6451.

PR #5975, opened 2017-05-03
Fix various errors in sv translations

lokal-profil:
Fixed some typos and several entries which were translations of the
wrong English homonym.

Also added "Cash machine" as a (Brittish) English alias for ATM.

### DIFF
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -177,7 +177,7 @@ pl:^Sklep
 pt:^Compras
 ro:^Magazin
 es:^Tienda
-sv:^Butik
+sv:^Butik|Affär
 th:^ร้านค้า
 tr:^Mağaza
 uk:^Крамниця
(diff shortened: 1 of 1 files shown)


lokal-profil:
Anything I can do to help get this merged?

burivuh:
Hi! We'll definitely check this out! Sorry for such a delay: we had a very busy month

data/categories.txt
```diff
@@ -7186,7 +7190,7 @@ pl:^Wejście
 pt:^Entrada
 ro:^Intrare
 es:^Entrada
-sv:^Entré
+sv:^Entré|ingång
```
mpimenov: The rest seem ok, but here I'd like to double check.
The symbol '^' in this file marks the translation that will be shown in the UI. Is it true that "Entré" is more commonly used than "Ingång"?

lokal-profil:
Largely depends on the context.
In general "entré" is a slightly larger "ingång" and can also refer to the area imediately inside the door.

Ingång can be any entrance or opening.

PR #6006, opened 2017-05-05
[platform] Check the result of Settings::Get.

mpimenov:


platform/settings.hpp
```diff
@@ -4,6 +4,8 @@
 #include "std/map.hpp"
 #include "std/mutex.hpp"
 
+#include "base/macros.hpp"
```
ygorshenin: Base should be before std.

mpimenov: done

mpimenov:
PTAL

PR #6009, opened 2017-05-05
[android] fix wrong download size calculation

goblinr:
https://jira.mail.ru/browse/MAPSME-4433

@alexzatsepin PTAL


goblinr:
@alexzatsepin PTAL

PR #6034, opened 2017-05-11
[traffic] Fixed scary bug in OpenLR matching tool.

ygorshenin:


openlr/router.cpp
```diff
@@ -557,8 +557,12 @@ size_t Router::FindPrefixLengthToConsume(It b, It const e, double lengthM)
...
   {
-    auto const & u = b->first;
-    auto const & v = b->second;
+    // Need p here to prolongate lifetime of (*b) if iterator
+    // dereferencing returns a temprorary object instead of a
+    // reference.
+    auto const & p = *b;
+    auto const & u = p.first;
+    auto const & v = p.second;
     double const len = MercatorBounds::DistanceOnEarth(u, v);
```
mpimenov: ```double const len = MercatorBounds::DistanceOnEarth(p.first, p.second);``` ?

mgsergio:
LGTM

openlr/router.cpp
ygorshenin: I'd prefer the old code.

PR #6035, opened 2017-05-11
[ios] Fixed crash while deleting bookmark.

VladiMihaylenko:
https://jira.mail.ru/browse/MAPSME-3866

### DIFF
--- a/iphone/Maps/Bookmarks/BookmarksRootVC.mm
+++ b/iphone/Maps/Bookmarks/BookmarksRootVC.mm
@@ -9,6 +9,10 @@
 
 #define TEXTFIELD_TAG 999
 
+extern NSString * const kBookmarkCategoryDeletedNotification =
+    @"BookmarkCategoryDeletedNotification";
+
+
 @implementation BookmarksRootVC
 
 - (id)init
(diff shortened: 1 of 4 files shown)


iphone/Maps/Bookmarks/BookmarksRootVC.mm
```diff
@@ -9,6 +9,10 @@
 
 #define TEXTFIELD_TAG 999
 
+extern NSString * const kBookmarkCategoryDeletedNotification =
```
syershov: Extern keyword is not need here.

iphone/Maps/Bookmarks/BookmarksVC.mm
```diff
@@ -27,6 +27,7 @@
 #define EMPTY_SECTION -666
 
 extern NSString * const kBookmarksChangedNotification = @"BookmarksChangedNotification";
+extern NSString * const kBookmarkDeletedNotification = @"BookmarkDeletedNotification";
```
syershov: Here too.

VladiMihaylenko: __private_extern__ is enabled in our project

iphone/Maps/Bookmarks/BookmarksRootVC.mm
syershov: It's really need in our project, because we set GCC_SYMBOLS_PRIVATE_EXTERN.

PR #6045, opened 2017-05-12
Do not interrupt kinetic scroll by single arrow animation.

darina:


drape_frontend/animation/animation.cpp
```diff
@@ -90,4 +90,45 @@ bool Animation::GetMaxDuration(Interpolator const & interpolator, double & maxDu
   return true;
 }
 
+std::string DebugPrint(Animation::Type const & type)
+{
+  if (type == Animation::Type::Sequence)
```
ygorshenin: Please, use switch-case instead.

drape_frontend/animation/animation.cpp
```diff
@@ -90,4 +90,45 @@ bool Animation::GetMaxDuration(Interpolator const & interpolator, double & maxDu
...
+  if (type == Animation::Type::Arrow)
+    return "Arrow";
+  if (type == Animation::Type::KineticScroll)
+    return "KineticScroll";
+  return "Unknown type";
+}
+
+std::string DebugPrint(Animation::Object const & object)
+{
+  if (object == Animation::Object::MyPositionArrow)
```
ygorshenin: Please, use switch-case instead.

drape_frontend/animation/animation.cpp
```diff
@@ -90,4 +90,45 @@ bool Animation::GetMaxDuration(Interpolator const & interpolator, double & maxDu
...
+  if (object == Animation::Object::MapPlane)
+    return "MapPlane";
+  if (object == Animation::Object::Selection)
+    return "Selection";
+  return "Unknown object";
+}
+
+std::string DebugPrint(Animation::ObjectProperty const & property)
+{
+  if (property == Animation::ObjectProperty::Position)
```
ygorshenin: Please, use switch-case instead.

darina: done

drape_frontend/animation/animation.cpp
```diff
@@ -90,4 +90,41 @@ bool Animation::GetMaxDuration(Interpolator const & interpolator, double & maxDu
...
+  switch (type)
+  {
+  case Animation::Type::Sequence: return "Sequence";
+  case Animation::Type::Parallel: return "Parallel";
+  case Animation::Type::MapLinear: return "MapLinear";
+  case Animation::Type::MapScale: return "MapScale";
+  case Animation::Type::MapFollow: return "MapFollow";
+  case Animation::Type::Arrow: return "Arrow";
+  case Animation::Type::KineticScroll: return "KineticScroll";
+  default: return "Unknown type";
```
ygorshenin: Please, don't use default unless you're really want to handle all possible cases. Just
```return "Unknown type";``` at the end of the function.

darina: done

PR #6076, opened 2017-05-18
[platform] Read resources and writable paths from environment on macOS.

mpimenov:


platform/platform_mac.mm
```diff
@@ -72,6 +81,9 @@
     }
   }
 
+  m_resourcesDir += '/';
```
kshalnev: м.б. проверять наличие завершающего '/' в конце? во всех ранее существовавших случаях он и так ставился. // в путях не критично, но не красиво.

ygorshenin: Do we really need trailing slashes?

mpimenov: @kshalnev done
@ygorshenin Yes,
1. It is what this file does in other cases.
2. Otherwise we cant errors of the kind "cannot open omim/datasettings.ini".

platform/platform_mac.mm
```diff
@@ -72,6 +82,11 @@
     }
   }
 
+  if (!strings::EndsWith(m_resourcesDir, "/"))
+    m_resourcesDir += "/";
```
ygorshenin: As far as I know we have AppendSlashIfNeeded or something like that.

mpimenov:
PTAL

mpimenov:
Merge?

PR #6240, opened 2017-06-09
Update sound.csv

fafa127:
Add Persian Language


Zverik:
Note that this PR will have to be redone after #6290.

Zverik:
Added in #6451, for any fixes make a new pull request.

PR #6255, opened 2017-06-13
Fixed pt and pt-BR translations

marceldallagnol:
Hello!

I downloaded the app recently and it has been working great. Great work!
Since there were a few mistakes in the translation to Portuguese, I reviewed the strings.txt file, trying to follow your standard procedure for commits. I hope I did everything correctly.

There are a few things I didn't do, though, both because I don't know how to solve and because I don't know if my changes would be helpful. There are two things that came to mind (I could create issues for both if necessary):

1. Capitalization is not consistent throughout the file, with some strings with every word capitalized (like "Menu de Aplicação") while other languages only have the first letter of the string capitalized (like "Menú de aplicación"). I could have standardized capitalizing only the first letter (as would look more natural in Portuguese, at least), but it would be better to know if there is a standard to follow. If not, I would suggest standardizing that way (and I'd be more than willing to send another PR with the changes).
2. One of the things that prompted me to correct translations was the fact that, even though my phone is set for Brazilian Portuguese (pt-BR), the language on the app is Portuguese from Portugal (pt, or pt-PT). Looking through the code, I realized the strings for pt-BR are not being generated at all, even though they are present in the strings.txt file (as can be seen by the fact that `android/res/values-pt/strings.xml` and `iphone/Maps/LocalizedStrings/pt.lproj/Localizable.strings` are generated, while no corresponding file for pt-BR is; also, none of the pt-BR strings are present in these files). From what I could understand, it seems that the Twine command `generate-all-string-files` (in `generate_localizations.sh`) doesn't generate them. If you could give me any pointers on how to include pt-BR to be generated, I'd be glad to do so as well.

Cheers!
Marcel


marceldallagnol:
Sure! If there's anything I can do to help get pt-BR localization working, let me know.

Zverik:
Wait, now you have deleted a file, while I expected that you would just undo changes :)
Could you please restore the Chinese strings as they were, and also [sign this CLA](http://maps.me/cla/) so we could use your changes?

marceldallagnol:
Sure! I just signed and restored the Chinese strings (the other file, that was not present before my commit, shouldn't be, right?)

Zverik:
Rebased and re-done in #6463.

PR #6339, opened 2017-06-20
[ugc] Fake UGC are generated in API now.

ygorshenin:
@VladiMihaylenko @alexzatsepin PTAL


ygorshenin:
Actually, please, don't use MakeTestUGC1 and MakeTestUGC2 directly, use them implicitly from GetUGC().

VladiMihaylenko:
LGTM

PR #6487, opened 2017-07-04
[routing] Progress adjusting.

bykoianko:
@dobriy-eeh @rokuz

### DIFF
--- a/routing/index_router.cpp
+++ b/routing/index_router.cpp
@@ -360,7 +360,7 @@ IRouter::ResultCode IndexRouter::AdjustRoute(m2::PointD const & startPoint,
                                     false /* strictForward */),
       graph);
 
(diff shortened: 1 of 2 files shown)


routing/index_router.cpp
```diff
@@ -360,7 +360,7 @@ IRouter::ResultCode IndexRouter::AdjustRoute(m2::PointD const & startPoint,
                                     false /* strictForward */),
       graph);
 
-  AStarProgress progress(0, 100);
+  AStarProgress progress(0, 95);
```
ygorshenin: Could you please explain why do we need this? I don't see explaining comment or Jira ticket.

rokuz: Обсудили

PR #6623, opened 2017-07-17
Added default ctor for CSVReader.

VladiMihaylenko:


coding/csv_file_reader.hpp
```diff
@@ -16,6 +16,8 @@ class CSVReader
     char m_delimiter = ',';
   };
 
+  CSVReader() = default;
```
mpimenov: https://google.github.io/styleguide/cppguide.html#Declaration_Order

PR #6637, opened 2017-07-19
[cherry-pick][partners_api] taxi engine crash fix

milchakov:
https://fabric.io/mapsme/android/apps/com.mapswithme.maps.pro/issues/58cab62a0aeb16625b782e18/sessions/596DC09702D200016C85BA6DD1699CC3_59d0faf26b9011e79b0256847afe9799_0_v2?build=60932702

### DIFF
--- a/partners_api/uber_api.cpp
+++ b/partners_api/uber_api.cpp
@@ -218,6 +218,10 @@ void ProductMaker::MakeProducts(uint64_t const requestId, ProductsCallback const
 
     if (m_error)
       error = my::make_unique<taxi::ErrorCode>(*m_error);
+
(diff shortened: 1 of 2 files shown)


partners_api/uber_api.cpp
```diff
@@ -218,6 +218,10 @@ void ProductMaker::MakeProducts(uint64_t const requestId, ProductsCallback const
 
     if (m_error)
       error = my::make_unique<taxi::ErrorCode>(*m_error);
+
+    // Reset m_times and m_prices because we need to call callback only once.
```
mpimenov: Достаточно "We need to call callback only once."

PR #6661, opened 2017-07-21
[jansson] get from json object nullable

milchakov:


milchakov:
@mpimenov @ygorshenin @mgsergio PTAL

milchakov:
DUDE JTTP

PR #6684, opened 2017-07-24
Update categories.txt

ibraheem5000:
Update Arabic

### DIFF
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -5028,7 +5028,7 @@ sw:^Ghuba
 natural-spring
 en:^Spring
 ru:^Родник
-ar:^ربيع|طبيعة
+ar:^ينبوع
 cs:^Pramen
 da:^Kilde|kildevæld
 nl:^Bron
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -2344,7 +2344,7 @@ sw:^Makumbusho|kivutio|utalii
 waterway-waterfall|@tourism
 en:2^Waterfall
 ru:2^Водопад|достопримечательность
-ar:^شلال|مناظر
+ar:^شلال|مناظر|سياحة
 cs:2^Vodopád|zajímavost
 da:2^Vandfald
 nl:2^Watervallen|waterval
@@ -2376,7 +2376,7 @@ sw:^Maporomoko ya maji
 historic-memorial|@tourism
 en:^Memorial|monument|attraction|tourism|U+1F5FC|U+1F5FD|U+1F5FF
 ru:^Памятник|мемориал|достопримечательность|туризм
-ar:^مبنى او نصب تزكارى
+ar:^نصب تذكاري|معالم|آثار|أماكن جذابة|سياحة
 cs:^Pomník|památka|zajímavost
 da:^Mindesmærke|mindesten
 nl:^Gedenkteken|monument|attractie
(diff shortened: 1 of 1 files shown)


data/categories.txt
```diff
@@ -2344,7 +2344,7 @@ sw:^Makumbusho|kivutio|utalii
 waterway-waterfall|@tourism
 en:2^Waterfall
 ru:2^Водопад|достопримечательность
-ar:^شلال|مناظر
+ar:^شلال|مناظر|سياحة
```
mpimenov: If I'm reading it correctly, the '^' sign (translation to be used in the UI) should point to "شلال".
Do we really need the other two synonyms here?

data/categories.txt
```diff
@@ -2440,7 +2440,7 @@ sw:^Makumbusho|kivutio|utalii
 historic-ruins|@tourism
 en:^Ruins|attraction|tourism|U+1F5FF
 ru:^Руины|развалины|достопримечательность|туризм
-ar:^آثار|أماكن الجذب|مناظر
+ar:^آثار|أماكن جذابة|مناظر|سياحة
```
mpimenov: 1. Is the word "ruins" here somewhere?
2. Again, please make sure the '^' sign points to the main translation here and in other places. (It still should be to the left of the required word.)

data/categories.txt
```diff
@@ -2472,7 +2472,7 @@ sw:^Maeneo ya kale|jengo la kale|chakavu
 historic-castle|@tourism
 en:^Castle|attraction|tourism|U+1F3EF|U+1F3F0|U+1F451|U+1F478
 ru:^Замок|достопримечательность|туризм
-ar:^قلعةأماكن الجذب|مناظر
+ar:^قلعة|سياحة|أماكن جذابة|مناظر
```
mpimenov: Again, looks like "قلعة" is the primary synonym to "castle"?

data/categories.txt
```diff
@@ -5027,7 +5028,7 @@ sw:^Ghuba
 natural-spring
 en:^Spring
 ru:^Родник
-ar:^خليجc
+ar:^ربيع|طبيعة
```
mpimenov: No! It is this type of spring: https://en.wikipedia.org/wiki/Spring_(hydrology).

data/categories.txt
ibraheem5000: Yes .. the '^' sign point to "شلال" .. that happen because Arabic is Right-to-Left language.
And Yes the two other synonyms needed because if someone search by these other synonyms.
Thanks.

data/categories.txt
ibraheem5000: 1. Yes, The "ruins" here is somewhere!
2.Yes the '^' points to the main translation .. and we can say it exist Immediately before the main translation, not left of the required word because the Arabic language typing from Right-to-Left.

data/categories.txt
ibraheem5000: Yes it is.

data/categories.txt
ibraheem5000: Oh !
It's my mistake...
I correct it.
Thanks

data/categories.txt
mpimenov: Sure, but mixing left-to-right and right-to-left may be confusing.
Apparently you are right here.

PR #6695, opened 2017-07-24
Disabled transparent 3d buildings on Android 4.2 and less

rokuz:
Cherry-pick https://github.com/mapsme/omim/pull/6693


rokuz:
Нашел workaround! Закрываю

PR #6819, opened 2017-08-08
viator-strings

igortomko:


strings.txt
```diff
@@ -25906,3 +25906,11 @@
 [choose_destination]
     en = Choose a destination
     ru = Выберите конец маршрута
+
+[preloader_viator_title]
+    en = Discovery tours and unusual things to do
```
mpimenov: Discover?

strings.txt
```diff
@@ -25906,3 +25906,11 @@
...
+    ru = Откройте для себя экскурсии и необычные туры
+
+[preloader_viator_message]
+    en = Tours, things to do, sightseeing tours and more
```
mpimenov: И tours, и sightseeing?

Хотя бы просто
```Tours, things to do, sightseeing and more```

PR #6840, opened 2017-08-09
[search] Cian support in search index.

ygorshenin:
@mpimenov PTAL

Added support in the search index for buildings with empty names.


ygorshenin:
https://jira.mail.ru/browse/MAPSME-5192

PR #6894, opened 2017-08-15
[base] Switch from uint8_t to size_t for errors counts in LevenshteinDFA

ygorshenin:


mpimenov:
Why?

mpimenov:
Per offline discussion: to lesser the amount of casts and improve readability.

PR #6952, opened 2017-08-22
[routing] Fixed a data race in AltitudeLoader.

mpimenov:
AltitudeLoader holds a FileReader of the altitudes section.
This reader, being a caching one, is not thread safe. However,
AltitudeLoader did not keep the MwmHandle from which it was
supposed to construct its reader, so the reader could be reused
(via MwmValue via MwmHandle) by another thread.

FeaturesRoadGraph tries to avoid this issue by keeping its
own cache of MwmHandles but it did not stop RoadGeometry
from bypassing this cache when creating its AltitudeLoader.

The question of whether we need to fix this other cache
and whether we need it at all is out of the scope of this CL.

https://jira.mail.ru/browse/MAPSME-5407

### DIFF
--- a/indexer/altitude_loader.cpp
+++ b/indexer/altitude_loader.cpp
@@ -29,8 +29,14 @@ void LoadAndMap(size_t dataSize, ReaderSource<FilesContainerR::TReader> & src, T
 
 namespace feature
 {
-AltitudeLoader::AltitudeLoader(MwmValue const & mwmValue)
+AltitudeLoader::AltitudeLoader(Index const & index, MwmSet::MwmId const & mwmId)
+  : m_handle(index.GetMwmHandleById(mwmId))
 {
+  if (!m_handle.IsAlive())
+    return;
+
+  auto const & mwmValue = *m_handle.GetValue<MwmValue>();
+
   m_countryFileName = mwmValue.GetCountryFileName();
 
   if (mwmValue.GetHeader().GetFormat() < version::Format::v8)
--- a/indexer/altitude_loader.hpp
+++ b/indexer/altitude_loader.hpp
@@ -1,6 +1,7 @@
 #pragma once
 #include "indexer/feature_altitude.hpp"
 #include "indexer/index.hpp"
+#include "indexer/mwm_set.hpp"
 
 #include "coding/memory_region.hpp"
 
@@ -15,7 +16,7 @@ namespace feature
 class AltitudeLoader
 {
 public:
-  explicit AltitudeLoader(MwmValue const & mwmValue);
+  explicit AltitudeLoader(Index const & index, MwmSet::MwmId const & mwmId);
 
   /// \returns altitude of feature with |featureId|. All items of the returned vector are valid
   /// or the returned vector is empty.
@@ -36,5 +37,6 @@ private:
   map<uint32_t, TAltitudes> m_cache;
   AltitudeHeader m_header;
   string m_countryFileName;
+  MwmSet::MwmHandle m_handle;
 };
 }  // namespace feature
--- a/indexer/altitude_loader.hpp
+++ b/indexer/altitude_loader.hpp
@@ -16,7 +16,7 @@ namespace feature
 class AltitudeLoader
 {
 public:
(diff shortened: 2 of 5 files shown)


mpimenov:
@rokuz Полагаю, этот PR придётся делать на обе ветки. Поскольку сейчас release-75 не собирается, сначала сюда.

indexer/altitude_loader.cpp
```diff
@@ -29,8 +29,14 @@ void LoadAndMap(size_t dataSize, ReaderSource<FilesContainerR::TReader> & src, T
 
 namespace feature
 {
-AltitudeLoader::AltitudeLoader(MwmValue const & mwmValue)
+AltitudeLoader::AltitudeLoader(Index const & index, MwmSet::MwmId const & mwmId)
```
ygorshenin: Why not to pass handle here, instead of two params?

indexer/altitude_loader.hpp
```diff
@@ -15,7 +16,7 @@ namespace feature
 class AltitudeLoader
 {
 public:
-  explicit AltitudeLoader(MwmValue const & mwmValue);
+  explicit AltitudeLoader(Index const & index, MwmSet::MwmId const & mwmId);
```
ygorshenin: Explicit for two parameter ctor? No need imho.

ygorshenin:
Sorry, could you please explain the problem in detail? I still don't see the data race. AltitudesLoader gets reader from m_cont in mwm value and as far as I know this is a thread safe operation.

routing/features_road_graph.hpp
```diff
@@ -87,7 +88,7 @@ class FeaturesRoadGraph : public IRoadGraph
   struct Value
   {
     Value() = default;
-    explicit Value(MwmSet::MwmHandle handle);
+    explicit Value(Index const & index, MwmSet::MwmHandle handle);
```
tatiana-kondakova: Здесь тоже теперь не нужно explicit

rokuz:
@mpimenov Проверил, iOS и Android собираются на release-75

indexer/altitude_loader.cpp
mpimenov: That would work too. I chose this way for symmetry with FeaturesLoaderGuard: while we know we can reuse handles in the current model (i.e. all routing calls made on one thread), it seems safer to have a separate handle for each involved component exactly for the reason of this PR.

mpimenov:
> AltitudesLoader gets reader from m_cont in mwm value and as far as I know this is a thread safe operation.

1. AltitudesLoader's reader (from a tag in m_cont, indeed): https://github.com/mapsme/omim/blob/master/indexer/altitude_loader.cpp#L44
2. This reader may seem to be a mere ModelReaderPtr https://github.com/mapsme/omim/blob/master/coding/file_container.hpp#L116
But in fact it uses an underlying ModelReaderPtr  m_source https://github.com/mapsme/omim/blob/master/coding/file_container.hpp#L130
https://github.com/mapsme/omim/blob/master/coding/file_container.cpp#L70
for every tag in a file, which in turn points to a FileReader.
3. Despite allowing only read operations, FileReader is not thread safe (and is marked as such)
because of its internal caching: https://github.com/mapsme/omim/blob/master/coding/file_reader.cpp#L34

So we either have to revamp our caching policy, or at least use FileReaders as
intended, i.e. stop trying to read using an MwmHandle that went out of scope.

indexer/altitude_loader.cpp
ygorshenin: If you'd like to avoid re-use of handles, pass it by value here, not by const ref. So, it won't be possible to share it. But I'm not insisting.

ygorshenin:
Thanks fo the explanation. Even if you'll rewrite caching policy on FileReader, it still won't be thread safe because of corresponding entry in the process file table is not thread safe.

indexer/altitude_loader.hpp
mpimenov: done

routing/features_road_graph.hpp
mpimenov: done

PR #7039, opened 2017-09-06
Fixed the build.

mpimenov:


openlr/openlr_match_quality/openlr_assessment_tool/mainwindow.cpp
```diff
@@ -118,7 +118,7 @@ class PointsControllerDelegate : public PointsControllerDelegateBase
     : m_framework(framework)
     , m_index(framework.GetIndex())
     , m_roadGraph(m_index, routing::IRoadGraph::Mode::ObeyOnewayTag,
-                  make_unique<routing::CarModelFactory>())
+                  make_unique<routing::CarModelFactory>(routing::CountryParentNameGetterFn{}))
```
ygorshenin: Not LGTM. Why not to initialize CarModelFactory in a right way?

PR #7065, opened 2017-09-12
corr routing_add_finish_point fr translation

tbertels:

PR #7066, opened 2017-09-12
corr routing_add_finish_point fr translation

tbertels:


mpimenov:
This is a duplicate PR for #7065

PR #7070, opened 2017-09-12
[search] Adapted the sweep line algorithm to match POIs with streets.

mpimenov:


search/features_layer_matcher.hpp
```diff
@@ -181,36 +183,74 @@ class FeaturesLayerMatcher
...
-          uint32_t const streetId = street.m_id.m_index;
-          if (binary_search(streets.begin(), streets.end(), streetId))
-            fn(poiId, streetId);
-        }
+      for (size_t j = 0; j + 1 < streetFt.GetPointsCount(); ++j)
+      {
+        auto const & p1 = streetFt.GetPoint(j);
+        auto const & p2 = streetFt.GetPoint(j + 1);
+        m2::RectD rect(p1, p2);
+        rect.Inflate(inflationRect.SizeX(), inflationRect.SizeY());
```
ygorshenin: Looks like you need to Inflate by SizeX() * 0.5 and SizeY() * 0.5, right?

mpimenov: In fact, even 0.25 is enough: RectByCenterXYAndSizeInMeters treats size as a radius, thus creating a 2S x 2S rectangle for the size parameter S.
After that, the Inflate() method adds strips of size dx both from the left and from the right.

done

ygorshenin: Disagree - we need 0.5. Imagine a case when the street is horizontal or vertical - we need the strip of size 2 * kStreetRadiusMeters around the street.

search/segment_tree.hpp
```diff
@@ -64,11 +64,30 @@ class SegmentTree
...
+  void FindAny(double x, Fn && fn) const
   {
-    return Find(0 /* index */, x, fn);
+    FindAny(0 /* index */, x, fn);
+  }
+
+  // Calls |fn| on all segments containing |x| making
+  // exactly one call per segment.
+  template <typename Fn>
+  void FindAll(double x, Fn && fn)
```
ygorshenin: Looks good now, did you notice any speedups?

search/features_layer_matcher.hpp
```diff
@@ -181,36 +185,81 @@ class FeaturesLayerMatcher
...
+                            [&](size_t poiId, size_t streetId) {
+                              ASSERT_LESS(poiId, pois.size(), ());
+                              ASSERT_LESS(streetId, streets.size(), ());
+                              auto const & poiCenter = poiCenters[poiId].m_point;
+                              ProjectionOnStreet proj;
+                              if (streetProjectors[streetId].GetProjection(poiCenter, proj) &&
+                                  proj.m_distMeters < kStreetRadiusMeters)
+                              {
+                                auto const key = make_pair(poiId, streetId);
+                                if (fnAlreadyCalled.find(key) != fnAlreadyCalled.end())
```
kshalnev: м.б. короче:
```
if (fnAlreadyCalled.insert(key).second)
    fn(pois[poiId], streets[streetId]);
```

search/features_layer_matcher.hpp
mpimenov: Ah, it's because we need 2S x 2S in the end, as we treat a radius as, well, a radius. And not as the side length.

done

search/segment_tree.hpp
mpimenov: ~4x over the previous version
~150x over the original

search/features_layer_matcher.hpp
mpimenov: done

mpimenov:
@kshalnev @ygorshenin PTAL

search/features_layer_matcher.hpp
```diff
@@ -181,36 +185,79 @@ class FeaturesLayerMatcher
...
+      streetPoints.reserve(streetFt.GetPointsCount());
+      for (size_t j = 0; j < streetFt.GetPointsCount(); ++j)
+        streetPoints.emplace_back(streetFt.GetPoint(j));
+      streetProjectors.emplace_back(streetPoints);
     }
+
+    BailIfCancelled(m_cancellable);
+    // The matcher may call |fn| more than once if a POI falls inside
+    // the bounding rectangles of several street segments.
+    set<pair<size_t, size_t>> fnAlreadyCalled;
```
ygorshenin: Please, replace set of pairs by a vector of bools, the size of vector should be the same as the number of pois.

mpimenov: I move the check up the call tree.

PR #7073, opened 2017-09-12
Additional logging and crash fix

VladiMihaylenko:


storage/storage.cpp
```diff
@@ -1470,6 +1476,7 @@ bool Storage::IsPossibleToAutoupdate() const
 
 void Storage::OnDiffStatusReceived(diffs::Status const status)
 {
+  LOG(LINFO, ("Received diff status: ", status));
```
mpimenov: Пробел лишний

PR #7086, opened 2017-09-14
[geometry] ConvexHull.

ygorshenin:


geometry/convex_hull.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+  auto const n = hull.size();
+  if (n < 2)
+    return true;
+
+  auto const & p1 = hull[n - 2];
+  auto const & p2 = hull[n - 1];
+
+  // Checks whether (p2 - p1) x (p - p2) > 0.
+  return robust::OrientedS(p1, p, p2) < -kEps;
```
mpimenov: I suggest you reuse IsCCW here and remove the comment.

geometry/convex_hull.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+double const kEps = 1e-9;
+
+// Checks whether (p1 - p) x (p2 - p) > 0.
+bool IsCCW(PointD const & p1, PointD const & p2, PointD const & p)
+{
+  return robust::OrientedS(p1, p2, p) > kEps;
+}
+
+bool ContinuesHull(vector<PointD> const & hull, PointD const & p)
```
mpimenov: The word "Continues" doesn't seem right here. Extends? FitsInto?
Also, consider swapping the arguments (or changing to passive voice: IsContinuedBy(hull, p)).

geometry/convex_hull.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include <algorithm>
+
+using namespace std;
+
+namespace m2
+{
+namespace
+{
+double const kEps = 1e-9;
```
mpimenov: Please add a comment on why you chose this epsilon (even if the choice was arbitrary).

geometry/convex_hull.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "geometry/point2d.hpp"
+
+#include <vector>
+
+namespace m2
+{
+// Builds a convex hull around |points|.
```
mpimenov: The hull polygon does not have three points lying on the same straight line.
The hull polygon points are listed in the order of a counterclockwise traversal.

geometry/convex_hull.cpp
ygorshenin: How to reuse IsCCW here?

geometry/convex_hull.cpp
ygorshenin: IsContinuedBy SGTM

geometry/convex_hull.hpp
ygorshenin: Done

geometry/convex_hull.cpp
mpimenov:
```
    (p2 - p1) x (p - p2) > 0
≡ { anticommutativity of cross product }
    - (p - p2) x (p2 - p1) > 0
≡ { s(a x b) = (a x sb) for scalar s }
    (p - p2) x (p1 - p2) > 0
≡
    IsCCW(p, p1, p2)
```

ygorshenin: Thanks, done!

geometry/convex_hull.cpp
ygorshenin: Done

ygorshenin:
@mpimenov PTAL

geometry/convex_hull.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+#include "geometry/point2d.hpp"
+
+#include <vector>
+
+namespace m2
+{
+// Builds a convex hull around |points|.  The hull polygon points are
+// listed in the order of a counterclockwise traversal.  The hull
+// polygon does not have three points lying on the same straight line.
```
mpimenov: The hull polygon points are listed in the order of a counterclockwise traversal with no three points lying on the same straight line.

(I repeated the same prefix but aligned the sentences vertically. With automatic format it's better to avoid repetition.)

geometry/convex_hull.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+using namespace std;
+
+namespace m2
+{
+namespace
+{
+// 1e-12 is used here because of we're going to use the convex hull on
+// Mercator plane, where the precision of all coords is 1e-5, so we
+// off by two orders of magnitude from the precision of data.
```
mpimenov: what is the verb here? "we are off"?

PR #7137, opened 2017-09-21
Fix qmake directory detection for brew package installer on MacOSX.

greshilov:
Brew installs qt with specific version in /usr/local/opt/qt@5.?/ 
This path is not specified in detect_qmake.sh, that causes problem with building project.


alexzatsepin:
@therearesomewhocallmetim PTAL

therearesomewhocallmetim:
LGTM

PR #7144, opened 2017-09-22
Added authentication system

rokuz:
https://jira.mail.ru/browse/MAPSME-5500

### DIFF
--- a/android/src/com/mapswithme/maps/Framework.java
+++ b/android/src/com/mapswithme/maps/Framework.java
@@ -77,6 +77,13 @@ public class Framework
 
   public static final int ROUTE_REBUILD_AFTER_POINTS_LOADING = 0;
 
+  @Retention(RetentionPolicy.SOURCE)
+  @IntDef({ SOCIAL_TOKEN_FACEBOOK, SOCIAL_TOKEN_GOOGLE })
+  public @interface SocialTokenType {}
+
+  public static final int SOCIAL_TOKEN_FACEBOOK = 0;
+  public static final int SOCIAL_TOKEN_GOOGLE = 1;
+
   @SuppressWarnings("unused")
   public interface MapObjectListener
   {
@@ -370,4 +377,8 @@ public class Framework
   public static native void nativeDeleteSavedRoutePoints();
 
   public static native Banner[] nativeGetSearchBanners();
+
+  public static native void nativeAuthenticateUser(@NonNull String socialToken,
+                                                   @SocialTokenType int socialTokenType);
+  public static native boolean nativeIsUserAuthenticated();
 }
--- a/map/user.cpp
+++ b/map/user.cpp
@@ -1,0 +1,165 @@
+#include "map/user.hpp"
+
+#include "platform/http_client.hpp"
+#include "platform/platform.hpp"
+
+#include "coding/url_encode.hpp"
+
+#include "base/logging.hpp"
+
+#include "3party/jansson/myjansson.hpp"
+
+#include <chrono>
+#include <sstream>
+#include <thread>
+
+#define STAGE_PASSPORT_SERVER
+#include "private.h"
+
+namespace
+{
+std::string const kMapsMeTokenKey = "MapsMeToken";
+std::string const kServerUrl = PASSPORT_URL;
+
+std::string AuthenticationUrl(std::string const & socialToken,
+                              User::SocialTokenType socialTokenType)
+{
+  if (kServerUrl.empty())
+    return {};
+
+  std::string socialTokenStr;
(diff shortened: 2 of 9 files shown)


android/src/com/mapswithme/maps/Framework.java
```diff
@@ -77,6 +77,13 @@
 
   public static final int ROUTE_REBUILD_AFTER_POINTS_LOADING = 0;
 
+  @Retention(RetentionPolicy.SOURCE)
+  @IntDef({SOCIAL_TOKEN_FACEBOOK, SOCIAL_TOKEN_FACEBOOK})
```
alexzatsepin: `@IntDef({SOCIAL_TOKEN_FACEBOOK, SOCIAL_TOKEN_FACEBOOK}) -> @IntDef({ SOCIAL_TOKEN_FACEBOOK, SOCIAL_TOKEN_GOOGLE }) `(второй параметр неверный и пробелы пожалуйста)

map/user.cpp
```diff
@@ -0,0 +1,155 @@
...
+std::string const kMapsMeTokenKey = "MapsMeToken";
+std::string const kServerUrl = PASSPORT_URL;
+
+std::string AuthenticationUrl(std::string const & socialToken,
+                              User::SocialTokenType socialTokenType)
+{
+  if (kServerUrl.empty())
+    return {};
+
+  std::string socialTokenStr;
```
alexzatsepin: Ассерт на пустой socialTokenStr и на неизвестный тип токена!?

rokuz: done

map/user.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+private:
+  void SetAccessToken(std::string const & accessToken);
+
+  std::string m_accessToken;
+  mutable std::mutex m_mutex;
+  std::condition_variable m_condition;
+  bool m_needTerminate = false;
+  bool m_authenticationInProgress = false;
+  base::WorkerThread m_workerThread;
```
ygorshenin: Is it good that each User instance has its own worker thread? Do we have some other threads to share with this class?

map/user.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+class User
+{
+public:
+  enum SocialTokenType
+  {
+    Facebook,
+    Google
+  };
+  ~User();
```
ygorshenin: Blank line between enum and dtor, please.

map/user.hpp
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include "base/worker_thread.hpp"
+
+#include <condition_variable>
+#include <mutex>
+#include <string>
+
+class User
```
ygorshenin: Comment to this class, note thread-safety, please.

map/user.cpp
```diff
@@ -0,0 +1,158 @@
...
+        }
+        else if (request.ErrorCode() == 403) // Forbidden.
+        {
+          ResetAccessToken();
+          break;
+        }
+      }
+
+      // Wait for some time and retry.
+      std::unique_lock<std::mutex> lock(m_mutex);
```
ygorshenin: Are you sure it's a good idea to lock mutex for a while in the worker thread and in the same time use it in the public API?

map/user.hpp
rokuz: Обсудили

map/user.cpp
```diff
@@ -0,0 +1,158 @@
...
+      request.SetRawHeader("Accept", "application/json");
+      // TODO: Now passport service uses redirection. If it becomes false, uncomment checking.
+      if (request.RunHttpRequest())// && !request.WasRedirected())
+      {
+        if (request.ErrorCode() == 200) // Ok.
+        {
+          SetAccessToken(ParseAccessToken(request.ServerResponse()));
+          break;
+        }
+        else if (request.ErrorCode() == 403) // Forbidden.
```
ygorshenin: No need in else after break.

map/user.cpp
ygorshenin: Okay, discussed.

map/user.hpp
```diff
@@ -0,0 +1,35 @@
...
+  enum SocialTokenType
+  {
+    Facebook,
+    Google
+  };
+  ~User();
+  void Init();
+
+  void Authenticate(std::string const & socialToken, SocialTokenType socialTokenType);
+  bool IsAuthenticated() const;
```
igrechuhin: мб возвращать SocialTokenType?

rokuz: Обсудили, пока не нужно

PR #7150, opened 2017-09-25
[categories] Updated translations

liondog:
Updated some translations and added translations for railway-level_crossing (OSM equivalent: railway=level_crossing).

### DIFF
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -4899,6 +4899,36 @@ el:^Πισίνα|Αθλητισμός
 sk:4^Kúpalisko|3bazén|aquapark|4kúpele
 sw:^Bwawa la kuogelea
 
+sport-equestrian
+en:Equestrian Sports
+ru:Конный спорт
+de:Pferdesport
+es:Deportes ecuestres
+fr:Sport équestre
+hu:Lovassportok
+nl:Paardensport
+
+sport-basketball
+en:^Basketball
+ru:^Баскетбол
+de:^Basketball
+es:^Baloncesto
+fr:^Basket-ball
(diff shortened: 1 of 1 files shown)


mpimenov:
Please sign the CLA: https://maps.me/cla/

liondog:
I've already signed it.

data/categories.txt
```diff
@@ -1807,6 +1807,21 @@ el:^Υπόγειος|σωλήνας|μετρό|υπόγεια|μεταφορές
...
+de:^Bahnübergang
+it:^Passaggio a livello
+es:^Paso a nivel
+pt:^Passagem de nível
+pl:^Przejazd kolejowy
+cs:^Železniční přejezd
+nl:^Overweg
+da:^Jernbaneoverskæring
+fi:^Tasoristeys
+hu:^Útátjáró
```
mpimenov: What about other languages?

mpimenov:
@Zverik PTAL

data/categories.txt
liondog: What are your contribution guidelines regarding translations? What languages are at least necessary?

mpimenov: The only necessary language is English, but we try to add as many as possible at least in the simple situations such as categories.txt where Grammar is not needed.
Here is the list: https://github.com/mapsme/omim/blob/master/indexer/categories_holder.cpp#L152

mpimenov:
The merge was unintended.
We've discussed this offline and won't revert the changes but please consider adding all languages if you plan to add strings in the future, at least for simple words like "Basketball".

PR #7171, opened 2017-09-27
.[indexer] Simplified boundaries encoding/decoding.

ygorshenin:
Cities boundaries encoding/decoding. When used 19 bits per coord (~100m accuracy), achieves ~20bytes/city compression.

### DIFF
--- a/base/base_tests/visitor_tests.cpp
+++ b/base/base_tests/visitor_tests.cpp
@@ -1,0 +1,45 @@
+#include "testing/testing.hpp"
+
+#include "base/visitor.hpp"
+
+#include <string>
+
+using namespace std;
+
+namespace
+{
+struct Foo
+{
+  DECLARE_VISITOR()
+  DECLARE_DEBUG_PRINT(Foo)
+};
+
+struct Bar
+{
+  Bar() = default;
+  Bar(int d, string const & s) : m_d(d), m_s(s) {}
+
+  DECLARE_VISITOR(visitor(m_d), visitor(m_s, "string"))
+  DECLARE_DEBUG_PRINT(Bar)
+
+  int m_d = 0;
+  string m_s;
+};
+
+UNIT_TEST(DebugPrintVisitor_Smoke)
+{
+  {
+    Foo foo;
+    TEST_EQUAL(DebugPrint(foo), "Foo []", ());
+  }
+
+  {
+    Bar bar;
+    TEST_EQUAL(DebugPrint(bar), "Bar [0, string: ]", ());
+
+    bar.m_d = 7;
+    bar.m_s = "Hello, World!";
+    TEST_EQUAL(DebugPrint(bar), "Bar [7, string: Hello, World!]", ());
+  }
+}
+}  // namespace
--- a/base/visitor.hpp
+++ b/base/visitor.hpp
@@ -1,0 +1,13 @@
+#pragma once
+
+#define DECLARE_VISITOR(...)          \
+  template <typename Visitor>         \
+  void Visit(Visitor & visitor)       \
+  {                                   \
+    __VA_ARGS__;                      \
+  }                                   \
+  template <typename Visitor>         \
+  void Visit(Visitor & visitor) const \
+  {                                   \
+    __VA_ARGS__;                      \
+  }
--- a/base/visitor.hpp
+++ b/base/visitor.hpp
@@ -1,5 +1,36 @@
 #pragma once
 
+#include <sstream>
+#include <string>
+
+namespace base
+{
+class DebugPrintVisitor
+{
+public:
+  DebugPrintVisitor(std::string const & name) : m_name(name) {}
+
+  template <typename T>
+  void operator()(T const & t, char const * name = nullptr)
+  {
+    if (!m_empty)
+      m_os << ", ";
+    m_empty = false;
+
+    if (name)
+      m_os << name << ": ";
+    m_os << DebugPrint(t);
+  }
+
+  std::string ToString() const { return m_name + " [" + m_os.str() + "]"; }
+
+private:
+  std::string m_name;
+  bool m_empty = true;
+  std::ostringstream m_os;
+};
+}  // namespace base
+
 #define DECLARE_VISITOR(...)          \
   template <typename Visitor>         \
   void Visit(Visitor & visitor)       \
@@ -10,4 +41,12 @@
   void Visit(Visitor & visitor) const \
   {                                   \
     __VA_ARGS__;                      \
+  }
+
+#define DECLARE_DEBUG_PRINT(className)               \
+  friend std::string DebugPrint(className const & c) \
+  {                                                  \
+    base::DebugPrintVisitor visitor(#className);     \
+    c.Visit(visitor);                                \
+    return visitor.ToString();                       \
   }
--- a/geometry/bounding_box.cpp
+++ b/geometry/bounding_box.cpp
@@ -1,7 +1,6 @@
 #include "geometry/bounding_box.hpp"
 
 #include <algorithm>
-#include <sstream>
 
 using namespace std;
 
@@ -24,15 +23,5 @@ void BoundingBox::Add(double x, double y)
 bool BoundingBox::HasPoint(double x, double y) const
 {
   return x >= m_min.x && x <= m_max.x && y >= m_min.y && y <= m_max.y;
-}
-
-string DebugPrint(BoundingBox const & bbox)
-{
-  ostringstream os;
-  os << "BoundingBox [ ";
-  os << "min: " << DebugPrint(bbox.Min()) << ", ";
-  os << "max: " << DebugPrint(bbox.Max());
-  os << " ]";
-  return os.str();
 }
 }  // namespace m2
--- a/geometry/bounding_box.cpp
+++ b/geometry/bounding_box.cpp
@@ -1,6 +1,7 @@
 #include "geometry/bounding_box.hpp"
 
 #include <algorithm>
+#include <sstream>
 
 using namespace std;
 
@@ -14,14 +15,24 @@ BoundingBox::BoundingBox(vector<PointD> const & points)
 
 void BoundingBox::Add(double x, double y)
 {
-  m_minX = min(m_minX, x);
-  m_minY = min(m_minY, y);
-  m_maxX = max(m_maxX, x);
-  m_maxY = max(m_maxY, y);
+  m_min.x = min(m_min.x, x);
+  m_min.y = min(m_min.y, y);
+  m_max.x = max(m_max.x, x);
+  m_max.y = max(m_max.y, y);
 }
 
 bool BoundingBox::HasPoint(double x, double y) const
 {
-  return x >= m_minX && x <= m_maxX && y >= m_minY && y <= m_maxY;
+  return x >= m_min.x && x <= m_max.x && y >= m_min.y && y <= m_max.y;
+}
+
+string DebugPrint(BoundingBox const & bbox)
+{
+  ostringstream os;
+  os << "BoundingBox [ ";
+  os << "min: " << DebugPrint(bbox.Min()) << ", ";
+  os << "max: " << DebugPrint(bbox.Max());
+  os << " ]";
+  return os.str();
 }
 }  // namespace m2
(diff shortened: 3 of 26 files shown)


geometry/bounding_box.cpp
```diff
@@ -14,14 +15,24 @@ BoundingBox::BoundingBox(vector<PointD> const & points)
...
+  m_max.y = max(m_max.y, y);
 }
 
 bool BoundingBox::HasPoint(double x, double y) const
 {
-  return x >= m_minX && x <= m_maxX && y >= m_minY && y <= m_maxY;
+  return x >= m_min.x && x <= m_max.x && y >= m_min.y && y <= m_max.y;
+}
+
+string DebugPrint(BoundingBox const & bbox)
```
bykoianko: Не хочешь сделать через визир, коль скоро он уже есть в BoundingBox?

geometry/calipers_box.hpp
```diff
@@ -2,27 +2,37 @@
...
+// than 3, stores convex hull explicitly, otherwise stores smallest
 // rectangle containing the hull. Note that at least one side of the
 // rectangle should contain a facet of the hull, and in general sides
 // of the rectangle may not be parallel to the axes. In any case,
 // CalipersBox stores points in counter-clockwise order.
 class CalipersBox
 {
 public:
+  CalipersBox() = default;
   CalipersBox(std::vector<PointD> const & points);
```
bykoianko: explicit?

geometry/diamond_box.hpp
```diff
@@ -3,20 +3,47 @@
...
+
 namespace m2
 {
 // Bounding box for a set of points on the plane, rotated by 45
 // degrees.
 class DiamondBox
 {
 public:
+  DiamondBox() = default;
+  DiamondBox(std::vector<PointD> const & points);
```
bykoianko: explicit?

geometry/bounding_box.cpp
bykoianko: Комментарий относится и DebugPrint классов, в которых есть визиторы.

indexer/boundary_boxes.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <string>
+#include <vector>
+
+namespace indexer
+{
+struct BoundaryBoxes
+{
+  BoundaryBoxes() = default;
+
+  BoundaryBoxes(std::vector<m2::PointD> const & ps) : m_bbox(ps), m_cbox(ps), m_dbox(ps) {}
```
bykoianko: explicit?

indexer/boundary_boxes.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include "base/visitor.hpp"
+
+#include <sstream>
+#include <string>
+#include <vector>
+
+namespace indexer
+{
+struct BoundaryBoxes
```
bykoianko: final? (здесь и в других аналогичных местах)

indexer/indexer_tests/boundary_boxes_serdes_tests.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+void TestEqual(CalipersBox const & lhs, CalipersBox const & rhs, double eps) {}
+void TestEqual(DiamondBox const & lhs, DiamondBox const & rhs, double eps)
+{
+  auto const lps = lhs.Points();
+  auto const rps = rhs.Points();
+  TEST_EQUAL(lps.size(), 4, (lhs));
+  TEST_EQUAL(rps.size(), 4, (rhs));
+  for (size_t i = 0; i < 4; ++i)
+  {
```
bykoianko: {} можно удалить

indexer/boundary_boxes_serdes.hpp
```diff
@@ -0,0 +1,318 @@
...
+      while (ps.size() != 4)
+      {
+        auto const lp = ps.back();
+        ps.push_back(lp);
+
+        auto const lu = us.back();
+        us.push_back(lu);
+      }
+
+      ASSERT_EQUAL(ps.size(), 4, ());
```
mpimenov: Это лишнее.

geometry/diamond_box.hpp
ygorshenin: Sure!

geometry/calipers_box.hpp
ygorshenin: Sure!

indexer/boundary_boxes.hpp
ygorshenin: Sure!

indexer/boundary_boxes.hpp
ygorshenin: Well, I generally don't think why? From my point of view we need to mark classes as "final" only when we don't want them to be inherited. Not the case here.

indexer/indexer_tests/boundary_boxes_serdes_tests.cpp
ygorshenin: Okay

geometry/calipers_box.cpp
```diff
@@ -147,4 +147,6 @@ bool CalipersBox::HasPoint(PointD const & p) const
   }
   return true;
 }
+
+string DebugPrint(CalipersBox const & cbox) { return "CalipersBox " + ::DebugPrint(cbox.Points()); }
```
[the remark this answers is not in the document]
ygorshenin: Done

indexer/boundary_boxes_serdes.hpp
ygorshenin: Why? The statement is true and assertion for documentation purposes only.

geometry/bounding_box.cpp
ygorshenin: Well, only if I'll find a spare time.

indexer/boundary_boxes.hpp
ygorshenin: Sure!

geometry/bounding_box.cpp
bykoianko: Ok, but while implementing visitors for transit I implemented a visitor for DebugPrint(s). I like the idea of using visitors instead of serialize/deserialize/debugprint/and so on.

indexer/boundary_boxes.hpp
bykoianko: Ok. It's up to you. 

I prefer using the key word "final" and the key word "const" in the same manner. Form my point of view if we don't change a value we mark it as a "const", if we don't inherit from a class I mark it as "final". 

You wrote: "we need to mark classes as "final" only when we don't want them to be inherited". But  are you sure that you what a class without a virtual destructor to be inherited?  (My comment comes to all new classes in your PR.)

indexer/boundary_boxes_serdes.hpp
bykoianko: +1 for this ASSERT. It's line a comment. And It doesn't spend any time in a production built.

base/visitor.hpp
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+#include <sstream>
+#include <string>
+
+namespace base
+{
+class DebugPrintVisitor
```
bykoianko: I propose to write a test on DebugPrintVisitor.

indexer/boundary_boxes.hpp
ygorshenin: I think that absence of a virtual destructor is not a big deal and can be easily fixed, right?

When I say "we don't want them to be inherited" I mean that there're some serious reasons against usage of this class as a base class, and absence of virtual destructor is not the case. So, to me, "final" means "Beware! Author of the class designed it to be a final class in the hierarchy, and there should be some complex invariants that may be violated by subclasses".

ygorshenin: So I don't want to use "final" keyword only because we don't have subclasses right now.

geometry/calipers_box.hpp
```diff
@@ -2,27 +2,37 @@
...
+  bool operator==(CalipersBox const & rhs) const { return m_points == rhs.m_points; }
+
+  DECLARE_VISITOR(visitor(m_points, "points"))
+  DECLARE_DEBUG_PRINT(CalipersBox)
+
 private:
   std::vector<PointD> m_points;
 };
+
+std::string DebugPrint(CalipersBox const & cbox);
```
bykoianko: May you remove this line?

indexer/CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ set(
   SRC
   altitude_loader.cpp
   altitude_loader.hpp
+  boundary_boxes.hpp
```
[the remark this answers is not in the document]
ygorshenin: Done

indexer/boundary_boxes.hpp
ygorshenin: Of course, we can discuss the semantic of the "final" keyword (possibly, with other members of our team), and update the style guide, to prevent disputes in the future.

ygorshenin: What does Meyers think about this?

geometry/calipers_box.hpp
ygorshenin: Sure

base/visitor.hpp
ygorshenin: Good idea, thanks!

indexer/indexer_tests/cities_boundaries_serdes_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+using namespace m2;
+using namespace serial;
+using namespace std;
+
+namespace
+{
+using Boundary = vector<CityBoundary>;
+using Boundaries = vector<Boundary>;
+
+void TestEqual(BoundingBox const & lhs, BoundingBox const & rhs, double eps)
```
bykoianko: TestAlmostEqual()? (Here and below).

ygorshenin: Not sure, why? The method has eps parameter, used only here, what's the reason to complicate namings?

base/base_tests/visitor_tests.cpp
```diff
@@ -0,0 +1,44 @@
+#include "testing/testing.hpp"
+
+#include "base/visitor.hpp"
+
+#include <string>
+using namespace std;
```
bykoianko: A blank line before the line?

base/base_tests/visitor_tests.cpp
```diff
@@ -0,0 +1,44 @@
...
+struct Bar
+{
+  Bar() = default;
+  Bar(int d, std::string const & s) : m_d(d), m_s(s) {}
+
+  DECLARE_VISITOR(visitor(m_d), visitor(m_s, "string"))
+  DECLARE_DEBUG_PRINT(Bar)
+
+  int m_d = 0;
+  std::string m_s;
```
bykoianko: string without std::?

indexer/indexer_tests/cities_boundaries_serdes_tests.cpp
bykoianko: Ok. Let's leave it as TestEqual()

indexer/boundary_boxes.hpp
ygorshenin: For example, here I found an example where final is *really* useful: https://akrzemi1.wordpress.com/2012/09/30/why-make-your-classes-final/

bykoianko: Ok. I see you point of view. Probably it's worth making a meeting one day to agree about using the word final. It would nice if we use "final" key word in the same manner. As for Meyers he wrote about it nothing significant in his "Effective Modern C++ 2014.pdf".

ygorshenin: We can start to follow the following rule: "class should have either virtual dtor or marked as final". WDYT?

ygorshenin: Also, take a look at http://foonathan.net/blog/2016/05/27/final.html - if the class don't have virtual functions, there's no need to mark it as final too.

bykoianko: Thanks for the links. Especially for the second one.

base/base_tests/visitor_tests.cpp
ygorshenin: Done

base/base_tests/visitor_tests.cpp
ygorshenin: Done

PR #7214, opened 2017-10-07
[categories] Updated Finnish translations.

jleh:
Fixed some incorrect Finnish translations in categories.


data/categories.txt
```diff
@@ -7032,7 +7035,7 @@ ar:^ﺓﺭﺎﻴﺴﻟﺍ ءﺍﺰﺟﺃ
 cs:^Autodíly
 da:^Bildele
 nl:^Auto onderdelen
-fi:^Auton osat
+fi:^Autotarvikeet
```
mpimenov: Autotarvikkeet?

jleh: Well maybe "Auton osat" were better translation than new one. Reverted that.

PR #7215, opened 2017-10-09
update docs/MAPS.md

Locke:
Remove outdated instructions about generatig the car routing index. The
`routing_profile` parameter has been removed in b79686c73 and the
intra-mwm routing is since then supposed to be created by default.

Note: the intra-mwm routing is currently not generated without
generating the inter-mwm routing, according to @Zverik this is a known
bug. I'm not sure whether that should be mentioned here. The instructions
for inter-mwm routing are creating intra-mwm routing as well.


Zverik:
Thanks!

PR #7219, opened 2017-10-09
Special handling for pedestrian feature ids.

bykoianko:
При сериализации/десериализции секции transit возникает ситуация, при которой поле одной из структур данных должно быть заполнено не из исходного json, а вычислено после десериализации из json, но перед сериализацией в mwm. Затем, при работе программы оно должно десериализоваться из mwm и использоваться.

@tatiana-kondakova @ygorshenin @mpimenov PTAL

### DIFF
--- a/generator/transit_generator.cpp
+++ b/generator/transit_generator.cpp
@@ -44,6 +44,8 @@ string GetFileName(string const & filePath)
 }
 
 /// \brief Reads from |root| (json) and serializes an array to |serializer|.
+/// \param handler is function which fixes up vector of |Item|(s) after deserialization from json
+/// but before serialization to mwm.
 template <class Item>
 void SerializeObject(my::Json const & root, string const & key, Serializer<FileWriter> & serializer,
                      function<void(vector<Item> &)> handler = nullptr)
@@ -132,12 +134,12 @@ void BuildTransit(string const & mwmPath, string const & transitDir)
   SerializeObject<Stop>(root, "stops", serializer);
   header.m_gatesOffset = base::checked_cast<uint32_t>(w.Pos() - startOffset);
 
-  auto const fillPedestrianFeatureIds = [](function<void(vector<Gate> &)> & handler)
+  auto const fillPedestrianFeatureIds = [](vector<Gate> & gates)
   {
     // @TODO(bykoianko) |m_pedestrianFeatureIds| is not filled from json but should be calculated based on |m_point|.
   };
-  UNUSED_VALUE(fillPedestrianFeatureIds);
-  SerializeObject<Gate>(root, "gates", serializer);
+
+  SerializeObject<Gate>(root, "gates", serializer, fillPedestrianFeatureIds);
   header.m_edgesOffset = base::checked_cast<uint32_t>(w.Pos() - startOffset);
 
   SerializeObject<Edge>(root, "edges", serializer);
--- a/generator/transit_generator.cpp
+++ b/generator/transit_generator.cpp
@@ -15,6 +15,8 @@
 #include "base/logging.hpp"
 #include "base/macros.hpp"
 
+#include <functional>
+
 #include "3party/jansson/src/jansson.h"
 
 using namespace platform;
@@ -43,12 +45,17 @@ string GetFileName(string const & filePath)
 
 /// \brief Reads from |root| (json) and serializes an array to |serializer|.
 template <class Item>
-void SerializeObject(my::Json const & root, string const & key, Serializer<FileWriter> & serializer)
+void SerializeObject(my::Json const & root, string const & key, Serializer<FileWriter> & serializer,
+                     function<void(vector<Item> &)> handler = nullptr)
 {
   vector<Item> items;
 
   DeserializerFromJson deserializer(root.get());
   deserializer(items, key.c_str());
+
+  if (handler)
+    handler(items);
+
   serializer(items);
 }
 }  // namespace
@@ -125,6 +132,11 @@ void BuildTransit(string const & mwmPath, string const & transitDir)
   SerializeObject<Stop>(root, "stops", serializer);
   header.m_gatesOffset = base::checked_cast<uint32_t>(w.Pos() - startOffset);
 
+  auto const fillPedestrianFeatureIds = [](function<void(vector<Gate> &)> & handler)
+  {
+    // @TODO(bykoianko) |m_pedestrianFeatureIds| is not filled from json but should be calculated based on |m_point|.
+  };
+  UNUSED_VALUE(fillPedestrianFeatureIds);
   SerializeObject<Gate>(root, "gates", serializer);
   header.m_edgesOffset = base::checked_cast<uint32_t>(w.Pos() - startOffset);
 
--- a/routing_common/transit_types.hpp
+++ b/routing_common/transit_types.hpp
@@ -104,20 +104,16 @@ public:
   m2::PointD const & GetPoint() const { return m_point; }
 
   DECLARE_VISITOR_AND_DEBUG_PRINT(Gate, visitor(m_featureId, "osm_id"),
-                                  visitor(m_entrance, "entrance"),
-                                  visitor(m_exit, "exit"), visitor(m_weight, "weight"),
-                                  visitor(m_stopIds, "stop_ids"), visitor(m_point, "point"))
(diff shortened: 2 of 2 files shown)


generator/transit_generator.cpp
```diff
@@ -43,12 +45,17 @@ string GetFileName(string const & filePath)
 
 /// \brief Reads from |root| (json) and serializes an array to |serializer|.
 template <class Item>
-void SerializeObject(my::Json const & root, string const & key, Serializer<FileWriter> & serializer)
+void SerializeObject(my::Json const & root, string const & key, Serializer<FileWriter> & serializer,
+                     function<void(vector<Item> &)> handler = nullptr)
```
mpimenov: Please (very briefly) document the handler.

generator/transit_generator.cpp
```diff
@@ -125,6 +132,11 @@ void BuildTransit(string const & mwmPath, string const & transitDir)
   SerializeObject<Stop>(root, "stops", serializer);
   header.m_gatesOffset = base::checked_cast<uint32_t>(w.Pos() - startOffset);
 
+  auto const fillPedestrianFeatureIds = [](function<void(vector<Gate> &)> & handler)
+  {
+    // @TODO(bykoianko) |m_pedestrianFeatureIds| is not filled from json but should be calculated based on |m_point|.
+  };
+  UNUSED_VALUE(fillPedestrianFeatureIds);
```
mpimenov: Why don't you pass it to its SerializeObject? The implementation is empty anyway.

generator/transit_generator.cpp
```diff
@@ -43,12 +45,17 @@ string GetFileName(string const & filePath)
 
 /// \brief Reads from |root| (json) and serializes an array to |serializer|.
 template <class Item>
-void SerializeObject(my::Json const & root, string const & key, Serializer<FileWriter> & serializer)
+void SerializeObject(my::Json const & root, string const & key, Serializer<FileWriter> & serializer,
```
mpimenov: I'm not sure we really need to pass handlers to every instantiation. Perhaps it's better to create a separate function. Up to you.

bykoianko: In case of separate method I'll have a significant amount of copypaste code:
```c++
template <class Item>
void SerializeObject(my::Json const & root, string const & key, Serializer<FileWriter> & serializer)
{
  vector<Item> items;

  DeserializerFromJson deserializer(root.get());
  deserializer(items, key.c_str());

  serializer(items);
}

template <>
void SerializeObject<Gate>(my::Json const & root, string const & key, Serializer<FileWriter> & serializer)
{
  vector<Gate> items;

  DeserializerFromJson deserializer(root.get());
  deserializer(items, key.c_str());

  //Some more code to fill |m_pedestrianFeatureIds|

  serializer(items);
}
```

mpimenov: Looks simpler to me since you don't need to keep track of the calls with non-trivial handler. Unless there are more opinions, up to you.

generator/transit_generator.cpp
bykoianko: Done

generator/transit_generator.cpp
bykoianko: Done

bykoianko:
@tatiana-kondakova @ygorshenin @mpimenov PTAL

routing_common/transit_types.hpp
```diff
@@ -104,20 +104,16 @@ class Gate
   m2::PointD const & GetPoint() const { return m_point; }
 
   DECLARE_VISITOR_AND_DEBUG_PRINT(Gate, visitor(m_featureId, "osm_id"),
-                                  visitor(m_entrance, "entrance"),
-                                  visitor(m_exit, "exit"), visitor(m_weight, "weight"),
-                                  visitor(m_stopIds, "stop_ids"), visitor(m_point, "point"))
+                                  visitor(m_pedestrianFeatureIds, "" /* name */),
```
mpimenov: Does this field needs visiting at all? And where is the visitor for "gates"?

bykoianko: This fields need to be visited with routing::transit::Serializer/Deserializer at routing_commont/transit_serdes.hpp and with DebugPrint. It doesn't need to be visited with DeserializerFromJson()

tatiana-kondakova:
LGTM

routing_common/transit_types.hpp
ygorshenin: Why did you skip name for m_predestrianFeatureIds?

bykoianko: Because the name is not used. There's no field for |m_predestrianFeatureIds| at json. |m_predestrianFeatureIds| is generated while transit section generation based on gate (exit or entrance) coords.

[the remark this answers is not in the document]
bykoianko: We cann't. Because feature ids are kept at the vector (|m_pedestrianFeatureIds|). As I notice before, they are generated on transit section generation step.

bykoianko:
@ygorshenin @mpimenov PTAL

routing_common/transit_types.hpp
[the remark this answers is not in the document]
ygorshenin: So why do you use "osm_id" for visitors? Looks like an error and raises a lot of questions.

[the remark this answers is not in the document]
bykoianko: Are you asking about |m_featureId| field in Gate class?
|osm_id| is a field name at json of transit graph. It has the name because it contains osm id of some feature. The field should be converted to feature id before writing it to mwm and kept at |m_featureId|.

What can we do about it?
* At DeserializerFromJson visitor reading json field name (osm id) and converting its value to feature id, and keeping it at |m_featureId| field;
* Adding field |m_osm_id| to Gate and leaving |m_featureId| field in Gate. In DeserializerFromJson filling |m_osm_id| field of Gate. After that filling |m_featureId| based on |m_osm_id| and probably seting |m_osm_id| to some invalid value. Not writing to mwm |m_osm_id| based on its name while visiting all Gate field to serialize it to mwm.

What do you think about that? The second variant looks better to me.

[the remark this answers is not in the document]
ygorshenin: Both approaches look good, but the second is less hacky.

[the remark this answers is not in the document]
bykoianko: Yes. Let's implement the second approach. But this work is planed after I finish with filling |m_pedestrianFeatureIds| based on gate point. 

In this PR I'm adding |m_pedestrianFeatureIds| without any name and in next PR I'll fill it based on gate point.

bykoianko:
@ygorshenin @mpimenov Let's finish with this PR. I'm about to push the next one.

routing_common/transit_types.hpp
bykoianko: I decided not to skip field name because of debug print.

ygorshenin: It looks like that it's a visitor should decide what fields to skip and what fields not to skip. Missing the field name is generally a bad idea, becase DECLARE_VISITOR declares a generic Visit function that can be used with any visitor, not only with json.

ygorshenin: Therefore I dislike the idea to skip the name for m_pedestrianFeatureIds. If you don't have a field in json, please, modify your json visitor to skip this field.

bykoianko:
> It looks like that it's a visitor should decide what fields to skip and what fields not to skip. Missing the field name is generally a bad idea, becase DECLARE_VISITOR declares a generic Visit function that can be used with any visitor, not only with json.

Yes. Besides that we need DebugPrint to work correctly. And DebugPrint needs all fields have names. In the next PR I added a name for the field. To minimize the number of conflicts I prefer to set the name of the field in the next PR. If you insist to do it right now let's name the field right now.

ygorshenin:
No problem, let's fix that in the next PR.

routing_common/transit_types.hpp
bykoianko: Now the field is not visited with my json visitor because there's no such field in the json. So the field has a default value (it's a empty vector) after deserialization from json. 

> If you don't have a field in json, please, modify your json visitor to skip this field.

How do you propose to modify my json visitor to skip the field? Based on field name?

bykoianko:
@mpimenov PTAL

PR #7259, opened 2017-10-12
Fixed crash in ugc sending.

VladiMihaylenko:


ugc/api.cpp
```diff
@@ -26,7 +26,7 @@ void Api::SetUGCUpdate(FeatureID const & id, UGCUpdate const & ugc)
 
 void Api::GetUGCToSend(UGCJsonToSendCallback const & fn)
 {
-  m_thread.Push([&fn, this] { GetUGCToSendImpl(fn); });
+  m_thread.Push([fn, this] { GetUGCToSendImpl(fn); });
```
ygorshenin: Test?

VladiMihaylenko: there are no unit tests for api yet :(

bykoianko: +1

PR #7308, opened 2017-10-16
Additional ugc test & small fixes

VladiMihaylenko:


ugc/ugc_tests/storage_tests.cpp
```diff
@@ -259,3 +264,43 @@ UNIT_TEST(StorageTest_LoadIndex)
...
+  auto array = my::NewJSONArray();
+  vector<char> data;
+  {
+    using Sink = MemWriter<vector<char>>;
+    Sink sink(data);
+    SerializerJson<Sink> ser(sink);
+    ser(newUGC);
+  }
+
+  data.push_back('\0');
```
milchakov: может сразу использовать `MemWriter<string>`? Кажется @ygorshenin уже предлагал что-то подобное...

milchakov: Опционально

ygorshenin: Yes, is it possible to use string instead of vector here?

ygorshenin: And in the prod code.

ugc/ugc_tests/storage_tests.cpp
```diff
@@ -259,3 +264,43 @@ UNIT_TEST(StorageTest_LoadIndex)
...
+  my::Json ugcNode(data.data());
+  auto embeddedNode = my::NewJSONObject();
+  ToJSONObject(*embeddedNode.get(), "data_version", cafeId.GetMwmVersion() );
+  ToJSONObject(*embeddedNode.get(), "mwm_name", cafeId.GetMwmName());
+  ToJSONObject(*embeddedNode.get(), "feature_id", cafeId.m_index);
```
ygorshenin: Wow, wow, can visitors help you to serialize this stuff?

VladiMihaylenko: It can help in this particular case, but I don't use featureID as a source of these data outside of this unit test. There is a special struct which holds those data which is called ```Storage::UGCIndex```. And I use visitors for its serialization. But do you think it's a good idea to add a visitor into FeatureID only for one unit test?

ugc/ugc_tests/storage_tests.cpp
VladiMihaylenko: да, так и сделаю

VladiMihaylenko:
@ygorshenin ptal

ugc/ugc_tests/storage_tests.cpp
```diff
@@ -259,3 +264,42 @@ UNIT_TEST(StorageTest_LoadIndex)
...
+  auto const oldUGC = MakeTestUGCUpdate(Time(chrono::hours(24 * 10)));
+  auto const newUGC = MakeTestUGCUpdate(Time(chrono::hours(24 * 300)));
+  Storage storage(builder.GetIndex());
+  storage.Load();
+  storage.SetUGCUpdate(cafeId, oldUGC);
+  storage.SetUGCUpdate(cafeId, newUGC);
+  TEST_EQUAL(storage.GetIndexesForTesting().size(), 2, ());
+  auto const toSendActual = storage.GetUGCToSend();
+
+  auto array = my::NewJSONArray();
```
ygorshenin: Please, keep array declaration closer to the place where it's being used.

ugc/ugc_tests/storage_tests.cpp
ygorshenin: No, it's not a good idea, but can't you just override default UGC visitor and add an operator() for FeatureId?

ugc/ugc_tests/storage_tests.cpp
```diff
@@ -259,3 +264,42 @@ UNIT_TEST(StorageTest_LoadIndex)
...
+  ToJSONObject(*embeddedNode.get(), "mwm_name", cafeId.GetMwmName());
+  ToJSONObject(*embeddedNode.get(), "feature_id", cafeId.m_index);
+  ToJSONObject(*ugcNode.get(), "feature", *embeddedNode.release());
+  json_array_append_new(array.get(), ugcNode.get_deep_copy());
+  auto reviewsNode = my::NewJSONObject();
+  ToJSONObject(*reviewsNode.get(), "reviews", *array.release());
+  unique_ptr<char, JSONFreeDeleter> buffer(json_dumps(reviewsNode.get(), JSON_COMPACT | JSON_ENSURE_ASCII));
+  string const toSendExpected(buffer.get());
+  TEST_EQUAL(toSendActual, toSendExpected, ());
+  TEST(DeleteUGCFile(), ());
```
ygorshenin: All these tests call DeleteUGCFile. I suggest you to use UNIT_CLASS_TESTs, and delete ugc files in dtor.

ugc/ugc_tests/storage_tests.cpp
VladiMihaylenko: ok

ugc/ugc_tests/storage_tests.cpp
VladiMihaylenko: done

ugc/ugc_tests/storage_tests.cpp
```diff
@@ -155,107 +160,176 @@ class MwmBuilder
...
+};
+}  // namespace ugc_tests
 
-  storage.SetUGCUpdate(cafeId, cafeUGC);
-  TEST_EQUAL(indexArray.size(), 3, ());
-  TEST(DeleteIndexFile(), ());
-  TEST(DeleteUGCFile(), ());
+using namespace ugc_tests;
+
+UNIT_CLASS_TEST(StorageTest, Smoke)
```
ygorshenin: You're doing it wrong. The right way is to get rid of all methods in StorageTests (except dtor), and write them here:
```c++
UNIT_CLASS_TEST(StorageTest, Smoke)
{
  ... code for the smoke here ...
}
```

VladiMihaylenko: ok :))

PR #7337, opened 2017-10-18
[transit] Conversion osm id to feature id.

bykoianko:
Данный PR решает следующие задачи:
1. обработка osm id представленных в виде строк. Это было необходимо сделать, поскольку osm id требуют uint64_t, а json парсер поддерживает для целого типа только long long (не менее, чем int64_t);
2. конвертация из osm id в feature id. Это реализовано для фичь (обычно, но не всегда точечных), представляющих выходы и остановки;

@ygorshenin @darina @tatiana-kondakova PTAL

### DIFF
--- a/generator/generator_tests/transit_test.cpp
+++ b/generator/generator_tests/transit_test.cpp
@@ -7,6 +7,7 @@
 #include "base/assert.hpp"
 
 #include <string>
+#include <utility>
 #include <vector>
 
 using namespace routing;
@@ -21,7 +22,7 @@ void TestDeserializerFromJson(string const & jsonBuffer, string const & name, ve
   my::Json root(jsonBuffer.c_str());
   CHECK(root.get() != nullptr, ("Cannot parse the json."));
 
-  DeserializerFromJson deserializer(root.get());
+  DeserializerFromJson deserializer(root.get(), make_shared<OsmIdToFeatureIdsMap>());
 
   vector<Obj> objects;
   deserializer(objects, name.c_str());
--- a/generator/generator_tests/transit_test.cpp
+++ b/generator/generator_tests/transit_test.cpp
@@ -115,7 +115,7 @@ UNIT_TEST(DeserializerFromJson_Gates)
     {
       "entrance": true,
       "exit": true,
-      "osm_id": "46116861",
+      "osm_id": "18446744073709551615",
       "point": {
         "x": 43.9290544,
         "y": 68.41120791512581
@@ -133,7 +133,8 @@ UNIT_TEST(DeserializerFromJson_Gates)
 
   auto mapping = make_shared<OsmIdToFeatureIdsMap>();
   (*mapping)[osm::Id(46116860)] = vector<FeatureId>({0});
-  (*mapping)[osm::Id(46116861)] = vector<FeatureId>({2});
+  // Note. std::numeric_limits<uint64_t>::max() == 18446744073709551615
+  (*mapping)[osm::Id(18446744073709551615U)] = vector<FeatureId>({2});
   TestDeserializerFromJson(jsonBuffer, mapping, "gates", expected);
 }
 
--- a/generator/generator_tests/transit_test.cpp
+++ b/generator/generator_tests/transit_test.cpp
@@ -33,6 +33,12 @@ void TestDeserializerFromJson(string const & jsonBuffer, OsmIdToFeatureIdsMap co
     TEST(objects[i].IsEqualForTesting(expected[i]), (objects[i], expected[i]));
 }
 
+template <typename Obj>
+void TestDeserializerFromJson(string const & jsonBuffer, string const & name, vector<Obj> const & expected)
+{
+  return TestDeserializerFromJson(jsonBuffer, OsmIdToFeatureIdsMap(), name, expected);
+}
+
 UNIT_TEST(DeserializerFromJson_TitleAnchors)
 {
   string const jsonBuffer = R"(
@@ -44,7 +50,7 @@ UNIT_TEST(DeserializerFromJson_TitleAnchors)
 
   vector<TitleAnchor> expected = {TitleAnchor(11 /* min zoom */, 4 /* anchor */),
                                   TitleAnchor(14 /* min zoom */, 6 /* anchor */)};
-  TestDeserializerFromJson(jsonBuffer, OsmIdToFeatureIdsMap(), "title_anchors", expected);
+  TestDeserializerFromJson(jsonBuffer, "title_anchors", expected);
 }
 
 UNIT_TEST(DeserializerFromJson_Stops)
@@ -134,7 +140,7 @@ UNIT_TEST(DeserializerFromJson_Gates)
   OsmIdToFeatureIdsMap mapping;
   mapping[osm::Id(46116860)] = vector<FeatureId>({0});
   // Note. std::numeric_limits<uint64_t>::max() == 18446744073709551615
-  mapping[osm::Id(18446744073709551615U)] = vector<FeatureId>({2});
+  mapping[osm::Id(18446744073709551615ULL)] = vector<FeatureId>({2});
   TestDeserializerFromJson(jsonBuffer, mapping, "gates", expected);
 }
 
@@ -166,7 +172,7 @@ UNIT_TEST(DeserializerFromJson_Edges)
     Edge(442018445 /* stop 1 id */, 442018446 /* stop 2 id */, 345.6 /* weight */,
          72551680 /* line id */,  false /* transfer */, {} /* shape ids */)};
 
-  TestDeserializerFromJson(jsonBuffer, OsmIdToFeatureIdsMap(), "edges", expected);
+  TestDeserializerFromJson(jsonBuffer, "edges", expected);
 }
 
 UNIT_TEST(DeserializerFromJson_Transfers)
@@ -191,7 +197,7 @@ UNIT_TEST(DeserializerFromJson_Transfers)
       Transfer(922337203 /* stop id */, {27.5619844, 64.24325959173672} /* point */,
                {209186416, 277039518} /* stopIds */, {} /* anchors */)};
 
-  TestDeserializerFromJson(jsonBuffer, OsmIdToFeatureIdsMap(), "transfers", expected);
+  TestDeserializerFromJson(jsonBuffer, "transfers", expected);
 }
 
 UNIT_TEST(DeserializerFromJson_Lines)
@@ -239,7 +245,7 @@ UNIT_TEST(DeserializerFromJson_Lines)
                                       {246659391, 246659390, 209191855, 209191854, 209191853,
                                        209191852, 209191851} /* stop ids */)};
 
-  TestDeserializerFromJson(jsonBuffer, OsmIdToFeatureIdsMap(), "lines", expected);
+  TestDeserializerFromJson(jsonBuffer, "lines", expected);
 }
 
 UNIT_TEST(DeserializerFromJson_Shapes)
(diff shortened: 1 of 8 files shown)


bykoianko:
Добавил тест, что osmid в графе общественного транспорта может быть корректно прочитано и разобрано, если оно равно максимальному значению от uint64_t.

generator/transit_generator.cpp
```diff
@@ -198,8 +210,15 @@ void DeserializerFromJson::operator()(m2::PointD & p, char const * name)
   FromJSONObject(pointItem, "x", p.x);
   FromJSONObject(pointItem, "y", p.y);
 }
+DeserializerFromJson::DeserializerFromJson(json_struct_t * node,
+                                           shared_ptr<OsmIdToFeatureIdsMap> const & osmIdToFeatureIds)
+  : m_node(node), m_osmIdToFeatureIds(osmIdToFeatureIds)
```
ygorshenin: Why do we need shared pointer here? Why not to pass mapping by const ref?

generator/transit_generator.hpp
```diff
@@ -15,16 +23,35 @@ namespace routing
...
+  DeserializerFromJson(json_struct_t* node, std::shared_ptr<OsmIdToFeatureIdsMap> const & osmIdToFeatureIds);
 
   template<typename T>
   typename std::enable_if<std::is_integral<T>::value || std::is_enum<T>::value || std::is_same<T, double>::value>::type
       operator()(T & t, char const * name = nullptr)
   {
-    GetField(t, name);
+    if (name == nullptr || strcmp(name, "osm_id") != 0)
+    {
+      GetField(t, name);
```
ygorshenin: I strongly suggest you to use OsmId struct and implement deserialization for OsmId individually. This strcmp looks ugly.

generator/transit_generator.cpp
bykoianko: It's a possible solution and I thought about it. And yes. I'll be easier and more efficient.

On the other hand to implement it I need:
- to create an instance of `OsmIdToFeatureIdsMap` in `BuildTransit(..)` method
- to pass it by ref to `DeserializeFromJson(...)` and `SerializeObject(...)` method
- then to pass by ref to constructor of `DeserializerFromJson` and use it in it

I'm afraid of dangling references in the future. What do you think about that?

generator/transit_generator.hpp
bykoianko: Yes. I'll be better I guess. I'll implement that. Thanks.

generator/transit_generator.cpp
ygorshenin: What dangling references do you afraid? OsmIdToFeatureIds mapping should be alive for the whole generator_tool lifetime, what's the problem?

bykoianko: I'm afraid of dangling ref which may appear in the future because of the code would be modified. Nevertheless I don't mind to rewrite it with const ref.

bykoianko:
@ygorshenin @darina @tatiana-kondakova PTAL

generator/transit_generator.cpp
bykoianko: Done

generator/generator_tests/transit_test.cpp
```diff
@@ -121,12 +126,16 @@ UNIT_TEST(DeserializerFromJson_Gates)
...
            {442018474} /* stop ids */, {43.8594864, 68.33320554776377} /* point */),
-      Gate(46116861 /* feature id */, true /* entrance */, true /* exit */, 60.0 /* weight */,
+      Gate(2 /* feature id */, true /* entrance */, true /* exit */, 60.0 /* weight */,
            {442018465} /* stop ids */, {43.9290544, 68.41120791512581} /* point */)};
 
-  TestDeserializerFromJson(jsonBuffer, "gates", expected);
+  OsmIdToFeatureIdsMap mapping;
+  mapping[osm::Id(46116860)] = vector<FeatureId>({0});
+  // Note. std::numeric_limits<uint64_t>::max() == 18446744073709551615
+  mapping[osm::Id(18446744073709551615U)] = vector<FeatureId>({2});
```
ygorshenin: ```U```?
Why not ```LLU```?

ygorshenin: Or, ```ULL```.

generator/generator_tests/transit_test.cpp
```diff
@@ -157,7 +166,7 @@ UNIT_TEST(DeserializerFromJson_Edges)
     Edge(442018445 /* stop 1 id */, 442018446 /* stop 2 id */, 345.6 /* weight */,
          72551680 /* line id */,  false /* transfer */, {} /* shape ids */)};
 
-  TestDeserializerFromJson(jsonBuffer, "edges", expected);
+  TestDeserializerFromJson(jsonBuffer, OsmIdToFeatureIdsMap(), "edges", expected);
```
ygorshenin: Is it possible to implement a short wrapper, with the same name, say:
```c++
void TestDeserializerFromJson(string const & jsonBuffer, string const & name, vector<Obj> const & expected)
{
  return TestDeserializerFromJson(jsonBuffer, OsmIdToFeatureIdsMap{}, name, expected);
}
```
and get rid of these modifications of the old code?

generator/transit_generator.cpp
```diff
@@ -179,13 +182,28 @@ void CalculateEdgeWeight(vector<Stop> const & stops, vector<transit::Edge> & edg
...
   }
 }
+
+void FillOsmIdToFeatureIdMap(string const & osmIdsToFeatureIdPath, OsmIdToFeatureIdsMap & map)
+{
+  CHECK(ForEachOsmId2FeatureId(osmIdsToFeatureIdPath,
+                               [&map](osm::Id const & osmId, uint32_t featureId) {
+                                 map[osmId].push_back(featureId);
+                               }),
+        ());
```
ygorshenin: It would be nice to print at least path in the diagnostic message.

generator/transit_generator.cpp
```diff
@@ -199,7 +217,24 @@ void DeserializerFromJson::operator()(m2::PointD & p, char const * name)
...
 
-void BuildTransit(string const & mwmPath, string const & transitDir)
+void DeserializerFromJson::operator()(OsmId & osmId, char const * name)
+{
+  // Conversion osm id to feature id.
+  string osmIdStr;
+  GetField(osmIdStr, name);
+  CHECK(strings::is_number(osmIdStr), ());
+  uint64_t osmIdNum;
+  CHECK(strings::to_uint64(osmIdStr.c_str(), osmIdNum), ());
```
ygorshenin: You don't need to call c_str(), right?

routing_common/transit_types.cpp
```diff
@@ -125,7 +125,7 @@ bool SingleMwmSegment::IsValid() const
 // Gate -------------------------------------------------------------------------------------------
 Gate::Gate(FeatureId featureId, bool entrance, bool exit, double weight,
            std::vector<StopId> const & stopIds, m2::PointD const & point)
-  : m_featureId(featureId)
+  : m_osmId(featureId)
```
ygorshenin: Do we need to fix param name too?

routing_common/transit_types.hpp
```diff
@@ -76,6 +78,26 @@ struct TransitHeader
...
+  bool IsValid() const { return m_featureId != kInvalidFeatureId; }
+
+  FeatureId GetFeatureId() const { return m_featureId; }
+
+private:
+  DECLARE_TRANSIT_TYPE_FRIENDS
+  DECLARE_VISITOR_AND_DEBUG_PRINT(OsmId, visitor(m_featureId, "feature_id"))
+
+  // |m_featureId| is a feature id corresponding to osm id which presents the class.
+  FeatureId m_featureId = kInvalidFeatureId;
```
ygorshenin: Wow, so what's the difference between OsmId and FeatureId, if any? Why OsmId keeps FeatureId and is identical to FeatureId?

generator/generator_tests/transit_test.cpp
bykoianko: Done

generator/transit_generator.cpp
bykoianko: Done

generator/transit_generator.cpp
bykoianko: Right!

routing_common/transit_types.cpp
bykoianko: I guess, yes. In that case we have param name and field name the same: `DECLARE_VISITOR_AND_DEBUG_PRINT(Gate, visitor(m_osmId, "osm_id"), ...`

routing_common/transit_types.hpp
bykoianko: There's a string field at json with name "osm_id". It contains osm id. This osm id should be converted to feature id before serialization to mwm.

When DeserializerFromJson  meats an OsmId while parsing transit data structure it (DeserializerFromJson) reads the string of osm id, converts the string to uint64_t, convert it to feature id and keep it in OsmId class. Then OsmId class is process regularly at other visitors: Serializer, Deserializer and DebugPrint.

generator/generator_tests/transit_test.cpp
bykoianko: Oh. Yes. I didn't see the second one.

bykoianko: Ok. I changed to `LLU`.

routing_common/transit_types.cpp
bykoianko: Not only name, I guess. But the type of param as well.

bykoianko: I'm adding a commit about that

bykoianko: Don

bykoianko:
@ygorshenin @tatiana-kondakova PTAL

routing_common/transit_types.hpp
ygorshenin: Okay, finally I get it. From my point of view, this is quite dangerous and strange, and looks like abuse of visitors. If you need to parse feature id from json and convert it to osm_id, please, either keep two different structures (one with m_featureId field and another with m_osmId field), parse the first one and then via mapping convert it to the second one. Or, write your own Visit() function that will work with only specified visitors. Current approach is very misleading.

bykoianko: Ok. I'll try with two different structure. If there will be any problems with it Let's write a special method Visit:

```c++
DECLARE_VISITOR_AND_DEBUG_PRINT(Stop, visitor(m_id, "id"), visitor.VisitOsmId(m_osmId, "osm_id"), ...
```

> If you need to parse feature id from json and convert it to osm_id

On the contrary, I need to parse osm id from json and convert it to feature id.

ygorshenin: I mean, if you use visitors to parse to/from json, your internal structure must correspond to the json structure. Your internal structure holds osm id. Your json structure holds feature id. IMHO, that's wrong, and misleading for the readers of the code. Any changes in json structure will result in changes in the DECLARE_VISITOR, and, therefore, will affect other potential visitors. So, either move DECLARE_VISITOR to the private part of the class and declare a couple of friends, either create auxiliary structure with m_featureId as a field, and then convert this structure explicitly to the structure with m_osmId. Note, that you may use templates to avoid code duplication, for example:
```c++
template <typename Id>
struct MyStructureWithId
{
  DECLARE_VISITOR(visitor(m_id, IdTraits<Id>::GetName()));

  template <typename OtherId>
  void BuildFrom(MyStructureWithId<OtherId> const & other, map<OtherId, Id> const & mapping)
  {
     copy all fields
     m_id  = mapping[other.m_id];
  }

  ... other fields ...
  Id m_id;
};

Comment (bykoianko) [reply routing_common/transit_types.hpp]: I added a variant of code which is more understandable I guess, but a little less efficient in term of memory.

> either keep two different structures (one with m_featureId field and another with m_osmId field), parse the first one and then via mapping convert it to the second one.

I added only one class `FeatureIdentifiers` which keep osm id and feature id. And I wrote deserialization from json, serialization and deserialization method for it. PTAL at the last commit.

**Comment by bykoianko** [comment]
@ygorshenin @tatiana-kondakova PTAL

File: routing_common/transit_types.hpp (hunk 8 above)
Comment (bykoianko) [reply routing_common/transit_types.hpp]: > So, either move DECLARE_VISITOR to the private part of the class and declare a couple of friends

BTW It's done now for all transit structures.

```c++
template <typename Id>
struct MyStructureWithId
{
  DECLARE_VISITOR(visitor(m_id, IdTraits<Id>::GetName()));
...
```

Thanks for idea but I did it without template as I wrote before. I guess my current implementation is not misleading. Please have a look at it.

PR #7365, opened 2017-10-20
[ugc] ugc_types.csv is unified

milchakov:


indexer/ftraits.hpp
```diff
@@ -120,6 +120,16 @@ class UGC : public TraitsBase<UGC, UGCItem>
     return {std::istream_iterator<std::string>(iss), std::istream_iterator<std::string>()};
   }
 
+  std::vector<std::string> ParseByDashes(std::string const & str)
+  {
+    std::vector<std::string> result;
```
ygorshenin: We have SimpleTokenizer, why not to use it?
```c++
return Tokenize(str, "-");
```
???

PR #7395, opened 2017-10-24
[android] Returned system locale for ugc, refactored getting locale code

alexzatsepin:


android/src/com/mapswithme/util/Language.java
```diff
@@ -11,21 +14,37 @@
...
+    String lang = Locale.getDefault().toString();
+    if (TextUtils.isEmpty(lang))
+      return Locale.US.toString();
+
+    // Replace deprecated language codes:
+    // in* -> id
+    // iw* -> he
+    if (lang.startsWith("in"))
+      return "id";
+    else if (lang.startsWith("iw"))
```
mpimenov: else лишнее

alexzatsepin: Fixed

PR #7415, opened 2017-10-26
[search] Move viewport to search params.

ygorshenin:


map/framework.cpp
```diff
@@ -1339,6 +1342,8 @@ bool Framework::SearchEverywhere(search::EverywhereSearchParams const & params)
   p.m_query = params.m_query;
   p.m_inputLocale = params.m_inputLocale;
   p.m_mode = search::Mode::Everywhere;
+  if (m_isViewportInitialized)
```
mpimenov: I cannot see these checks in the old code. Could you please explain why they are important?

mpimenov: We've discussed this offline.
Completely filled params are expected by the time they will be used, but no other guarantees are given. So all fields of params are filled here except for viewport, which is filled later for delayed intents.
The old code behaved the same, overwriting the previously written viewport with the actual one for the delayed intents.

It's not clear whether we can fill all params in one place but it doesn't seem that we can.

map/framework.cpp
```diff
@@ -1369,8 +1368,7 @@ bool Framework::SearchInViewport(search::ViewportSearchParams const & params)
   search::SearchParams p;
   p.m_query = params.m_query;
   p.m_inputLocale = params.m_inputLocale;
-  if (m_isViewportInitialized)
-    p.m_viewport = GetCurrentViewport();
+  SetViewportIfPossible(p);  // search request will be delayed if viewport is not available.
```
mpimenov: s/search/Search/

ygorshenin: Done

ygorshenin:
@bykoianko PTAL

search/engine.hpp
```diff
@@ -154,8 +152,8 @@ class Engine
   template <typename... TArgs>
   void PostMessage(TArgs &&... args);
 
-  void DoSearch(SearchParams const & params, m2::RectD const & viewport,
-                shared_ptr<ProcessorHandle> handle, Processor & processor);
+  void DoSearch(SearchParams const & params, shared_ptr<ProcessorHandle> handle,
```
bykoianko: `#include "std/shared_ptr.hpp"`?

ygorshenin:
Merge plz.

PR #7430, opened 2017-10-30
[search] Refactored MakeResult.

mpimenov:


search/ranker.cpp
```diff
@@ -329,20 +329,74 @@ void Ranker::Init(Params const & params, Geocoder::Params const & geocoderParams
...
+  // todo(@m) Used because Result does not have a default constructor. Factor out?
+  auto mk = [&](RankerResult const & r) -> Result {
+    switch (r.GetResultType())
+    {
+    case RankerResult::Type::TYPE_FEATURE:
+    case RankerResult::Type::TYPE_BUILDING:
+      return Result(r.GetID(), r.GetCenter(), name, address,
+                    m_categories.GetReadableFeatureType(type, m_params.m_currentLocaleCode), type,
+                    r.GetMetadata());
+    default:
```
ygorshenin: Please, replace default by:
```c++
case RankerResult::Type::TYPE_LATLON:
   ...
```
and move ASSERTION below the switch.

search/ranker.cpp
```diff
@@ -329,20 +329,74 @@ void Ranker::Init(Params const & params, Geocoder::Params const & geocoderParams
...
 }
 
-Result Ranker::MakeResult(RankerResult const & r) const
+Result Ranker::MakeResult(RankerResult const & rankerResult, bool needAddress,
+                          bool needHighlighting) const
 {
-  Result res = GenerateFinalResult(r, true /* needAddress */);
-  HighlightResult(m_params.m_tokens, m_params.m_prefix, res);
-  if (ftypes::IsLocalityChecker::Instance().GetType(r.GetTypes()) == ftypes::NONE)
+  ReverseGeocoder::Address addr;
```
ygorshenin:
```c++
class AddressGetter 
{
public:
  AddressGetter(ReverseGeocoder const & geocoder, m2::PointD const & center): m_geocoder(geocoder), m_center(center) {}

  ReverseGeocoder::Address const & GetAddress()
  {
    if (m_computed)
      return m_address;
    m_address = m_geocoder.GetNearbyAddress(m_center, m_address);
    m_computed = true;
    return m_address;
  }

private:
  ReverseGeocoder & m_geocoder;
  m2::PointD const m_center;
  ReverseGeocoder::Address m_address;
  bool m_computed = false;
};

Comment (ygorshenin) [reply search/ranker.cpp]: This will simplify somehow the function.

File: search/ranker.cpp [hunk 3]
```diff
@@ -391,21 +445,22 @@ void Ranker::UpdateResults(bool lastUpdate)
...
     {
       if (count >= m_params.m_limit)
         break;
 
-      LOG(LDEBUG, (m_tentativeResults[i]));
-
-      auto const & rankerResult = m_tentativeResults[i];
-      if (m_emitter.AddResult(MakeResult(rankerResult)))
+      LOG(LDEBUG, (rankerResult));
+      Result res = MakeResult(rankerResult, true /* needAddress */, true /* needHighlighting */);
```
ygorshenin:
```c++
auto result = ...
```

mpimenov: The explicit type here is to make it more clear that we're working with Result (and not with RankerResult or PreRankerResult).

search/ranker.cpp
mpimenov: done

search/ranker.cpp
mpimenov: I've simplified even more.

mpimenov:
PTAL @ygorshenin

search/ranker.cpp
```diff
@@ -391,21 +430,22 @@ void Ranker::UpdateResults(bool lastUpdate)
 
     BailIfCancelled();
 
+    auto const & rankerResult = m_tentativeResults[i];
+
     if (m_params.m_viewportSearch)
     {
       // Viewport results don't need calculated address.
-      Result res = GenerateFinalResult(m_tentativeResults[i], false /* needAddress */);
+      Result res = MakeResult(rankerResult, false /* needAddress */, false /* needHighlighting */);
```
ygorshenin: Could you please rename "res" to "result"?

search/ranker.cpp
```diff
@@ -541,8 +532,9 @@ void Ranker::ProcessSuggestions(vector<RankerResult> & vec) const
       GetSuggestion(r, m_params.m_query, m_params.m_tokens, m_params.m_prefix, suggestion);
       if (!suggestion.empty() && added < MAX_SUGGESTS_COUNT)
       {
-        // todo(@m) RankingInfo is not set here. Should it be?
-        if (m_emitter.AddResult(Result(MakeResult(r), suggestion)))
+        // todo(@m) RankingInfo is lost here. Should it be?
+        if (m_emitter.AddResult(Result(
+                MakeResult(r, true /* needAddress */, true /* needHighlighting */), suggestion)))
```
ygorshenin: Looks like we don't need address for suggestions.

search/ranker.cpp
```diff
@@ -329,20 +329,59 @@ void Ranker::Init(Params const & params, Geocoder::Params const & geocoderParams
...
+    ReverseGeocoder::Address addr;
+    m_reverseGeocoder.GetNearbyAddress(rankerResult.GetCenter(), addr);
+
+    // Insert exact address (street and house number) instead of empty result name.
+    if (name.empty() && addr.GetDistance() == 0)
+      name = FormatStreetAndHouse(addr);
+
+    address = rankerResult.GetRegionName(m_infoGetter, type);
+    // Format full address only for suitable results.
+    if (ftypes::IsAddressObjectChecker::Instance()(rankerResult.GetTypes()))
```
ygorshenin: So, there can be cases when addr is unused. That's why I proposed you to implement a lazy wrapper around addr.

search/ranker.cpp
mpimenov: done

search/ranker.cpp
mpimenov: done

search/ranker.cpp
mpimenov: Let's remove it.
done

search/ranker.cpp
```diff
@@ -329,20 +353,62 @@ void Ranker::Init(Params const & params, Geocoder::Params const & geocoderParams
...
+
+  string address;
+  if (needAddress)
+  {
+    LazyAddressGetter addressGetter(m_reverseGeocoder, rankerResult.GetCenter());
+
+    // Insert exact address (street and house number) instead of empty result name.
+    if (name.empty())
+    {
+      auto const addr = addressGetter.GetAddress();
```
ygorshenin: Const ref?

mpimenov: done

PR #7433, opened 2017-10-31
[search] Refactored the lang indexes in the KeywordLangMatcher.

mpimenov:
The efficiency should drop a little because now new vectors
are created all the time but, taking into account how small
a part of processing a query this is, I think that introducing
additional lighter methods is not justified.

### DIFF
--- a/search/keyword_lang_matcher.hpp
+++ b/search/keyword_lang_matcher.hpp
@@ -2,6 +2,8 @@
 
 #include "search/keyword_matcher.hpp"
 
+#include "base/string_utils.hpp"
+
 #include <array>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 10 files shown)


search/keyword_lang_matcher.hpp
```diff
@@ -1,20 +1,23 @@
...
+#include <utility>
+#include <vector>
 
 namespace search
 {
-
 class KeywordLangMatcher
 {
 public:
+  using StringT = KeywordMatcher::StringT;
```
ygorshenin: Is it possible to get rid of "T" suffixes?

mpimenov: done

search/keyword_matcher.cpp
```diff
@@ -99,13 +98,17 @@ KeywordMatcher::ScoreT KeywordMatcher::Score(StringT const * tokens, size_t coun
...
 
-KeywordMatcher::ScoreT::ScoreT()
-  : m_sumTokenMatchDistance(0), m_nameTokensMatched(0), m_nameTokensLength(0),
-    m_numQueryTokensAndPrefixMatched(0), m_bFullQueryMatched(false), m_bPrefixMatched(false)
+KeywordMatcher::Score::Score()
+  : m_sumTokenMatchDistance(0)
+  , m_nameTokensMatched(0)
+  , m_nameTokensLength(0)
+  , m_numQueryTokensAndPrefixMatched(0)
+  , m_bFullQueryMatched(false)
```
ygorshenin: How about to get rid of this "b" prefixes?

mpimenov: done

PR #7439, opened 2017-10-31
Small EdgeEstimator refactoring.

mgsergio:
@bykoianko @tatiana-kondakova @ygorshenin

### DIFF
--- a/routing/edge_estimator.hpp
+++ b/routing/edge_estimator.hpp
@@ -21,17 +21,16 @@ namespace routing
 class EdgeEstimator
 {
 public:
-  explicit EdgeEstimator(double maxSpeedKMpH);
   virtual ~EdgeEstimator() = default;
 
-  double CalcHeuristic(m2::PointD const & from, m2::PointD const & to) const;
+  virtual double CalcHeuristic(m2::PointD const & from, m2::PointD const & to) const = 0;
   // Returns time in seconds it takes to go from point |from| to point |to| along a leap (fake)
   // edge |from|-|to|.
   // Note 1. The result of the method should be used if it's necessary to add a leap (fake) edge
   // (|from|, |to|) in road graph.
   // Note 2. The result of the method should be less or equal to CalcHeuristic(|from|, |to|).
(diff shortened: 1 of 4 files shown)


tatiana-kondakova:
Планируется ли, что у EdgeEstimator появится наследник, который не будет при этом наследником TimedEdgeEstimator?

routing/edge_estimator.hpp
```diff
@@ -21,17 +21,16 @@ namespace routing
 class EdgeEstimator
 {
 public:
-  explicit EdgeEstimator(double maxSpeedKMpH);
   virtual ~EdgeEstimator() = default;
 
-  double CalcHeuristic(m2::PointD const & from, m2::PointD const & to) const;
+  virtual double CalcHeuristic(m2::PointD const & from, m2::PointD const & to) const = 0;
   // Returns time in seconds it takes to go from point |from| to point |to| along a leap (fake)
```
ygorshenin: The comment is obsolete now, and must be moved to TimedEdgeEstimator.

routing/edge_estimator.hpp
```diff
@@ -21,17 +21,16 @@ namespace routing
 class EdgeEstimator
 {
 public:
-  explicit EdgeEstimator(double maxSpeedKMpH);
   virtual ~EdgeEstimator() = default;
 
-  double CalcHeuristic(m2::PointD const & from, m2::PointD const & to) const;
+  virtual double CalcHeuristic(m2::PointD const & from, m2::PointD const & to) const = 0;
```
ygorshenin: BTW, why do we need one more class? If you're going to override EdgeEstimator, just mark CalcHeuristic and CalcLeapWeight as virtual, and override them in your subclass. Moreover, please, mark current subclasses as final. Hope, this will somehow reduce the cost of virtual methods, because these methods are on the hot path in routing.

tatiana-kondakova: +1

mgsergio:
You shall not pass!

PR #7448, opened 2017-11-01
[search] Fixed dancing search results.

ygorshenin:


map/framework.cpp
```diff
@@ -1326,16 +1328,15 @@ void Framework::SetCurrentCountryChangedListener(TCurrentCountryChanged const &
...
 
-void Framework::PokeSearchInViewport()
+void Framework::PokeSearchInViewport(bool forceSearch)
 {
   if (!m_isViewportInitialized || !IsViewportSearchActive())
     return;
 
-  auto & intent = m_searchIntents[static_cast<size_t>(search::Mode::Viewport)];
-  auto & params = intent.m_params;
+  auto params = m_searchIntents[static_cast<size_t>(search::Mode::Viewport)].m_params;
```
mpimenov: I guess we need the comment about the intentional copy back.

map/framework.hpp
```diff
@@ -557,7 +557,7 @@ class Framework : public search::ViewportSearchCallback::Delegate,
 
   // When search in viewport is active or delayed, restarts search in
   // viewport.
-  void PokeSearchInViewport();
+  void PokeSearchInViewport(bool forceSearch = true);
```
mpimenov: Add a comment to the parameter, please.

map/framework.cpp
ygorshenin: Done

map/framework.hpp
ygorshenin: Done

PR #7459, opened 2017-11-02
Fixing and updating some pt_BR translations

naoliv:


data/categories.txt
```diff
@@ -1816,6 +1816,7 @@ de:^Bahnübergang
 it:^Passaggio a livello
 es:^Paso a nivel
 pt:^Passagem de nível
+pt-BR: ^Passagem em nível
```
mpimenov: Please remove the leading space.

naoliv: Fixed.

data/categories.txt
```diff
@@ -8816,6 +8848,7 @@ nb:^Billigbutikk
 nl:^Bazaar
 pl:^Sklep z różnościami
 pt:^Loja dos 300
+pt-BR:^1,99|Loja de variedades
```
mpimenov: nit: "1,99" will not work as you expect in the current version (nor in the near future)
We'll most likely have to remove it.

naoliv: Because it starts with a number? If it gets updated later to "Loja de 1,99", will it be valid then?

mpimenov: Yes. And also because we change commas and some other symbols to spaces during simplification.
"Loja de 1,99" is better but still may have issues.

mpimenov:
@naoliv We've merged this but please note it may take some time to see it in the app because we don't fully support pt-BR now.

data/categories.txt
naoliv: Will a dot be better than a comma? "Loja de 1.99", for example.

mpimenov: The full list is here: https://github.com/mapsme/omim/blob/master/indexer/search_delimiters.cpp

The rule of thumb is that most non-alphanumeric tokens are transformed to spaces.
Let's keep it as is for now.

naoliv: Right, thanks!

PR #7460, opened 2017-11-03
[transit] Optimization. Creating world graph once for building mwm transit section.

bykoianko:
Пару недель назад, при написании кода поиска ближайшего (лучшего) пешеходного сегмента к гейту мы спрятали создание WorldGraph в метод: IndexRouter::FindBestSegmentInSingleMwm(). Это оказалось не эффективно, поскольку для каждого Gate-та при создании секции transit создавался WorldGraph. Я передал код так, чтоб WorldGraph создавался один раз, на сборку секции transit.

@tatiana-kondakova @ygorshenin PTAL

### DIFF
--- a/routing/index_router.cpp
+++ b/routing/index_router.cpp
@@ -318,13 +318,20 @@ IndexRouter::IndexRouter(VehicleType vehicleType, bool loadAltitudes,
   CHECK(m_directionsEngine, ());
 }
 
-bool IndexRouter::FindBestSegmentInSingleMwm(m2::PointD const & point, m2::PointD const & direction,
-                                             bool isOutgoing, Segment & bestSegment)
+std::unique_ptr<WorldGraph> IndexRouter::MakeSingleMwmWorldGraphForGenerator()
 {
   auto worldGraph = MakeWorldGraph();
   worldGraph->SetMode(WorldGraph::Mode::SingleMwm);
+  return worldGraph;
+}
+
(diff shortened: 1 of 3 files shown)


routing/index_router.cpp
```diff
@@ -318,13 +318,19 @@ IndexRouter::IndexRouter(VehicleType vehicleType, bool loadAltitudes,
   CHECK(m_directionsEngine, ());
 }
 
-bool IndexRouter::FindBestSegmentInSingleMwm(m2::PointD const & point, m2::PointD const & direction,
-                                             bool isOutgoing, Segment & bestSegment)
+std::unique_ptr<WorldGraph> IndexRouter::MakeSingleMwmWorldGraphForGenerator()
```
ygorshenin: Maybe it's worth to remove ```ForGenerator``` suffixes?

bykoianko: Maybe. But in that case we show explicitly that we uncover WorldGraph for generator only. Beside that we write suffix `ForTesting`. Do you think it's better to remove `ForGenerator` anyway?

ygorshenin: ForTesting is used for methods that expose somehow internal structure of the class, or methods that are doing something that is very dangerous in prod, but ok in tests.
 
These methods looks good to me, not sure that it's worth to specify that they must be used in generator only. Up to you.

tatiana-kondakova: s/std::unique_ptr/unique_ptr

bykoianko: Let's remove the suffix.

PR #7463, opened 2017-11-03
Android build fix: add explicit initialization for default constructed members of TransitInfo

tatiana-kondakova:


tatiana-kondakova:
DUDE JTTP

PR #7486, opened 2017-11-08
[search] Tracer.

ygorshenin:
Tracer for search, collects all valid parses of the search query.

### DIFF
--- a/search/tracer.cpp
+++ b/search/tracer.cpp
@@ -10,7 +10,7 @@ using namespace std;
 namespace search
 {
 // Tracer::Parse -----------------------------------------------------------------------------------
-Tracer::Parse::Parse(std::vector<TokenType> const & types)
+Tracer::Parse::Parse(vector<TokenType> const & types)
 {
   size_t i = 0;
   while (i != types.size())
@@ -30,7 +30,7 @@ Tracer::Parse::Parse(vector<pair<TokenType, TokenRange>> const & ranges)
     m_ranges[kv.first] = kv.second;
 }
 
-std::string DebugPrint(Tracer::Parse const & parse)
+string DebugPrint(Tracer::Parse const & parse)
 {
   using TokenType = Tracer::Parse::TokenType;
 
--- a/search/tracer.cpp
+++ b/search/tracer.cpp
@@ -1,0 +1,65 @@
+#include "search/tracer.hpp"
+
+#include "base/stl_helpers.hpp"
+
+#include <cstddef>
+#include <sstream>
+
+using namespace std;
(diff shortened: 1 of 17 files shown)


mpimenov:
(discussed offline)
1. Only successful parses are emitted now. This is probably right, and emitting
the incomplete parses that is currently in the works will cover other cases.
EmitResult should still be the only point in the geocoder that emits results (and parses).
The search by lat-lon does not fit in here, but it is not important as the parses are trivial.
2. We probably should add a LogTracer that is easy to turn on and that would
print the parses to the console for debug purposes.

search/tracer.hpp
```diff
@@ -0,0 +1,42 @@
...
+{
+class Tracer
+{
+public:
+  struct Parse
+  {
+    using TokenType = BaseContext::TokenType;
+
+    explicit Parse(std::vector<TokenType> const & types);
+    explicit Parse(std::vector<std::pair<TokenType, TokenRange>> const & ranges);
```
tatiana-kondakova: do we really need explicit here?

tatiana-kondakova: oops, sorry

search/tracer.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <cstddef>
+#include <sstream>
+
+using namespace std;
+
+namespace search
+{
+// Tracer::Parse -----------------------------------------------------------------------------------
+Tracer::Parse::Parse(std::vector<TokenType> const & types)
```
tatiana-kondakova: `s/std::vector/vector`

ygorshenin: Done

ygorshenin:
After discussion with @burivuh we concluded that it would be useful to emit matched categories for the innermost layer. IOW, for search request "cafe moscow" there can be two parses - ["poi", "city"] and ["poi"] (for the cafe named "Moscow"). In the former case it would be nice to somehow specify that this was not just "poi", but also a category "amenity-cafe".

ygorshenin:
After discussion with @burivuh we concluded that it would be useful to emit matched categories for the innermost layer. IOW, for search request "cafe moscow" there can be two parses - ["poi", "city"] and ["poi"] (for the cafe named "Moscow"). In the former case it would be nice to somehow specify that this was not just "poi", but also a category "amenity-cafe".

ygorshenin:
@burivuh - addition of the list of categories is not a big deal, and if you agree with the generic structure of the tracer - feel free to merge this PR, as @mpimenov works on geocoder concurrently.

PR #7496, opened 2017-11-09
CMake Designer Tool 

greshilov:
Usage:
```
cmake {your_omim_path} -DBUILD_DESIGNER:bool=True -DAPP_VERSION={app_version}
sudo make package -j8
```
Necessity of sudo is explained by read-only permissions on qt library directories. There is also may be some problems with copying tools/shaders_compiler.


greshilov:
PTAL @rokuz @ygorshenin

PR #7583, opened 2017-11-17
[search] Flattened the call graph of UpdateResults and Finish.

mpimenov:
Previously, Processor called Emitter directly to set
the end marker, but all other calls to Emitter were made
through the Geocoder-Preranker-Ranker pipeline.

Now all calls must be made through the pipeline, which
at least makes in easier to update the state in case
of cancellation (which is only done in PreRanker's m_currEmit currently).

Also, removed an unneeded call of UpdateResults in Processor:
all UpdateResults calls from Processor must now be made indirectly,
either via the normal search process in Geocoder, or via calling
Finish.

### DIFF
--- a/search/geocoder.hpp
+++ b/search/geocoder.hpp
@@ -101,6 +101,19 @@ public:
   void GoEverywhere();
   void GoInViewport();
 
+  // Ends geocoding and informs the following stages
+  // of the pipeline (PreRanker).
+  // This method must be called from the previous stage
+  // of the pipeline (the Processor).
+  // If |cancelled| is true, the reason for calling Finish must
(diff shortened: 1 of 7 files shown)


search/geocoder.hpp
```diff
@@ -101,6 +101,15 @@ class Geocoder
   void GoEverywhere();
   void GoInViewport();
 
+  // Ends geocoding and informs the following stages of the pipeline
+  // (Preranker, Ranker and Emitter).
```
ygorshenin: Please, omit Ranker and Emitter. Geocoder should know about PreRanker only.
s/Preranker/PreRanker/

mpimenov: done

PR #7602, opened 2017-11-20
[search] Fixed an uncaught CancelException.

mpimenov:
https://jira.mail.ru/browse/MAPSME-6018

### DIFF
--- a/search/ranker.cpp
+++ b/search/ranker.cpp
@@ -420,7 +420,8 @@ void Ranker::SuggestStrings()
 
 void Ranker::UpdateResults(bool lastUpdate)
 {
-  BailIfCancelled();
+  if (!lastUpdate)
+    BailIfCancelled();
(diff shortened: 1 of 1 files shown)


mpimenov:
This PR restores the rules of the contract at https://github.com/mapsme/omim/blob/master/search/geocoder.hpp#L112

We need a way to shut down the search (probably due to a cancellation) without throwing exceptions, and PreRanker's UpdateResults stands in the way (by calling heavy Ranker's UpdateResults).

ygorshenin:
Hm, I dislike the idea that some knowledge is spread among several places in the code. LGTM for a temporary fix, but we need to fix this logic.

mpimenov:
I don't like it either. The problem seems to be in the interplay of ```lastUpdate``` and ```isCancelled```.

PR #7642, opened 2017-11-24
[search] Relaxed parsing in Geocoder.

mpimenov:


ygorshenin:
I understood the approach taken here, but dislike the idea of code/idea-duplication here. Is it possible instead to somehow count number of emit results in BaseContext, and, if you need to check that some action caused emission of results, you may ask BaseContext. For example:
```c++
auto const numEmitted = ctx.GetNumEmitted();
... complex actions ...
if (numEmitted != ctx.GetNumEmitted())
  // Okay, we can do something here.
```
?
This will allow you to get rid of these size_t return values.

mpimenov:
Yes, that was another possible approach.
We lost const in FindPaths and EmitResult but overall it now looks better to me.

PTAL

search/search_quality/scoring_model.py
```diff
@@ -37,6 +37,7 @@ def normalize_data(data):
     data['DistanceToPivot'] = data['DistanceToPivot'].apply(transform_distance)
     data['Rank'] = data['Rank'].apply(lambda v: v / MAX_RANK)
     data['Relevance'] = data['Relevance'].apply(lambda v: RELEVANCES[v])
+    data['AllTokensUsed'] = data['AllTokensUsed'].apply(lambda v : +1 if bool(v) else -1)
```
ygorshenin: Please, use dummy-coding, i.e. 0 and 1.

mpimenov: done

PR #7715, opened 2017-12-05
Fixed transfer marker rotation.

darina:


drape_frontend/shaders/route_marker.vsh.glsl
```diff
@@ -6,6 +6,7 @@ uniform mat4 modelView;
 uniform mat4 projection;
 uniform mat4 pivotTransform;
 
+uniform vec2 u_angle;
```
rokuz: u_cosSinAngle

darina:
@rokuz PTAL

PR #7786, opened 2017-12-12
Transit routing in case of neighboring mwm without transit section.

bykoianko:
cherry-pick https://github.com/mapsme/omim/pull/7776 to master


bykoianko:
@tatiana-kondakova PTAL

PR #7797, opened 2017-12-13
[OpenLR-tool] Mark matched path as golden.

mgsergio:
@ygorshenin

### DIFF
--- a/openlr/openlr_match_quality/openlr_assessment_tool/traffic_mode.cpp
+++ b/openlr/openlr_match_quality/openlr_assessment_tool/traffic_mode.cpp
@@ -268,29 +268,33 @@ Qt::ItemFlags TrafficMode::flags(QModelIndex const & index) const
   return QAbstractItemModel::flags(index);
 }
 
+void TrafficMode::GoldifyMatchedPath()
+{
+  if (!m_currentSegment->HasMatchedPath())
+  {
+    QMessageBox::information(nullptr /* parent */, "Error",
+                             "The selected segment does not have a matched path");
+    return;
+  }
+
+  if (!StartBuildingPathChecks())
+    return;
+
+  m_currentSegment->SetGoldenPath(m_currentSegment->GetMatchedPath());
+  m_goldenPath.clear();
+  m_drawerDelegate->DrawGoldenPath(GetPoints(m_currentSegment->GetGoldenPath()));
+}
+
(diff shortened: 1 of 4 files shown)


openlr/openlr_match_quality/openlr_assessment_tool/traffic_mode.cpp
```diff
@@ -268,29 +268,32 @@ Qt::ItemFlags TrafficMode::flags(QModelIndex const & index) const
...
-  if (m_currentSegment->HasGoldenPath())
+  if (!m_currentSegment->HasMatchedPath())
   {
-    auto const btn = QMessageBox::question(
-        nullptr,
-        "Override warning",
-        "The selected segment already has a golden path. Do you want to override?");
-    if (btn == QMessageBox::No)
-      return;
+    QMessageBox::information(nullptr, "Error", "The selected segment does not have a matched path");
```
ygorshenin: Comment to the nullptr, please.

openlr/openlr_match_quality/openlr_assessment_tool/traffic_mode.cpp
```diff
@@ -502,4 +505,24 @@ void TrafficMode::HandlePoint(m2::PointD clickPoint, Qt::MouseButton const butto
     return;
   }
 }
+
+bool TrafficMode::StartBuildingPathChecks() const
+{
+  CHECK(m_currentSegment, ("A segment should be selected before path building is started."));
+
+  if (m_buildingPath)
```
ygorshenin: Could you please explain why do you mix CHECKs with exceptions? What's the reason to use exceptions here?

openlr/openlr_match_quality/openlr_assessment_tool/traffic_mode.cpp
mgsergio: Done.

openlr/openlr_match_quality/openlr_assessment_tool/traffic_mode.cpp
mgsergio: Exception may be handled by ui. The situation that leads to this exception is triggered by a user.
While CHECK(m_currentSegment, (...)) is for catching logical error.

ygorshenin: Okay.

PR #7831, opened 2017-12-15
[ads] search banner is removed

milchakov:


ygorshenin:
Am I right that search banners are completely removed from search?

rokuz:
Да, пока удалены насовсем

burivuh:
This is just a mapsme google banner rage quit )

PR #7858, opened 2017-12-20
[ios] Fixed viator crash with japanese locale and removed extended ci…

VladiMihaylenko:
…an place page.


VladiMihaylenko:
@milchakov ptal

PR #7866, opened 2017-12-20
[search] Count errors in the city part.

ygorshenin:
We decided to take into account number of typos made in the city part of an address.

### DIFF
--- a/search/model.hpp
+++ b/search/model.hpp
@@ -47,6 +47,11 @@ public:
     TYPE_COUNT
   };
 
+  static bool IsLocalityType(Type const type)
+  {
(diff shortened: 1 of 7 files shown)


search/model.hpp
```diff
@@ -47,6 +47,11 @@ class Model
     TYPE_COUNT
   };
 
+  static bool IsLocalityType(Type const type)
+  {
+    return type >= TYPE_VILLAGE && type <= TYPE_COUNTRY;
```
mpimenov: nit: add a note about the dangers of reordering in the Type enum

ygorshenin: Done

ygorshenin:
@tatiana-kondakova PTAL

PR #7874, opened 2017-12-22
[storage] Fixed diff applying on startup.

VladiMihaylenko:


storage/diff_scheme/diff_manager.cpp
```diff
@@ -107,17 +107,24 @@ Status Manager::GetStatus() const
...
 
-void Manager::SetStatus(Status status)
+bool Manager::SizeFor(storage::TCountryId const & countryId, uint64_t & size) const
 {
   std::lock_guard<std::mutex> lock(m_mutex);
-  m_status = status;
+  if (!HasDiffForUnsafe(countryId))
+    return false;
+
+  size = m_diffs.at(countryId).m_size;
```
ygorshenin:
```c++
size = m_diffs[countryId].m_size;
return true;
```

storage/diff_scheme/diff_manager.cpp
```diff
@@ -107,17 +107,24 @@ Status Manager::GetStatus() const
...
+  return true;
 }
 
-FileInfo const & Manager::InfoFor(storage::TCountryId const & countryId) const
+bool Manager::VersionFor(storage::TCountryId const & countryId, uint64_t & version) const
 {
   std::lock_guard<std::mutex> lock(m_mutex);
-  ASSERT(HasDiffForUnsafe(countryId), ());
-  return m_diffs.at(countryId);
+  if (!HasDiffForUnsafe(countryId))
```
ygorshenin:
```c++
template <typename Fn>
bool WithDiff(string const & countryId, Fn && fn)
{
  std::lock_guard<std::mutex> lock(m_mutex);
  auto it = m_diffs.find(countryId);
  if (it == m_diffs.end())
    return false;
  fn(it->second);
  return true;
}
```

```
bool Manager::SizeFor(...)

{
  return WithDiff([&](Diff const & diff) { size = diff.m_size; });
}

bool Manager::VersionFor(...)
{
  return WithDiff(countryId, [&](Diff & diff) { version = diff.m_version; });
}
```

storage/storage.cpp
```diff
@@ -614,14 +614,28 @@ void Storage::DownloadNextFile(QueuedCountry const & country)
...
+  // Since a downloaded valid diff file may be either with .diff or .diff.ready extension,
+  // we have to check these both cases in order to find
+  // the diff file which is ready to apply.
+  // If there is a such file we have to cause the success download scenario.
+  bool isDownloadedDiff = false;
+  if (opt == MapOptions::Diff)
+  {
+    string filePath = readyFilePath;
+    my::GetNameWithoutExt(filePath);
+    isDownloadedDiff = p.GetFileSizeByFullPath(filePath, size);
```
ygorshenin: Implicit convert from integral to boolean? What about empty files?

storage/storage.cpp
```diff
@@ -614,14 +614,28 @@ void Storage::DownloadNextFile(QueuedCountry const & country)
...
+  }
 
   // It may happen that the file already was downloaded, so there're
   // no need to request servers list and download file.  Let's
   // switch to next file.
-  if (GetPlatform().GetFileSizeByFullPath(filePath, size))
+  if (p.GetFileSizeByFullPath(readyFilePath, size) || isDownloadedDiff)
```
ygorshenin: Please, do not implicitly convert integrals to bools. Why not to use function, like, IsFileExistsByFullPath, or something like that?

storage/storage.cpp
VladiMihaylenko: I didn't get where you saw that. GetFileSizeByFullPath returns bool.

storage/storage.cpp
```diff
@@ -1461,25 +1494,8 @@ void Storage::OnDiffStatusReceived(diffs::Status const status)
...
-        TLocalFilePtr localFile = GetLocalFile(countryId, GetCurrentDataVersion());
-        ASSERT(localFile, ());
-        DeleteCountryIndexes(*localFile);
-        m_didDownload(countryId, localFile);
-        NotifyStatusChangedForHierarchy(countryId);
-      };
-
       if (m_diffManager.HasDiffFor(countryId))
-        ApplyDiff(countryId, registerDiffFn);
+        UpdateNode(countryId);
```
ygorshenin: Could you please comment what was wrong with the old code here?

storage/storage.cpp
ygorshenin: Ah, see now, sorry.

storage/storage.cpp
ygorshenin: OK, I get it. But would it be safer to swap operands here?

VladiMihaylenko: Because GetFileSizeByFullPath returns bool and we also need this size value. 
One thing that I should do it's checking isDownloadedDiff at first.

storage/storage.cpp
VladiMihaylenko: This was a situation when we found a valid diff and started to apply it (applying happens on the diff manager worker thread). At the same time, a user could initiate updating the same mwm by using the downloader. So at the point when applying is finished (after calling of ApplyDiff here), the downloader could initiate new applying because it had just downloaded the same diff file. But an old mwm which is necessary for applying was already deleted in the previous applying so it was CHECK for this situation.
With new approach, user will just see a regular situation as he could see before he closed the app: the mwm is being updated and it can't be canceled because it's already in applying state and in another hand the user can't start new updating/downloading of the same mwm, because it's being updated right now.

storage/diff_scheme/diff_manager.cpp
VladiMihaylenko: Do you think it's a good idea not to check the status of diff manager?  Because HasDiffForUnsafe does this check as well. Also, we still need HasDiffFor as a public method, which uses HasDiffForUnsafe. So it seems it will be more redundant lines with that approach.

PR #7909, opened 2018-01-09
Fix configure.sh file

greshilov:


configure.sh
```diff
@@ -12,7 +12,7 @@ PRIVATE_PUSHWOOSH_PROPERTIES="$BASE_PATH/android/pushwoosh.properties"
 SAVED_PRIVATE_REPO_FILE="$BASE_PATH/.private_repository_url"
 TMP_REPO_DIR="$BASE_PATH/.tmp.private.repo"
 
-if [ ! -f "$BASE_PATH/omim.pro" ]; then
+if [ "$(git rev-parse --show-toplevel)" != $PWD ]; then
```
ygorshenin:
```bash
"$PWD"
```

greshilov:
@Zverik @rokuz PTAL

configure.sh
greshilov: Sure, thanks!

PR #7923, opened 2018-01-11
[drape] Fixed linux build.

ygorshenin:


ygorshenin:
There's a call to ```std::remove_if```, but ```<algorithms>``` is missed.

PR #7934, opened 2018-01-15
[build] Fixed ASAN build.

ygorshenin:
This commit moves ASAN/TSAN compile options before any 3party
subdirectories, to prevent build of std library with and without
instrumentation under the same hood.


map/framework.cpp
```diff
@@ -177,13 +177,6 @@ string MakeSearchBookingUrl(booking::Api const & bookingApi, search::CityFinder
 
   return bookingApi.GetSearchUrl(city, name);
 }
-
-bool IsCianMode(string query)
```
mpimenov: Why in this PR?

mpimenov: Ah, it should not even be there, so I'm ok with it.

ygorshenin:
Could anybody merge this?

PR #7954, opened 2018-01-17
[types] Add leisure=resort

Zverik:
Простой фикс, добавляющий на карту 3700 прибрежных гостиниц, отмеченных правильным тегом `leisure=resort`. У нас, почему-то, был неправильный `tourism=resort`.

### DIFF
--- a/data/types.txt
+++ b/data/types.txt
@@ -1199,3 +1199,4 @@ tourism|theme_park
 tourism|wilderness_hut
 sponsored|cian
 sponsored|holiday
+tourism|resort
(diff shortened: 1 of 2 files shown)


data/types.txt
```diff
@@ -1199,3 +1199,4 @@ tourism|theme_park
 tourism|wilderness_hut
 sponsored|cian
 sponsored|holiday
+tourism|resort
```
mpimenov: put this under tourism|wilderness_hut, please

Zverik: What do you mean? This file is auto-generated. We cannot reorder types, since that would break all older mwm files.

mpimenov: Are both files in this PR auto-generated?

Zverik: No. I added `leisure|resort` to the `mapcss-mapping.csv` and run `generate_drules.sh` to update `types.txt`. We can add new types only to the end of the types list, since there are no holes to fill.

mpimenov: Ok. Are there any other "tourism-" types that we can add right now?

PR #7961, opened 2018-01-18
Fixing warnings in 3party lib.

bykoianko:
@ygorshenin @mpimenov @greshilov  PTAL

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,12 +331,6 @@ endfunction()
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-add_compile_options(
-  "-Wall"
-)
-
-add_clang_compile_options("-Wshorten-64-to-32")
-
 if (USE_ASAN)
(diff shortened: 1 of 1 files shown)


bykoianko:
@ygorshenin PTAL

CMakeLists.txt
```diff
@@ -368,6 +362,16 @@ add_subdirectory(3party/sdf_image)
 add_subdirectory(3party/stb_image)
 add_subdirectory(3party/succinct)
 
+if (PLATFORM_DESKTOP)
+  add_subdirectory(3party/libtess2)
+endif()
+
+add_compile_options(
```
ygorshenin: Please, add a comment that only options related to warnings can be placed here, between 3party and our code. Other options must be set before all ```add_subdirectory``` call.

bykoianko: Done

bykoianko:
@ygorshenin PTAL

PR #7987, opened 2018-01-23
[editor] Support migration for old versions of edits.

milchakov:
https://jira.mail.ru/browse/MAPSME-6593

### DIFF
--- a/editor/feature_matcher.cpp
+++ b/editor/feature_matcher.cpp
@@ -2,6 +2,7 @@
 
 #include "base/logging.hpp"
 #include "base/stl_helpers.hpp"
+#include "base/stl_iterator.hpp"
 
 #include "std/algorithm.hpp"
 #include "std/function.hpp"
@@ -109,8 +110,9 @@ double MatchByGeometry(LGeometry const & lhs, RGeometry const & rhs)
 MultiPolygon TrianglesToPolygon(vector<m2::PointD> const & points)
 {
   size_t const kTriangleSize = 3;
-  CHECK_EQUAL(points.size() % kTriangleSize, 0, ());
-  CHECK(!points.empty(), ());
+  if (points.size() % kTriangleSize != 0 || points.empty())
+    MYTHROW(matcher::NotAPolygonException, ("Count of points must be multiple of", kTriangleSize));
+
   vector<MultiPolygon> polygons;
   for (size_t i = 0; i < points.size(); i += kTriangleSize)
   {
@@ -121,11 +123,14 @@ MultiPolygon TrianglesToPolygon(vector<m2::PointD> const & points)
     for (size_t j = i; j < i + kTriangleSize; ++j)
       outer.push_back(PointXY(points[j].x, points[j].y));
     bg::correct(p);
-    ASSERT(bg::is_valid(polygon), ());
+    if (!bg::is_valid(polygon))
+      MYTHROW(matcher::NotAPolygonException, ("The triangle is not valid"));
     polygons.push_back(polygon);
   }
 
-  CHECK(!polygons.empty(), ());
+  if (polygons.empty())
+    return {};
+
   auto & result = polygons[0];
   for (size_t i = 1; i < polygons.size(); ++i)
   {
@@ -344,5 +349,29 @@ double ScoreTriangulatedGeometries(vector<m2::PointD> const & lhs, vector<m2::Po
     return kPenaltyScore;
 
   return MatchByGeometry(lhsPolygon, rhsPolygon);
+}
+
+double ScoreTriangulatedGeometriesByPoints(vector<m2::PointD> const & lhs,
+                                           vector<m2::PointD> const & rhs)
+{
+  // The default comparison operator used in sort above (cmp1) and one that is
+  // used in set_itersection (cmp2) are compatible in that sence that
+  // cmp2(a, b) :- cmp1(a, b) and
+  // cmp1(a, b) :- cmp2(a, b) || a almost equal b.
+  // You can think of cmp2 as !(a >= b).
+  // But cmp2 is not transitive:
+  // i.e. !cmp(a, b) && !cmp(b, c) does NOT implies !cmp(a, c),
+  // |a, b| < eps, |b, c| < eps.
+  // This could lead to unexpected results in set_itersection (with greedy implementation),
+  // but we assume such situation is very unlikely.
+  auto const matched = set_intersection(begin(lhs), end(lhs),
+                                        begin(rhs), end(rhs),
+                                        CounterIterator(),
+                                        [](m2::PointD const & p1, m2::PointD const & p2)
+                                        {
+                                          return p1 < p2 && !p1.EqualDxDy(p2, 1e-7);
+                                        }).GetCount();
+
+  return static_cast<double>(matched) / lhs.size();
 }
 }  // namespace matcher
--- a/editor/feature_matcher.cpp
+++ b/editor/feature_matcher.cpp
@@ -14,6 +14,8 @@
 #include <boost/geometry/geometries/point_xy.hpp>
 #include <boost/geometry/geometries/polygon.hpp>
 
+using namespace std;
+
 using editor::XMLFeature;
 
 namespace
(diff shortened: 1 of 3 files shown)


editor/feature_matcher.cpp
```diff
@@ -109,8 +110,9 @@ double MatchByGeometry(LGeometry const & lhs, RGeometry const & rhs)
 MultiPolygon TrianglesToPolygon(vector<m2::PointD> const & points)
 {
   size_t const kTriangleSize = 3;
-  CHECK_EQUAL(points.size() % kTriangleSize, 0, ());
-  CHECK(!points.empty(), ());
+  if (points.size() % kTriangleSize != 0 || points.empty())
+    MYTHROW(matcher::NotAPolygonException, ("Count of points must be multiple of", kTriangleSize));
```
mgsergio: Strictly speaking, zero is a multiple of anything but zero itself.

editor/feature_matcher.hpp
```diff
@@ -5,16 +5,24 @@
 #include "geometry/mercator.hpp"
 #include "geometry/point2d.hpp"
 
+#include "base/exception.hpp"
+
 #include "std/vector.hpp"
```
mgsergio: #include <vector>

editor/feature_matcher.hpp
```diff
@@ -5,16 +5,24 @@
...
+/// Throws NotAPolygon exception when |geometry| is not convertible to a polygon.
 pugi::xml_node GetBestOsmWayOrRelation(pugi::xml_document const & osmResponse,
                                        vector<m2::PointD> const & geometry);
 /// Returns value form [-1, 1]. Negative values are used as penalty, positive as score.
-/// |lhs| and |rhs| - triangulated polygons;
+/// |lhs| and |rhs| - triangulated polygons.
+/// Throws NotAPolygon exception when lhs or rhs is not convertible to a polygon.
 double ScoreTriangulatedGeometries(vector<m2::PointD> const & lhs, vector<m2::PointD> const & rhs);
+/// Deprecated, use ScoreTriangulatedGeometries instead.
+double ScoreTriangulatedGeometriesByPoints(vector<m2::PointD> const & lhs,
```
mgsergio: Why should we keep it?

indexer/edits_migration.cpp
```diff
@@ -69,7 +68,19 @@ FeatureID MigrateWayorRelatonFeatureIndex(
...
-        auto const score = matcher::ScoreTriangulatedGeometries(geometry, ftGeometry);
+        double score = 0.0;
+        try
+        {
+          score = matcher::ScoreTriangulatedGeometries(geometry, ftGeometry);
+        }
+        catch (matcher::NotAPolygonException & ex)
+        {
+          // Support migration for old application versions.
+          // TODO(a): To remove it after some time.
```
mgsergio: What time exactly? If I work with this code after "some time" how can I tell it is the right time to remove it?

editor/feature_matcher.cpp
```diff
@@ -345,4 +350,28 @@ double ScoreTriangulatedGeometries(vector<m2::PointD> const & lhs, vector<m2::Po
...
+  // used in set_itersection (cmp2) are compatible in that sence that
+  // cmp2(a, b) :- cmp1(a, b) and
+  // cmp1(a, b) :- cmp2(a, b) || a almost equal b.
+  // You can think of cmp2 as !(a >= b).
+  // But cmp2 is not transitive:
+  // i.e. !cmp(a, b) && !cmp(b, c) does NOT implies !cmp(a, c),
+  // |a, b| < eps, |b, c| < eps.
+  // This could lead to unexpected results in set_itersection (with greedy implementation),
+  // but we assume such situation is very unlikely.
+  auto const matched = set_intersection(begin(lhs), end(lhs),
```
mgsergio: This code looks very familiar to me. I see an addition of it, but where is the removing?

editor/feature_matcher.cpp
milchakov: убрал || points.empty(), т.к. такой кейс должен обрабатываться как пустой полигон.

editor/feature_matcher.cpp
milchakov: обсудили голосом.

editor/feature_matcher.hpp
milchakov: готово

indexer/edits_migration.cpp
milchakov: заменил на
To remove it when version 8.0.x will be not used.

editor/feature_matcher.hpp
milchakov: /// Deprecated, use ScoreTriangulatedGeometries instead. In old versions duplicated points were removed
/// from list of points, but it not possible to make polygon by using this list. We need to keep
/// this method to support migration from old application vesions (8.0.x and below).

milchakov:
@mgsergio PTAL

editor/feature_matcher.hpp
```diff
@@ -5,16 +5,28 @@
...
-                                       vector<m2::PointD> const & geometry);
+                                       std::vector<m2::PointD> const & geometry);
 /// Returns value form [-1, 1]. Negative values are used as penalty, positive as score.
-/// |lhs| and |rhs| - triangulated polygons;
-double ScoreTriangulatedGeometries(vector<m2::PointD> const & lhs, vector<m2::PointD> const & rhs);
+/// |lhs| and |rhs| - triangulated polygons.
+/// Throws NotAPolygon exception when lhs or rhs is not convertible to a polygon.
+double ScoreTriangulatedGeometries(std::vector<m2::PointD> const & lhs,
+                                   std::vector<m2::PointD> const & rhs);
+/// Deprecated, use ScoreTriangulatedGeometries instead. In old versions duplicated points were removed
```
mgsergio: In older versions of editor the necessary geometry informations was missed form edits.xml file, so we cannot restore the original geometry of a particular feature and thus can't use the new algo that is dependant on a correct feature geometry to calculate scores.

mgsergio: @therearesomewhocallmetim, could you please review comments?

milchakov: готово

editor/feature_matcher.hpp
```diff
@@ -5,16 +5,29 @@
...
 
-#include "std/vector.hpp"
+#include "base/exception.hpp"
+
+#include <vector>
 
 namespace matcher
 {
+DECLARE_EXCEPTION(NotAPolygonException, RootException);
 /// Returns closest to the latLon node from osm or empty node if none is close enough.
```
therearesomewhocallmetim: Returns a node from OSM closest to the latLon, or an empty node if none is close enough.

editor/feature_matcher.hpp
```diff
@@ -5,16 +5,29 @@
...
 /// Returns a way from osm with similar geometry or empty node if can't find such way.
+/// Throws NotAPolygon exception when |geometry| is not convertible to a polygon.
 pugi::xml_node GetBestOsmWayOrRelation(pugi::xml_document const & osmResponse,
-                                       vector<m2::PointD> const & geometry);
+                                       std::vector<m2::PointD> const & geometry);
 /// Returns value form [-1, 1]. Negative values are used as penalty, positive as score.
-/// |lhs| and |rhs| - triangulated polygons;
-double ScoreTriangulatedGeometries(vector<m2::PointD> const & lhs, vector<m2::PointD> const & rhs);
+/// |lhs| and |rhs| - triangulated polygons.
+/// Throws NotAPolygon exception when lhs or rhs is not convertible to a polygon.
```
therearesomewhocallmetim: Throws NotAPolygon exception when either lhs or rhs is not convertible to a polygon.

editor/feature_matcher.hpp
```diff
@@ -5,16 +5,29 @@
...
+double ScoreTriangulatedGeometries(std::vector<m2::PointD> const & lhs,
+                                   std::vector<m2::PointD> const & rhs);
+/// Deprecated, use ScoreTriangulatedGeometries instead. In older versions of editor the necessary
+/// geometry informations was missed form edits.xml file, so we cannot restore the original geometry
+/// of a particular feature and thus can't use the new algo that is dependant on a correct feature
+/// geometry to calculate scores.
+// TODO(a): To remove it when version 8.0.x will no longer be supported.
```
therearesomewhocallmetim: Deprecated, use ScoreTriangulatedGeometries instead. In the older versions of the editor, the edits.xml file didn't contain the necessary geometry information, so we cannot restore the original geometry of a particular feature and thus can't use the new algo that is dependent on correct feature geometry to calculate scores.
 TODO(a): To remove it when version 8.0.x is no longer supported.

editor/feature_matcher.hpp
milchakov: готово

editor/feature_matcher.hpp
milchakov: заменил

editor/feature_matcher.hpp
milchakov: заменил

milchakov:
@mgsergio PTAL&MP

PR #8004, opened 2018-01-25
[MAPSME-6532] [ios] Fixed action bar buttons.

igrechuhin:


Jmaguerre:
Hi @igrechuhin sorry to comment on this PR that was merged so long ago but I was looking into the code after I reported this bug on the api repository https://github.com/mapsme/api-ios/issues/17 and found this PR where the back button is removed. Was this intended? I couldn't find the associated ticket for this if there's any.
Any response regarding this would be appreciated.

Thanks

PR #8010, opened 2018-01-26
Hotels in discovery manager.

VladiMihaylenko:


map/search_api.cpp
```diff
@@ -222,6 +223,34 @@ bool SearchAPI::SearchInViewport(ViewportSearchParams const & params)
...
+  p.m_mode = search::Mode::Viewport;
+  p.m_onResults = [params](Results const & results) {
+    if (!results.IsEndMarker())
+      return;
+
+    if (params.m_sortingType == DiscoverySearchParams::SortingType::None)
+      params.m_onResults(results);
+
+    Results r(results);
+    sort(r.begin(), r.end(), DiscoverySearchParams::HotelsSort());
```
mpimenov: Здесь их надо перенумеровать с помощью ```SetPositionInResults```.

map/search_api.cpp
```diff
@@ -222,6 +223,34 @@ bool SearchAPI::SearchInViewport(ViewportSearchParams const & params)
...
+  p.m_viewport = params.m_viewport;
+  p.m_position = params.m_viewportCenter;
+  p.m_maxNumResults = params.m_itemsCount;
+  p.m_mode = search::Mode::Viewport;
+  p.m_onResults = [params](Results const & results) {
+    if (!results.IsEndMarker())
+      return;
+
+    if (params.m_sortingType == DiscoverySearchParams::SortingType::None)
+      params.m_onResults(results);
```
mpimenov: return?

map/search_api.cpp
mpimenov: Please add an assert for SortingType

map/search_api.cpp
VladiMihaylenko: да

map/discovery/discovery_manager.cpp
```diff
@@ -27,28 +27,16 @@ Manager::Manager(Index const & index, search::CityFinder & cityFinder, APIs cons
...
+  search::DiscoverySearchParams p;
   p.m_query = GetQuery(type);
-  p.m_inputLocale = "en";
   p.m_viewport = params.m_viewport;
-  p.m_position = params.m_viewportCenter;
-  p.m_maxNumResults = params.m_itemsCount;
-  p.m_mode = search::Mode::Viewport;
-  return p;
-}
+  p.m_viewportCenter = params.m_viewportCenter;
```
ygorshenin: BTW, what's the reason to pass viewportCenter and viewport as individual params? It should not be hard to calculate center of a rectangle, right?

map/discovery/discovery_search_params.hpp
```diff
@@ -0,0 +1,43 @@
...
+{
+struct DiscoverySearchParams
+{
+  enum class SortingType
+  {
+    None,
+    HotelRating
+  };
+
+  struct HotelsSort
```
ygorshenin: HotelRatingSort? BTW, it should be named HotelRatingComparer, right?

map/discovery/discovery_search_params.hpp
```diff
@@ -0,0 +1,43 @@
...
+      UNUSED_VALUE(strings::to_float(lhs.m_metadata.m_hotelRating, l));
+      UNUSED_VALUE(strings::to_float(rhs.m_metadata.m_hotelRating, r));
+      return l > r;
+    }
+  };
+
+  using OnResults = std::function<void(Results const & results)>;
+
+  std::string m_query;
+  size_t m_itemsCount = 0;
```
ygorshenin: #include cstddef please.

map/search_api.cpp
```diff
@@ -222,6 +223,34 @@ bool SearchAPI::SearchInViewport(ViewportSearchParams const & params)
...
+  CHECK(!params.m_query.empty(), ());
+  CHECK_GREATER(params.m_itemsCount, 0, ());
+
+  SearchParams p;
+  p.m_query = params.m_query;
+  p.m_inputLocale = "en";
+  p.m_viewport = params.m_viewport;
+  p.m_position = params.m_viewportCenter;
+  p.m_maxNumResults = params.m_itemsCount;
+  p.m_mode = search::Mode::Viewport;
```
ygorshenin: Note that this effectively cancels current viewport search, do you really want this?

map/search_api.cpp
```diff
@@ -222,6 +223,36 @@ bool SearchAPI::SearchInViewport(ViewportSearchParams const & params)
...
+  p.m_onResults = [params](Results const & results) {
+    if (!results.IsEndMarker())
+      return;
+
+    if (params.m_sortingType == DiscoverySearchParams::SortingType::None) {
```
ygorshenin: Please, use switch instead.

map/search_api.cpp
VladiMihaylenko: ok

ygorshenin: @mpimenov, can we get rid of this position-in-results? As was discussed, it would be nice to split results by types: suggests, results, lat-lons. After that there will be no need in "position in results".

mpimenov: Sure, but it is a matter for another PR.

map/discovery/discovery_manager.cpp
ygorshenin: Aha, it seems that m_viewportCenter should be renamed to m_position?

map/discovery/discovery_search_params.hpp
VladiMihaylenko: done

map/search_api.cpp
VladiMihaylenko: I definitely want to use Viewport mode, so I guess in this case it's ok to cancel the previous search.

ygorshenin: Okay, np. But note that it's not a big deal to add your own mode, that will work exactly as search in Viewport.

map/search_api.cpp
```diff
@@ -222,6 +223,44 @@ bool SearchAPI::SearchInViewport(ViewportSearchParams const & params)
...
+
+    switch (params.m_sortingType)
+    {
+    case DiscoverySearchParams::SortingType::None:
+      params.m_onResults(results);
+      break;
+    case DiscoverySearchParams::SortingType::HotelRating:
+    {
+      Results r(results);
+      sort(r.begin(), r.end(), DiscoverySearchParams::HotelRatingComparator());
```
ygorshenin: Much better (and safer) is to implement following method in Results:
```c++
class Results {
  template <typename Fn>
  void SortBy(Fn && comparer)
  {
    sort(begin(), end(), forward<Fn>(comparer));
    for (...)
       r[i].SetPositionInResults(i);
  }
};
```

ygorshenin: After that you'll be able to test that in UNIT TESTS, and @mpimenov won't forget to fix this code after refactoring of results.

VladiMihaylenko: Ok, nice idea

VladiMihaylenko:
@mpimenov ptal

search/search_tests/results_tests.cpp
```diff
@@ -0,0 +1,27 @@
...
+  Results r;
+  MwmSet::MwmId id;
+  for (uint32_t i = 5; i != 0; --i)
+  {
+    r.AddResultNoChecks({{id, i}, {} /* pt */, {} /* str */, {} /* address */,
+                         {} /* featureTypeName */, {} /* featureType */, {} /* metadata */});
+  }
+
+  TEST_EQUAL(r[0].GetFeatureID().m_index, 5, ());
+  TEST_EQUAL(r[0].GetPositionInResults(), 0, ());
```
mpimenov: Let's test all indices, and not just 0?

VladiMihaylenko: ok

PR #8046, opened 2018-02-01
Fix boost minimal version to 1.59

greshilov:
I've caught error like this, when tried to build our project with boost ver. 1.58:
```
/*/omim/editor/feature_matcher.cpp:270:24: error: no member
      named 'is_empty' in namespace 'boost::geometry'
  if (boost::geometry::is_empty(their))
```
Hence minimal version is now 1.59.


greshilov:
JCRT

greshilov:
JCRT

greshilov:
JCRT

syershov:
We can install boost 1.54 or 1.64 on Centos6 and 1.66 on Centos7.

greshilov:
JTRIT

greshilov:
@bykoianko PTAL

PR #8051, opened 2018-02-02
[fix-text] [ios] Fixed booking title in discovery manager.

igrechuhin:


mapsmetest:

PR #8117, opened 2018-02-17
Hungarian translation strings

baueri:
I made a bunch of translation fixes for the Hungarian language.

### DIFF
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -43,7 +43,7 @@ nl:Eten
 fi:Ruoka
 fr:Nourriture
 de:Essen
-hu:Élelmiszer
+hu:Étkezés
 id:Makanan
 it:Cibo
 ja:飲食
@@ -309,7 +309,7 @@ nl:3^Geldautomaat
 fi:^Pankkiautomaatti
 fr:^GAB
 de:3^Geldautomat
-hu:4^Pénzautomata
+hu:4^Bankautomata|ATM
 id:2^ATM
 it:3^Bancomat
 ja:1^ATM
@@ -343,7 +343,7 @@ nl:geld
 fi:2Raha
 fr:DAB|argent
 de:Bankomat|Geld
-hu:pénz|ATM
+hu:bankautomata|ATM|pénz
 id:uang
 it:soldi
 ja:atm|お金|マネー|金|引き出し|預金|振り込み
@@ -536,7 +536,7 @@ nl:3^Tearoom|3brasserie|3café|3restaurant
 fi:2^Kahvila
 fr:3^Café|3restaurant|3resto
 de:3^Café|3Restaurant|Kaffee
-hu:3^Kávézó|3étterem|3élelmiszer
+hu:3^Kávézó|cukrászda|3étterem|3élelmiszer
 id:3^Kafe|3restoran
 it:3^Caffè|3ristorante|pasto|mangiare
 ja:1^カフェ|喫茶|食べ物|コーヒー|珈琲|お茶|茶|紅茶|食事|軽食
@@ -569,7 +569,7 @@ nl:4^Fast food|3frituur|fritkot|3restaurant|afhaaleten
 fi:3^Pikaruokala
 fr:4^Restauration rapide|3restaurant|3café|à emporter|3resto
 de:4^Fast-Food|Takeaway|Restaurant|Café|Pizzeria
-hu:4^Gyorsétterem|3étterem|3kávézó|3élelmiszer
+hu:4^Gyorsétterem|büfé|3étterem|3kávézó|3élelmiszer
 id:^Cepat saji|3restoran
 it:4^Fast food|takeaway|caffè|ristorante|pasto|mangiare
 ja:1^ファストフード|レストラン|テイクアウト|持ち帰り|軽食|食事
@@ -601,7 +601,7 @@ nl:3^Restaurant|3café
 fi:3^Ravintola
 fr:3^Restaurant|3café|3resto
 de:3^Restaurant|3Café|Gasthaus|Gaststube
-hu:3^Étterem|3kávézó|3élelmiszer
+hu:3^Étterem|vendéglő|3kávézó|3élelmiszer
 id:^Restoran|3kafe
 it:3^Ristorante|3caffè|mangiare
 ja:1^レストラン|食べ物|食事|2ファミレス
@@ -713,7 +713,7 @@ nl:4^Schoonheidsmiddelen
 fi:^Kosmetiikka|kauneudenhoito
 fr:4^Produits de beauté|soins de beauté
 de:4^Kosmetik|Schönheitspflege
-hu:4^Kozmetikum|házi gondozás
+hu:4^Kozmetikum|szépségápolás
 id:4^Kosmetik
 it:^Cosmetici|materiale per la casa
 ja:^化粧品|ホームケア
@@ -747,7 +747,7 @@ nl:^Buurtwinkel|kruidenier
 fi:^Lähikauppa
 fr:^Supérette
 de:^Gemischtwarenladen
-hu:^Csemegebolt|ábécé
+hu:^Kis élelmiszerbolt|vegyesbolt
 id:^Mini market
 it:^Minimarket
 ja:^コンビニエンスストア|コンビニ
@@ -780,7 +780,7 @@ nl:^Tuincentrum
 fi:^Taimitarha
 fr:^Crèche|garderie d'enfants
 de:^Gärtnerei|Garten-Center
-hu:^Faiskola|kertészeti üzlet
+hu:^Kertészet
 id:^Penitipan anak
 it:^Vivaio|articoli da giardinaggio
 ja:^保育園|託児所
@@ -812,7 +812,7 @@ nl:^Mobiele telefoonwinkel|elektronicawinkel
 fi:^Elektroniikkakauppa
 fr:^Magasin de téléphonie|magasin d'électronique
 de:^Handyladen|Eletronikgeschäft
-hu:^Mobiltelefon üzlet|elektronikai szaküzlet
+hu:^Mobiltelefon-szaküzlet|elektronikai szaküzlet
 id:^Toko telepon
 it:^Negozio di telefonia mobile|negozio di elettronica
 ja:^携帯ショップ
@@ -845,7 +845,7 @@ nl:^Bloemist|bloemenwinkel
 fi:^Floristi|kukkakauppa
 fr:^Fleuriste
 de:^Florist|Blumengeschäft|Blumenladen|2Laden
-hu:^Virágos|virágbolt
+hu:^Virágbolt|virágárus
 id:^Tukang bunga|toko bunga
 it:^Fiorista|fioraio
 ja:^フローリスト/花屋|お買い物|ショップ|商店|雑貨
@@ -909,7 +909,7 @@ nl:^Meubelzaak
 fi:^Huonekalukauppa
 fr:^Magasin de meubles
 de:^Möbelgeschäft|2Laden
-hu:^Bútoráruház
+hu:^Bútorbolt
 id:^Toko mebel
 it:^Negozio di mobili
 ja:^家具店|お買い物|ショップ|商店|雑貨
@@ -941,7 +941,7 @@ nl:^Slijterij|drankhandel
 fi:^Alkoholikauppa|alkoholi
 fr:^Magasin de vins et spiritueux|vins et spiritueux
 de:^Spirituosengeschäft|Spirituosen|2Laden
-hu:^Italbolt|ital
+hu:^Italbolt|ital|alkohol|szesz
 id:^Toko alkohol|minuman keras
 it:^Enoteca
 ja:^酒屋|お買い物|ショップ|商店|雑貨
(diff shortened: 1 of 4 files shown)


data/categories.txt
```diff
@@ -536,7 +536,7 @@ nl:3^Tearoom|3brasserie|3café|3restaurant
 fi:2^Kahvila
 fr:3^Café|3restaurant|3resto
 de:3^Café|3Restaurant|Kaffee
-hu:3^Kávézó|3étterem|3élelmiszer
+hu:3^Kávézó|cukrászda|3étterem|3élelmiszer
```
mpimenov: Are you sure? We do not include it in other languages, there is shop-confectionery for this.

data/categories.txt
```diff
@@ -43,7 +43,7 @@ nl:Eten
 fi:Ruoka
 fr:Nourriture
 de:Essen
-hu:Élelmiszer
+hu:Étkezés
```
mpimenov: Is this more used than what we had when searching for "Food"?

data/categories.txt
```diff
@@ -747,7 +747,7 @@ nl:^Buurtwinkel|kruidenier
 fi:^Lähikauppa
 fr:^Supérette
 de:^Gemischtwarenladen
-hu:^Csemegebolt|ábécé
+hu:^Kis élelmiszerbolt|vegyesbolt
```
mpimenov: Is the word "Kis" required here? Isn't plain "élelmiszerbolt" enough?

data/categories.txt
```diff
@@ -1263,7 +1263,7 @@ nl:^Sportartikelen
 fi:^Urheilukauppa
 fr:^Articles de sport
 de:^Sportartikel|Fitnessgeschäft|2Laden
-hu:^Sporteszközök
+hu:^Sporteszközbolt
```
mpimenov: From what I understand, you're replacing things like "sports equipment" to "sports equipment shop" in several places. Are you sure that people do not search for these shops by typing the terms of the former kind?
That is, if you think it is common to type "Sporteszközök" when searching for this shop, please leave the synonym in here and in other places: ```hu:^Sporteszközbolt|Sporteszközök```.
The "^" sign is before the term that will be shown to the user, the other synonyms are there only to help the search.

data/categories.txt
```diff
@@ -1295,7 +1295,7 @@ nl:3^Supermarkt
 fi:3^Supermarketti
 fr:3^Supermarché
 de:3^Supermarkt|Laden
-hu:3^Bevásárlóközpont
+hu:3^Szupermarket|élelmiszerbolt
```
mpimenov: The synonyms with "^" will be preferred when showing objects to users. Is "Szupermarket" really the more commonly used word here?

Also, the "3" here means that we will start suggesting "szupermarket" after 3 typed characters, i.e. "szu". This seems appropriate here but in other places you might want to reconsider (by removing or adding a digit).

data/categories.txt
```diff
@@ -1426,7 +1426,7 @@ nl:^Computerwinkel
 fi:^Tietokonekauppa
 fr:^Magasin d'informatique
 de:^Computergeschäft
-hu:^Számítógép üzlet
+hu:^Számítógépüzlet
```
mpimenov: Are you sure? I see both forms being used.

data/categories.txt
```diff
@@ -1584,7 +1584,7 @@ nl:^Kledingwinkel|4kleding
 fi:^Vaatekauppa
 fr:^Boutique de vêtements|4vêtements
 de:^Bekleidungsgeschäft|Kleidung|Laden
-hu:^Ruhabolt|4ruhák
+hu:^Ruházati bolt|3ruhák
```
mpimenov: Maybe move the "3" to the first synonym?

data/categories.txt
```diff
@@ -9164,6 +9168,7 @@ zh-Hant:^山屋
 sponsored-holiday
 en:^Christmas or New Year Celebrations
 ru:^Празднование Рождества или Нового года
+hu:^Gyűjtsd össze az összes tételt, és nyerj!
```
mpimenov: Your translation seems outdated.

data/categories.txt
```diff
@@ -4775,7 +4778,7 @@ nl:3^Berghut|alpenhut|hotel
 fi:^Vuorimajoitus|hotelli
 fr:^Hôtel de montagne|hôtel|auberge de jeunesse|cabane alpine
 de:4^Berghütte|Almhütte|Berghotel|Hotel|Unterkunft
-hu:^Hegyi menedék|kabin|hotel|szálloda
+hu:^Menedékház (személyzettel)|alpesi menedékház|szálloda|ifjúsági szálló|hostel
```
mpimenov: Why parentheses?

data/categories.txt
```diff
@@ -5669,7 +5673,7 @@ nl:^Airconditioning|luchtbehandeling
 fi:^Ilmastointilaite
 fr:^Climatiseur
 de:^Heizung|Lüftung|Klima|Installateur
-hu:^Hűtés-fűtés szerelő|légkondicionáló
+hu:^Fűtés-, szellőztetés- és légkondicionálás-szerelő|hűtés- és fűtésszerelő|fűtésszerelő|légkondicionálás-szerelő
```
mpimenov: Can we shorten it somehow? Break the first synonym into multiple ones maybe?
It may be too long even for the UI, and I'm sure noone really types exactly such strings when using search.

data/categories.txt
```diff
@@ -6271,7 +6275,7 @@ nl:^Droogkuis|wassen
 fi:^Kuivapesula|pesula
 fr:^Nettoyage à sec|Lavage
 de:^Chemische Reinigung|Reinigung
-hu:^Ruhatisztító|tisztító
+hu:^Ruhatisztító|vegytisztító|2patyolat|2tisztító
```
mpimenov: Are these important enough to start suggesting them after the first 2 letters?

data/categories.txt
```diff
@@ -8845,7 +8849,7 @@ el:^Κατάστημα ερωτικών ειδών
 es:^Sex Shop
 fi:^Erotiikkaliike
 fr:^Boutique érotique
-hu:^Erotikus bolt
+hu:^Erotikus bolt|szexbolt
```
mpimenov: Isn't this attachment of "bolt" supposed to be written without extra spaces?

mpimenov:
@baueri ping

data/categories.txt
baueri: It is a constant question what Hungarian "cukrászda" is. A cukrászda in Hungary is actually a café: you enter, order your coffee from the waiter, chat with your friends etc. The only difference between a café and a cukrászda is that in a cukrászda you have (more and self-made) sweets (cakes, pies, biscuits and, in summertime, ice) available. However, the border between the two is vague. Therefore café and cukrászda may be together. A Hungarian cukrászda is not really a shop, it is more like a café or a restaurant (i.e. an amenity in OSM terms)

A shop=confectionery​ is a shop where you enter, buy your sweets and take them away, so no chairs and tables, no waiter, just like a real shop. This is a very rare species in Hungary and would be "édességbolt".

data/categories.txt
baueri: Élelmiszer is food in the sense that it is a product which you buy and eat. If you tap on an icon "Élelmiszer" you expect shops where you can buy food (e.g. a supermarket or a smaller convenience store). However, in maps.me it refers to places where you can eat, i.e. restaurants, cafés and so on. So it is more suitable to have "Étkezés" (literally eating) there than "Élelmiszer" which is  bit misleading.

data/categories.txt
baueri: ​It might be but élelmiszerbolt is a general term which may apply also to big shops, e.g. supermarkets. Kis élelmiszerbolt is more specific and gives a clear explanation that at that POI you will find a relatively small shop. It is a common term also in other OSM applications (OSMand, Vespucci, JOSM, etc.).​

data/categories.txt
baueri: OK, it is not a big difference so may have both or just have "sporteszközök". I added "bolt" because it is a shop so it is perhaps more exact that way.​

data/categories.txt
baueri: Bevásárlóközpont is a big mall where several shops can be found. Szupermarket means supermarket. Élelmiszerbolt is a broader term (see at convenience).

With the numbers of the leters (e.g. 3^) you may have right but I don't feel an expert in how to determine the right number of letters after wich solutions will be suggested by the app. So it may remain so.​

data/categories.txt
baueri: ​According to Hungarian grammar, they should be written together.​ (It is so even many people don't know this and write it separated. I think it better for maps.me to be correct.)

data/categories.txt
baueri: ​OK, it can be so.​

data/categories.txt
baueri: ​Yes, indeed. The HU translation should be:
"Karácsonyi vagy újévi ünnepség"

data/categories.txt
```diff
@@ -2349,7 +2349,7 @@ nl:^Synagoge|tempel|gebedsplaats|gebedshuis
 fi:^Synagoga|uskonnollinen kohde
 fr:^Synagogue|lieu de culte|temple
 de:^Synagoge|Anbetungsstätte|Tempel
-hu:^Zsinagóga|templom
+hu:^Zsinagóga|imaház
```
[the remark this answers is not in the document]
baueri: ​No, a synagogoue is a jewish imaház (house of prayer). It is relativey common in Hungarian. Templom refers​ mostly to christian places, a synagogue is not a templom. (Historic reference: there was one jewish "templom", in Jerusalem, but it doesn't exist anymore since ~100 AD.)

data/categories.txt
```diff
@@ -2512,7 +2512,7 @@ nl:^Monument|attractie
 fi:^Monumentti
 fr:^Monument|attraction|tourisme
 de:^Monument|Denkmal
-hu:^Műemlék|Szobor|túrizmus|látványosság
+hu:^Épületszerű, monumentális emlékmű|monumentális emlékmű|3turizmus|3látványosság
```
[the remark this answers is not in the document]
baueri: ​Yes, I am sure. Monument is a very specific and difficult term in OSM and there aren't really ​short, concise but also exact terms to it in Hungarian. In Hungary there are very few buildings which can be considered "Monument" in the OSM-sense of the word. Therefore we coined this definition (also used in other OSM applications) which is though a bit long and clumsy, but at least exact, so if you find a "Monument" on the map, you can expect what you will find on site, in reality.

data/categories.txt
```diff
@@ -2578,7 +2578,7 @@ nl:^Kasteel|burcht|attractie
 fi:^Linna
 fr:^Château|attraction|tourisme
 de:^Burg|Schloss|Festung
-hu:^Kastély|túrizmus|látványosság
+hu:^Váe|kastély|erőd|turizmus|látványosság
```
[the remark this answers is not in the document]
baueri: ​It should be "Vár", sorry. Should I make a new commit with the fix?

data/categories.txt
```diff
@@ -2841,7 +2841,7 @@ nl:3^Theater|schouwburg
 fi:^Teatteri
 fr:3^Théâtre
 de:3^Theater
-hu:3^Színház
+hu:3^Színház/teátrum
```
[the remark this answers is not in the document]
baueri: ​Yes, you are right. Sorry.​

data/categories.txt
```diff
@@ -4388,7 +4391,7 @@ nl:^Straat|st|str|laan|weg
 fi:^Katu
 fr:^Rue|avenue|av|boul|boulevard|route
 de:^Straße|Boulevard|Avenue|Weg|Gasse
-hu:^Utca|út|u
+hu:^Utca|út|u.|sugárút|sgt.|körút|krt.
```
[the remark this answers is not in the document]
baueri: You are welcome!

data/categories.txt
baueri: You may have right: it is not indispensable so you can leave it. (In another application there are two types of huts, one is with staff and the other is without staff and they are indicated in parentheses. But for maps.me it is perfect also without parenteses.)

data/categories.txt
baueri: ​Yes, the Hungarian translation is very long, even if it is exact. Unfortunately, we don't such a nice abbreviation like that in English. Then let's option for the German solution:
"^Fűtésszerelő|hűtésszerelő|szellőztetésszerelő|​légkondicionálás-szerelő"

data/categories.txt
baueri: ​Well, it should be rather 3 or 4 there.​

data/categories.txt
baueri: ​No, it is good so.

data/categories.txt
```diff
@@ -3435,6 +3436,7 @@ ar:^متجر أدوية|صيدلية
 nl:verdeelstation
 fi:3Farmasia
 fr:dispensaire
+hu:Gyógyszertár
```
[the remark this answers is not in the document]
baueri: ​OK, you are right. You can also put "|patika​" afterwards which is also commonly used for pharmacies.

data/categories.txt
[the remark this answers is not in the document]
mpimenov: Yes, add a commit to this PR please. You may also need to rebase to the master branch.

greshilov:

mpimenov:
@baueri ping

mpimenov:
@baueri ping

PR #8171, opened 2018-02-26
[search] Fixed PrependCity.

mpimenov:


search/result.hpp
```diff
@@ -108,7 +108,7 @@ class Result
   std::pair<uint16_t, uint16_t> const & GetHighlightRange(size_t idx) const;
   size_t GetHighlightRangesCount() const { return m_hightlightRanges.size(); }
 
-  void AppendCity(string const & name);
+  void PrependCity(string const & name);
```
bykoianko: `std::string`?

search/result.cpp
```diff
@@ -3,6 +3,8 @@
 #include "search/common.hpp"
 #include "search/geometry_utils.hpp"
 
+#include "base/string_utils.hpp"
+
 #include <algorithm>
```
bykoianko: `#include <algorithm>` есть в hpp

search/result.hpp
mpimenov: done

search/result.cpp
mpimenov: done

PR #8186, opened 2018-02-27
[base] Fixed a data race in RootException.

mpimenov:


mpimenov:
JTTP

base/exception.hpp
```diff
@@ -8,24 +9,19 @@
...
-  {
-  }
+  RootException(char const * what, std::string const & msg);
 
   virtual ~RootException() throw()
   {
   }
 
-  virtual char const * what() const throw();
+  virtual char const * what() const throw() { return m_whatWithAscii.c_str(); }
```
VladiMihaylenko: noexcept instead of throw()

VladiMihaylenko: override

mpimenov: done

PR #8250, opened 2018-03-05
Fixed several warnings.

mpimenov:


map/map_tests/ge0_parser_tests.cpp
```diff
@@ -223,7 +221,7 @@ UNIT_TEST(LatLonFullAndClippedCoordinates)
     }
   }
 
-  for (int coordSize = 1; coordSize <= 8; ++coordSize)
+  for (size_t coordSize = 1; coordSize <= 8; ++coordSize)
```
VladiMihaylenko: uint8_t ?

map/map_tests/ge0_parser_tests.cpp
```diff
@@ -81,13 +80,12 @@ bool ConvergenceTest(double lat, double lon, double latEps, double lonEps)
...
 }
-
-}  // unnamed namespace
+}  // namespace
 
 UNIT_TEST(Base64DecodingWorksForAValidChar)
 {
   Ge0ParserForTest parser;
-  for (size_t i = 0; i < 64; ++i)
+  for (int i = 0; i < 64; ++i)
```
VladiMihaylenko: uint8_t?

map/map_tests/ge0_parser_tests.cpp
mpimenov: No, why?
uint8_t is still too big, so I prefer to stick to a more common type.

map/map_tests/ge0_parser_tests.cpp
mpimenov: That would require changing the signature of the obscure MapsWithMe_Base64Char function. I'd rather not.

VladiMihaylenko: ok

PR #8330, opened 2018-03-19
Fix strings autoupd

igortomko:


mpimenov:
Cherry-pick of #8328, I assume?

PR #8391, opened 2018-03-27
[cmake] Fix CMP0071 policy

greshilov:
Fix for cmake version 3.10 and higher. ([More info](https://cmake.org/cmake/help/git-stage/policy/CMP0071.html))


greshilov:
@Zverik @mpimenov PTAL

PR #8407, opened 2018-03-30
Using saved countryId instead of reference.

VladiMihaylenko:
We used to use the countryId which is stored in queuedCountry. But queuedCountry was stored by a reference and sometimes it could be removed from the queue so reference became dangling. Using a copy of the countryId must be more safety.


VladiMihaylenko:
@bykoianko ptal

PR #8482, opened 2018-04-11
[coding] Refactored EncodeDelta for points.

mpimenov:


coding/geometry_coding.hpp
```diff
@@ -28,9 +29,61 @@ using InDeltasT = array_read<uint64_t>;
...
+  WriteVarInt(sink, dy);
+}
+
+// Writes the difference of two 2d vectors to sink. The vector |next|
+// must have both its coordinates greater than or equal to those
+// of |curr|. While this condition is unlikely when encoding polylines,
+// the function may be useful when encoding rectangles, in particular
+// bounding boxes of shapes.
+template <typename Sink>
+void EncodePositivePointDelta(Sink & sink, m2::PointU const & curr, m2::PointU const & next)
```
ygorshenin: NonNegative :)

mpimenov: I though about this (see the comment to function) but won't it in fact hard readability?

ygorshenin: Is it possible to subtract one PointU from another? If so, why do you need this method at all? I mean, just create a method that'll write to sink components of PointU as VarUints.

mpimenov: Note the paranoid checks.

mpimenov: And the asserts.
While this function looks superfluous, I thought it's better to leave it for now so that someone who needs to serialize bounding boxes stumbles upon it.

ygorshenin: I mean, why do you need to pass two arguments? Just pass a single argument - delta. So it'll be responsibility of the caller side to check that delta is non-negative (as it happens in bounding boxes).

mpimenov:
@ygorshenin @rokuz PTAL

coding/geometry_coding.hpp
mpimenov: In this case the method looks even more superfluous.
I've returned it back to the private sections in cities_boundaries_serdes for now.

PR #8483, opened 2018-04-11
[DNM] [base] Marked the expressions in ASSERT and CHECK as unlikely.

mpimenov:


mpimenov:
JTTP

base/assert.hpp
```diff
@@ -30,33 +30,33 @@ namespace my
     ASSERT_CRASH();
 
 // TODO: Evaluate X only once in CHECK().
-#define CHECK(X, msg) do { if (X) {} else { \
+#define CHECK(X, msg) do { if (unlikely(X)) {} else { \
```
biodranik: Should the logic be reversed (likely)? All `CHECK(false)` in the code are for exceptional/rare run-time fatal cases, so obviously, the most popular case is when X == true (and we do nothing in the if {} clause).

mpimenov: Indeed. Nice catch, thanks.

mpimenov:
JTTP

biodranik:
@mpimenov it would be great if you share any performance improvements after this fix.

ygorshenin:
@biodranik I waited for this comment!!! :)

ygorshenin:
@biodranik I waited for this comment!!! :)

base/base.hpp
```diff
@@ -33,3 +33,6 @@ static_assert(MY_DEBUG_DEFINED ^ MY_RELEASE_DEFINED, "Either Debug or Release sh
 
 // platform macroses
 #include "std/target_os.hpp"
+
+#define likely(expr) (__builtin_expect(!!(expr), 1))
```
VladiMihaylenko: Why do you need excess parentheses? Do you assume to use it like following ```if likely(a)```?

mpimenov: done

mapsmetest:

biodranik:
@mpimenov ping?

PR #8503, opened 2018-04-13
[geocoder] Add borders class to support region borders in reverse serverside geocoder

tatiana-yan:


tatiana-yan:
JTTP

indexer/borders.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "base/logging.hpp"
+#include "base/macros.hpp"
+
+#include "defines.hpp"
+
+using namespace std;
+
+namespace
+{
+template <class ReaderT>
```
mpimenov: s/ReaderT/Reader/

indexer/borders.cpp
```diff
@@ -0,0 +1,95 @@
...
+class BordersVector
+{
+  DISALLOW_COPY(BordersVector);
```
mpimenov: This is usually the last line, not the first.

indexer/borders.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <list>
+#include <map>
+#include <string>
+
+namespace indexer
+{
+class Borders
+{
+public:
+  bool IsPointInside(uint64_t id, m2::PointD point) const
```
mpimenov: please move to .cpp

indexer/borders.hpp
```diff
@@ -0,0 +1,65 @@
...
+    return false;
+  }
+
+  bool Deserialize(std::string filename);
+
+  template <typename BordersVec>
+  void DeserializeFromVec(BordersVec const & vec)
+  {
+    vec.ForEach([this](uint64_t id, std::vector<m2::PointD> const & outer,
+                       std::list<std::vector<m2::PointD>> const & inners) {
```
mpimenov: why std::list?

indexer/borders.hpp
```diff
@@ -0,0 +1,65 @@
...
+      it->second.m_outer = m2::RegionD(outer);
+      for (auto const & inner : inners)
+        it->second.m_inners.push_back(m2::RegionD(inner));
+    });
+  }
+
+private:
+  struct Border
+  {
+    bool IsPointInside(m2::PointD point) const
```
mpimenov: const & point

indexer/borders.hpp
```diff
@@ -0,0 +1,65 @@
...
+    {
```
mpimenov: please move to .cpp

indexer/borders.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+private:
+  FilesContainerR m_cont;
+  BordersVector<ModelReaderPtr> m_vector;
+};
+}  // namespace
+
+namespace indexer
+{
+bool Borders::Deserialize(string filename)
```
mpimenov: string const &

indexer/borders.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+public:
+  BordersVector(ReaderT const & reader) : m_recordReader(reader, 256 /* expectedRecordSize */) {}
+
+  template <class ToDo>
+  void ForEach(ToDo && toDo) const
+  {
+    m_recordReader.ForEachRecord([&](uint32_t pos, char const * data, uint32_t /*size*/) {
+      ArrayByteSource src(data);
+      serial::GeometryCodingParams cp = {};
```
mpimenov: This should be read from the header, shouldn't it?

indexer/borders.cpp
```diff
@@ -0,0 +1,95 @@
...
+      uint64_t id;
+      ReadPrimitiveFromSource(src, id);
+      size_t size;
+      ReadPrimitiveFromSource(src, size);
+      size_t outerSize;
+      ReadPrimitiveFromSource(src, outerSize);
+      vector<m2::PointD> outer(outerSize);
+      // todo:(@t.yan) consider delta coding techniques.
+      for (size_t i = 0; i < outer.size(); ++i)
```
mpimenov: nit: if you don't use reserve+emplace_back, you can write range-based for here instead

indexer/borders.cpp
tatiana-yan: done

indexer/borders.cpp
tatiana-yan: done

indexer/borders.hpp
tatiana-yan: done

indexer/borders.hpp
tatiana-yan: done

indexer/borders.hpp
tatiana-yan: done

indexer/borders.cpp
tatiana-yan: пока что генерация в несколько разобранном состоянии и хедера там нет:) по мере работы над генерацией буду менять и этот код -- чтение параметров из хедера, дельта кодирование

indexer/borders.cpp
tatiana-yan: done

tatiana-yan:
JTTP

indexer/borders.hpp
tatiana-yan: ok, vector

indexer/borders.cpp
tatiana-yan: done

indexer/borders.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <cstdint>
+#include <map>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace indexer
+{
+class Borders
```
mpimenov: Please add a comment to this class: what it holds, where it is supposed to be used and what the inner structure is . The meaning of inner/outer is not immediately obvious.

tatiana-yan: done

PR #8605, opened 2018-04-26
[android] updater dialog crash fix

milchakov:
https://fabric.io/mapsme/android/apps/com.mapswithme.maps.pro/issues/5ae1c43c638393737a4c2da5?time=last-seven-days


alexzatsepin:
@milchakov PTAL

PR #8626, opened 2018-05-04
[RELEASE_ONLY] Using first matched ugc type to send to server.

VladiMihaylenko:


ugc/ugc_tests/storage_tests.cpp
```diff
@@ -313,7 +312,13 @@ UNIT_CLASS_TEST(StorageTest, ContentTest)
   ToJSONObject(*embeddedNode.get(), "data_version", lastIndex.m_dataVersion);
   ToJSONObject(*embeddedNode.get(), "mwm_name", lastIndex.m_mwmName);
   ToJSONObject(*embeddedNode.get(), "feature_id", lastIndex.m_featureId);
-  ToJSONObject(*embeddedNode.get(), "feature_type", classif().GetReadableObjectName(lastIndex.m_matchingType));
+  ftraits::UGCRatingCategories categories;
+  for (auto & r : newUGC.m_ratings) {
```
mpimenov: {
}

indexer/ftraits.hpp
```diff
@@ -190,6 +197,19 @@ class UGC : public TraitsBase<UGC, UGCItem>
...
   }
+
+  static boost::optional<std::string> GetMatchedType(UGCRatingCategories const & categories)
+  {
+    auto const & matcher = GetMatcher();
+    auto const end = matcher.End();
+    for (auto it = matcher.Begin(); it != end; ++it)
+    {
+      if (it->second.m_categories == categories)
+        return it->second.m_matchedType;
```
mpimenov: Consider adding a comment that explains why we return the first type that matches here.

PR #8715, opened 2018-05-17
update categories.txt

imanmoghimiq30:
some persian translate fix

### DIFF
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -131,7 +131,7 @@ el:Βενζίνη
 he:גז
 sk:3^Čerpacia stanica
 sw:Sheli
-fa:سوخت
+fa:سوخت|بنزین|دیزل|گازوئیل
 
 @parking
 en:^Parking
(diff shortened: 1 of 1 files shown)


data/categories.txt
```diff
@@ -3731,7 +3731,7 @@ el:^Τουαλέτα
 he:^שירותים
 sk:^Záchody
 sw:^Choo
-fa:توالت
+fa:دستشویی|توالت
```
mpimenov: Please add the "^" sign to the preferable translation.

data/categories.txt
```diff
@@ -3751,7 +3751,7 @@ vi:Nhà vệ sinh
 zh-Hant:2洗手間|wc
 el:μπάνιο|χώρος ανάγκης|αποχωρητήριο|wc
 sk:3Toalety|wc
-fa:توالت
+fa:توالت|دستشویی
```
mpimenov: It's better to remove this line: both synonyms will be added by the group term (@toilet).

imanmoghimiq30:
I do it 
tnx

PR #8771, opened 2018-05-28
[desktop] Fix numeric locale for desktop tests

greshilov:


.gitignore
```diff
@@ -25,6 +25,7 @@ data/drules_proto_design.bin
 data/colors_design.txt
 data/patterns_design.txt
 data/route_points.dat
+data/marketing_settings.ini
```
mpimenov: sort?

greshilov: Moved in correct section.

mpimenov:
Previous PR here: https://github.com/mapsme/omim/pull/996

1. I don't really like the idea of changing env from within the test code. Why not pass when running the tests?
2. One possible problem is that we are going to have all tests passing and the app will still crash. This PR will only make things worse: the code working with floating point textual representations should be as robust as possible.
3. The common code from qt/main.cpp and this PR should be factored out, all comments and todos included. Consider adding https://github.com/google/double-conversion to the list of suggestions.

PR #8797, opened 2018-05-30
[android] Added showing custom ads choice icon for all mopub native ads

alexzatsepin:
https://jira.mail.ru/browse/MAPSME-7768


alexzatsepin:
@yoksnod @milchakov PTAL

PR #8827, opened 2018-06-04
[ios] Add terms and privacy policy to about screen

beloal:


beloal:
https://jira.mail.ru/browse/MAPSME-7291

PR #8870, opened 2018-06-15
[tools][qmake] Delete old tools and qmake files

greshilov:
We don't support qmake.


mpimenov:
Please add the [tools] and [qmake] tags.

greshilov:
@mpimenov in designer tool build:
`qtCMakeLists.txt:268`

PR #8873, opened 2018-06-17
Fix expat_impl.h.

rdoeffinger:
Missing argument to XML_SetUnknownEncodingHandler
(compare to lib/expat.h) causes compile error.
Not sure how nobody else has noticed that yet?


3party/expat/expat_impl.h
```diff
@@ -307,7 +307,7 @@ class CExpatImpl
   {
     assert (m_p != NULL);
     XML_SetUnknownEncodingHandler (m_p,
-      fEnable ? UnknownEncodingHandler : NULL);
```
tatiana-yan: It looks like build fix, but not EnableUnknownEncodingHandler fix.
XML_SetUnknownEncodingHandler sets UnknownEncodingHandler and data pointer.
The first thing UnknownEncodingHandler does is data pointer dereference. So it's bad to pass NULL.
I think it's better to comment out EnableUnknownEncodingHandler than such fix.

tatiana-yan:
>Missing argument to XML_SetUnknownEncodingHandler
(compare to lib/expat.h) causes compile error.
Not sure how nobody else has noticed that yet?

It's strange but Clang 9.1 and gcc 6.3 are ok with it.

rdoeffinger:
I have not yet tested, maybe this file is only used and the issue triggered when running "cmake ." instead of using the script?
That causes something weird in the build system, I also had lots of warnings that disappeared since I instead used the script to run cmake and build in a separate directory.

tatiana-yan:
>I have not yet tested, maybe this file is only used and the issue triggered when running "cmake ." instead of using the script?
That causes something weird in the build system, I also had lots of warnings that disappeared since I instead used the script to run cmake and build in a separate directory.

Looks more like different compiler versions used.
Anyway it's a bug in expat and I tested build with gcc7, build is broken with gcc7 due to this bug. I did not find an easy way to fix it without significant expat changes. I prefer to avoid significant changes because it's a 3party so if you want to provide build fix for gcc7 and above I suggest just to comment out EnableUnknownEncodingHandler. If you don't I'll do it.

rdoeffinger:
Updated the patch/branch, but feel free to commit your own variant instead!

PR #8879, opened 2018-06-18
[android] Added initialization filter controller on tablets after dis…

yoksnod:
…covery

https://jira.mail.ru/browse/MAPSME-7837


yoksnod:
@alexzatsepin PTAL

PR #8899, opened 2018-06-19
Warnfix

rdoeffinger:
Not sure you like this way of fixing them, but it would be nice to not have these warnings in gcc 7.3.

### DIFF
--- a/generator/osm_element.cpp
+++ b/generator/osm_element.cpp
@@ -28,6 +28,7 @@ std::string DebugPrint(OsmElement::EntityType e)
     case OsmElement::EntityType::Member:
       return "member";
   }
+  return "unknown";
 }
 
 
--- a/indexer/cell_coverer.hpp
+++ b/indexer/cell_coverer.hpp
@@ -169,8 +169,8 @@ void CoverSpiral(m2::RectD rect, int maxDepth, vector<CellId> & result)
   };
 
   auto const coordsAreValid = [](pair<int32_t, int32_t> const & xy) {
-    return xy.first >= 0 && xy.second >= 0 && xy.first <= CellId::MAX_COORD &&
-           xy.second <= CellId::MAX_COORD;
+    return xy.first >= 0 && xy.second >= 0 && xy.first <= static_cast<int32_t>(CellId::MAX_COORD) &&
+           xy.second <= static_cast<int32_t>(CellId::MAX_COORD);
   };
 
   m2::RectD coveredRect;
--- a/kml/serdes.cpp
+++ b/kml/serdes.cpp
@@ -120,8 +120,9 @@ std::string GetStyleForPredefinedColor(PredefinedColor color)
   case PredefinedColor::Orange: return "placemark-orange";
   case PredefinedColor::None:
   case PredefinedColor::Count:
-    return {};
+    break;
   }
+  return {};
 }
   
 BookmarkIcon GetIcon(std::string const & iconName)
--- a/search/search_quality/assessment_tool/main_model.cpp
+++ b/search/search_quality/assessment_tool/main_model.cpp
@@ -115,7 +115,7 @@ void MainModel::OnSampleSelected(int index)
   CHECK(m_threadChecker.CalledOnOriginalThread(), ());
 
   CHECK_GREATER_OR_EQUAL(index, 0, ());
-  CHECK_LESS(index, m_contexts.Size(), ());
+  CHECK_LESS(static_cast<unsigned>(index), m_contexts.Size(), ());
   CHECK(m_view, ());
 
   m_selectedSample = index;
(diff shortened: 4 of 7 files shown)


generator/osm_element.cpp
```diff
@@ -28,6 +28,7 @@ std::string DebugPrint(OsmElement::EntityType e)
     case OsmElement::EntityType::Member:
       return "member";
   }
+  return "unknown";
```
mpimenov: Please add an assertion here. Something like
```
auto const eInt = static_cast<size_t>(OsmElement::EntityType);
string const msg = "Unknown OsmElement::EntityType " + strings::toString(eInt);
ASSERT(false, (msg));
return msg;
```

indexer/cell_coverer.hpp
```diff
@@ -169,8 +169,8 @@ void CoverSpiral(m2::RectD rect, int maxDepth, vector<CellId> & result)
   };
 
   auto const coordsAreValid = [](pair<int32_t, int32_t> const & xy) {
-    return xy.first >= 0 && xy.second >= 0 && xy.first <= CellId::MAX_COORD &&
-           xy.second <= CellId::MAX_COORD;
+    return xy.first >= 0 && xy.second >= 0 && xy.first <= static_cast<int32_t>(CellId::MAX_COORD) &&
```
mpimenov: I'd rather cast xy.first to uint32_t here.

kml/serdes.cpp
```diff
@@ -120,8 +120,9 @@ std::string GetStyleForPredefinedColor(PredefinedColor color)
   case PredefinedColor::Orange: return "placemark-orange";
   case PredefinedColor::None:
   case PredefinedColor::Count:
-    return {};
+    break;
   }
+  return {};
```
mpimenov: I think we need an assertion here too. @rokuz ?

search/search_quality/assessment_tool/main_model.cpp
```diff
@@ -115,7 +115,7 @@ void MainModel::OnSampleSelected(int index)
   CHECK(m_threadChecker.CalledOnOriginalThread(), ());
 
   CHECK_GREATER_OR_EQUAL(index, 0, ());
-  CHECK_LESS(index, m_contexts.Size(), ());
+  CHECK_LESS(static_cast<unsigned>(index), m_contexts.Size(), ());
```
mpimenov: s/unsigned/size_t/

generator/osm_element.cpp
rdoeffinger: Going by the name this is just a "DebugPrint" function, does it really make sense to clutter it with such a verbose assert?
My patch was fairly incomplete (I'll try to upload the more complete version later), if you'd want to have asserts for all such cases that would result in significant code bloat I think.

indexer/cell_coverer.hpp
rdoeffinger: Reasoning was that there is already an assert just below that ensures that this cast is safe.
So this seemed the least risky.
I don't really mind though.

kml/serdes.cpp
rdoeffinger: Well, not for the ::None case I expect.
If for the ::Count and any special cases then maybe it would be better to first change/fix the code, to not have bug fixes mixed in warning fixes?
I admit I did these changes in the assumption that you are relying on the "... not handled in switch" warnings to make sure all cases are updated if the enum is expanded, as runtime errors are really bad for handling this anyway.

rokuz: @mpimenov As I remember we intentionally do not put default 'return' in such methods. It helps to tracks new values in enums.
@rdoeffinger I recommend to put CHECK(false, ("Incorrect enum value")); instead of return {}

mpimenov: @rokuz I thought so too but couldn't find an example. I'm ok with CHECK instead of return.

indexer/cell_coverer.hpp
mpimenov: My reasoning was:
1. This is what happens anyway, an explicit cast to the unsigned type.
2. Imagine that we change MAX_COORD to uint64_t.

As for the assert below, I didn't notice it and I don't even think it is the right way to check what its author wanted to check:
https://pastebin.com/K9ts145N
https://ideone.com/wNg1pa

generator/osm_element.cpp
mpimenov: Why not if the goal is to catch incorrect usage attempts.
See the discussion below, looks like adding CHECKs is the preferred way.

maksimandrianov:
@rdoeffinger Thank you for noticing the warnings. We fixed almost all warnings.
Please read https://github.com/mapsme/omim/pull/8930/

rdoeffinger:
Thanks for taking care of it. I had planned to look into compiler behaviour for the switches to check the best solution but just didn't have time for it.

PR #8900, opened 2018-06-19
Fix dev build path to "data" directory.

rdoeffinger:
The MAPS.ME binary is created in the root of the
build directory, and the "data" symlink is at the
same level.
Though maybe I am just missing some reason why it is like this?
Tested on Linux, don't have OSX to test that for example.

### DIFF
--- a/platform/platform_linux.cpp
+++ b/platform/platform_linux.cpp
@@ -144,8 +144,8 @@ Platform::Platform()
   }
   else
   {
-    string const devBuildWithSymlink = my::JoinPath(path, "..", "..", "data");
-    string const devBuildWithoutSymlink = my::JoinPath(path, "..", "..", "..", "omim", "data");
+    string const devBuildWithSymlink = my::JoinPath(path, "data");
+    string const devBuildWithoutSymlink = my::JoinPath(path, "..", "omim", "data");
     string const installedVersionWithPackages = my::JoinPath(path, "..", "share");
     string const installedVersionWithoutPackages = my::JoinPath(path, "..", "MapsWithMe");
     string const customInstall = path;


rokuz:
I suppose, the reason is running under IDEs. @greshilov check it out pls

platform/platform_linux.cpp
```diff
@@ -144,8 +144,8 @@ Platform::Platform()
   }
   else
   {
-    string const devBuildWithSymlink = my::JoinPath(path, "..", "..", "data");
-    string const devBuildWithoutSymlink = my::JoinPath(path, "..", "..", "..", "omim", "data");
+    string const devBuildWithSymlink = my::JoinPath(path, "data");
+    string const devBuildWithoutSymlink = my::JoinPath(path, "..", "omim", "data");
```
Zverik: Shouldn't it be just `"..", "data"`?

rdoeffinger: If you check the git repo out into "/somepath/omim" then the build script recommended for Linux will generate e.g. "/somepath/omim-build-debug/" and a symlinnk "/somepath/omim-build-debug/data -> /somepath/omim/data". The binary will be generated as " /somepath/omim-build-debug/MAPS.ME".
There will also be a "/somepath/omim-build-debug/MAPS.ME.app/Contents/Resources/", though that seems a bit like a build system bug.
These paths are relative to the binary, so if you do "..", "data", then it would be "/somepath/data", which does not exist and thus will not work.
"data" will get the symlink and work if that exists. "..", "omim" will work if the checkout has the standard name.
"MAPS.ME.app", "Contents", "Resources" should work as well, but seems just wrong for a Linux build.

Zverik: Indeed, I forgot. Thanks.

mapsmetest:

PR #8952, opened 2018-06-27
[coding] [map] Fixed tests.

mpimenov:
Assertions in FileReader interfered with tests that checked
exceptions from reading broken files.


mpimenov:
JTTP

PR #9070, opened 2018-07-11
[search] Prefer objects with higher popularity

tatiana-yan:
cherry-pick https://github.com/mapsme/omim/pull/9069


search/ranking_info.cpp
```diff
@@ -14,6 +14,8 @@ namespace
 // these coeffs correspond to coeffs in a linear model.
 double const kDistanceToPivot = -1.0000000;
 double const kRank = 1.0000000;
+// todo: (@t.yan) Adjust.
+double const kPopularity = 1.0000000;
```
ygorshenin: Wow, wow. If you would like to adjust it later, correct value for now should be zero.

PR #9084, opened 2018-07-12
[android] Added ssl error handling during loading bookmark catalog

alexzatsepin:


android/src/com/mapswithme/maps/bookmarks/BookmarksCatalogFragment.java
```diff
@@ -185,6 +204,7 @@ public void onReceivedError(WebView view, WebResourceRequest request, WebResourc
       UiUtils.hide(frag.mWebView, frag.mProgressView);
       if (ConnectionState.isConnected())
       {
+        LOGGER.e(TAG, "Failed to load catalog: " + mError.toString());
```
yoksnod: Можно и без toString тут

PR #9099, opened 2018-07-16
[iOS] remove weird placeholder from languages section on poi edit screen

beloal:


mapsmetest:

PR #9137, opened 2018-07-18
[iOS] fix category info cell expanded state

beloal:
There was a problem with info cell on bookmarks list screen. If you tap "more" on description cell, scroll to bottom and back to info cell, description appears collapsed.
This PR fixes the bug.


iphone/Maps/Bookmarks/BookmarksVC.mm
```diff
@@ -38,6 +38,8 @@ @interface BookmarksVC() <MWMLocationObserver, MWMCategoryInfoCellDelegate>
   vector<Section> m_sections;
 }
 
+@property (nonatomic) BOOL infoExpanded;
```
vmihaylenko: лишний пробел

PR #9166, opened 2018-07-23
[cmake] Fix search dependency 

greshilov:


greshilov:
JTALL

mpimenov:
What is the commit that broke it and why did it work before?

greshilov:
It happened after merging `release-83` and `master` (PR)(https://github.com/mapsme/omim/pull/9157).

greshilov:
Looks like smth here: [commit](https://github.com/mapsme/omim/commit/a8d38bb7c90a19585933ae42f9b7c1ee69cec26b)

greshilov:
JTALL

track_analyzing/track_analyzer/CMakeLists.txt
```diff
@@ -21,29 +21,22 @@ omim_link_libraries(
...
-  map
   routing
   traffic
   routing_common
   storage
-  kml
-  editor
   indexer
   platform
   mwm_diff
```
syershov: Надо?

greshilov: Да, в первом случае `generator` тянет из `storage`:
```
generator::mwm_diff::ApplyDiff
```
А во втором из `libmwm_diff`:
```
divsuf::divsufsort_include_empty
```

search/pysearch/bindings.cpp
```diff
@@ -186,7 +186,7 @@ struct Context
 {
   Context() : m_engine(m_dataSource, CreateCountryInfoGetter(), search::Engine::Params{}) {}
 
-  DataSource m_dataSource;
+  FrozenDataSource m_dataSource;
```
mpimenov: add a note or todo here, please

mpimenov:
I'm ok with this as a temporary fix.
It looks like a better one would be to add dead_strip or something similar to the linker flags (see https://github.com/mapsme/omim/blob/05ae6caa244c64c9c0b9dac6fb5ee0ce7394e6b2/common.pri)

mpimenov:
The root cause is fixed in #9177

PR #9285, opened 2018-08-10
Fixed transfer to not follow mode, improved lose location logic

rokuz:


drape_frontend/my_position_controller.cpp
```diff
@@ -471,10 +523,13 @@ void MyPositionController::OnLocationUpdate(location::GpsInfo const & info, bool
       }
     }
   }
+  else if (m_mode == location::NotFollowNoPosition)
+  {
+    // Here we silently get the position and go to NotFollow mode.
+    ChangeMode(location::NotFollow);
```
darina: в роутинге надо возвращать в FollowAndRotate

PR #9366, opened 2018-08-28
[desktop] add comma in widget for desktop

gmoryes:
image

In `lat lon` row


tatiana-yan:
может между lat и lon тоже?

gmoryes:
@tatiana-yan between `lat` and `lon` comma too

PR #9398, opened 2018-09-04
[search] Fix Boundaries limit rect in geocoder

tatiana-yan:


mpimenov:
Add comment to class CityBoundary?

search/cities_boundaries_table.hpp
```diff
@@ -55,12 +55,6 @@ class CitiesBoundariesTable
       {
         for (auto const & p : boundary.m_bbox.Points())
           rect.Add(p);
```
mpimenov: nit: adding Min() and Max() is enough
Alternatively, add bbox.ToRect() (since the points are created anyway) or add ForEachCorner() that would avoid creating a vector.

tatiana-yan:
> Add comment to class CityBoundary?
ok, done

indexer/city_boundary.hpp
```diff
@@ -11,6 +11,9 @@
 
 namespace indexer
 {
+
+// Extrapolates city boundary.
```
mpimenov: s/Extrapolates/Approximates/
or Approximate

indexer/city_boundary.hpp
```diff
@@ -11,6 +11,9 @@
...
+// City points are inside boxes intersection.
```
mpimenov: inside the intersection of boxes

search/cities_boundaries_table.hpp
tatiana-yan: Min(), Max() done

PR #9404, opened 2018-09-05
[search] Add sanity checks, update ranking info coeffs.

tatiana-yan:


mapsmetest:

PR #9464, opened 2018-09-14
Mapsme 8404 immersive tips and tricks

yoksnod:
https://github.com/sjwall/MaterialTapTargetPrompt/pull/135


yoksnod:
@alexzatsepin PTAL

PR #9531, opened 2018-09-26
[android] Added subscription deactivation if sdk returns empty purcha…

alexzatsepin:
…se list
https://jira.mail.ru/browse/MAPSME-8641


alexzatsepin:
@yoksnod PTAL

PR #9533, opened 2018-09-26
[android] Fixed crash during show alert dialog on payment failure

alexzatsepin:
[android] Refactore AlertDialog class to use builder to reduce input parameters
https://jira.mail.ru/browse/MAPSME-8644


alexzatsepin:
@yoksnod PTAL

PR #9546, opened 2018-09-27
[android] Delayed core initialization depending on advertising id obt…

alexzatsepin:
…ainting
https://jira.mail.ru/browse/MAPSME-8652


alexzatsepin:
@yoksnod PTAL

PR #9557, opened 2018-09-29
translate all string to persian

herotux:
all strings translated and checked again .every things is good .
but i cant run script if you can please do this to generate fa localization files thanks alot.


mapsmetest:

herotux:
my good freind @mnameghi have a better pull request please merge that: #9558

PR #9563, opened 2018-10-01
[android] Added Mytracker configure location methods call

yoksnod:
https://jira.mail.ru/browse/MAPSME-8675


yoksnod:
@alexzatsepin  PTAL

PR #9627, opened 2018-10-07
add persian language

herotux:
add localization file for persian


herotux:
Someone check this please

PR #9789, opened 2018-11-03
Remove duplicates from strings.xml

dimqua:


rokuz:
Thanks, we've fixed it already in https://github.com/mapsme/omim/pull/9768

PR #9824, opened 2018-11-13
[android] Fixed crash - response is non utf8 string (html)

yoksnod:


yoksnod:
Jni can not  process non UTF-8 string as log message. Fix will be later

PR #9842, opened 2018-11-15
[iOS] handle pending transactions and errors

beloal:
https://jira.mail.ru/browse/MAPSME-8854


map/purchase.hpp
```diff
@@ -42,7 +42,7 @@ class Purchase
     std::string m_vendorId;
     std::string m_receiptData;
 
-    bool IsValid() const { return !m_serverId.empty() && !m_vendorId.empty() && !m_receiptData.empty(); }
+    bool IsValid() const { return !m_vendorId.empty() && !m_receiptData.empty(); }
```
rokuz: Add comment please "Here we don't check serverId, because it can be empty in some cases."

PR #9889, opened 2018-11-22
[search] Pass viewport as is for list search to be consistent with viewport search. Use boost::optional for m_position.

tatiana-yan:


tatiana-yan:
JTALL

PR #9910, opened 2018-11-26
[geocoder] Improvements.

mpimenov:
* Keeping only the top results (this is not a proper beam search
  yet but we probably will still need the data structure in the future).
* Fixed the parent-child layer check.
* A separate code path to index the buildings.

### DIFF
--- a/geocoder/beam.cpp
+++ b/geocoder/beam.cpp
@@ -1,0 +1,31 @@
+#include "geocoder/beam.hpp"
+
+#include "base/assert.hpp"
+
+#include <algorithm>
+
+namespace geocoder
+{
+Beam::Beam(size_t capacity) : m_capacity(capacity) { m_entries.reserve(m_capacity); }
+
+void Beam::Add(Key const & key, Value value)
+{
+  if (m_capacity == 0)
+    return;
+
+  Entry const e(key, value);
+  auto it = std::lower_bound(m_entries.begin(), m_entries.end(), e);
+
+  if (it == m_entries.end())
+  {
+    if (m_entries.size() < m_capacity)
+      m_entries.emplace_back(e);
+    return;
+  }
+
+  if (m_entries.size() == m_capacity)
+    m_entries.pop_back();
+
+  m_entries.insert(it, e);
+}
+}  // namespace geocoder
--- a/geocoder/geocoder.cpp
+++ b/geocoder/geocoder.cpp
@@ -1,10 +1,14 @@
 #include "geocoder/geocoder.hpp"
 
+#include "search/house_numbers_matcher.hpp"
+
 #include "indexer/search_string_utils.hpp"
 
 #include "base/assert.hpp"
 #include "base/logging.hpp"
 #include "base/scope_guard.hpp"
+#include "base/stl_helpers.hpp"
+#include "base/string_utils.hpp"
 #include "base/timer.hpp"
 
 #include <algorithm>
@@ -14,6 +18,20 @@ using namespace std;
 
 namespace
 {
+size_t const kMaxResults = 100;
+
+map<geocoder::Type, double> const kWeight = {
+    {geocoder::Type::Country, 10.0},
+    {geocoder::Type::Region, 5.0},
+    {geocoder::Type::Subregion, 4.0},
+    {geocoder::Type::Locality, 3.0},
+    {geocoder::Type::Suburb, 3.0},
+    {geocoder::Type::Sublocality, 2.0},
+    {geocoder::Type::Street, 1.0},
+    {geocoder::Type::Building, 0.1},
+    {geocoder::Type::Count, 0.0},
+};
+
 // todo(@m) This is taken from search/geocoder.hpp. Refactor.
 struct ScopedMarkTokens
 {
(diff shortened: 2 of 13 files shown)


geocoder/hierarchy.cpp
```diff
@@ -81,10 +84,19 @@ void Hierarchy::Entry::DeserializeFromJSONImpl(json_t * const root, string const
   else if (m_nameTokens != m_address[static_cast<size_t>(m_type)])
   {
     ++stats.m_mismatchedNames;
-    LOG(LDEBUG, ("Hierarchy entry name is not the most detailed field in its address:", jsonStr));
   }
 }
 
+bool Hierarchy::Entry::IsParentTo(Hierarchy::Entry const & e) const
+{
+  for (size_t i = 0; i < static_cast<size_t>(geocoder::Type::Count); ++i)
```
syershov: Maybe add constexpr function? Something like `geocoded::get_types_count` ?

mpimenov: We don't usually do this.

geocoder/geocoder.cpp
```diff
@@ -116,23 +138,33 @@ bool Geocoder::Context::AllTokensUsed() const { return m_numUsedTokens == m_toke
...
 
 void Geocoder::Context::FillResults(vector<Result> & results) const
 {
   results.clear();
-  results.reserve(m_results.size());
-  for (auto const & e : m_results)
-    results.emplace_back(e.first /* osmId */, e.second /* certainty */);
+  results.reserve(m_beam.GetEntries().size());
+
+  set<decltype(m_beam)::Key> seen;
```
tatiana-yan: include set

mpimenov: done

geocoder/hierarchy.cpp
```diff
@@ -145,4 +163,27 @@ vector<Hierarchy::Entry> const * const Hierarchy::GetEntries(
...
+
+      size_t const t = static_cast<size_t>(Type::Street);
+      auto const * streetCandidates = GetEntries(be.m_address[t]);
+      if (streetCandidates == nullptr)
+        continue;
+
+      for (auto & se : *streetCandidates)
+      {
+        if (se.IsParentTo(be))
+          se.m_buildingsOnStreet.emplace_back(&be);
```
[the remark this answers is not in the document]
mpimenov: Да, так намного лучше. Спасибо.
done

mpimenov:
@tatiana-yan PTAL

geocoder/beam.cpp
```diff
@@ -0,0 +1,31 @@
...
+{
+Beam::Beam(size_t capacity) : m_capacity(capacity) { m_entries.reserve(m_capacity); }
+
+void Beam::Add(Key const & key, Value value)
+{
+  if (m_capacity == 0)
+    return;
+
+  Entry const e(key, value);
+  auto it = std::lower_bound(m_entries.begin(), m_entries.end(), e);
```
ygorshenin: BTW, entries are sorted, why not to use limited_priority_queue?

geocoder/geocoder.cpp
```diff
@@ -172,39 +205,92 @@ void Geocoder::Go(Context & ctx, Type type) const
...
+        continue;
 
-        double const certainty =
-            static_cast<double>(ctx.GetNumUsedTokens()) / static_cast<double>(ctx.GetNumTokens());
+      ScopedMarkTokens mark(ctx, type, i, j + 1);
 
-        for (auto const * e : curLayer.m_entries)
-          ctx.AddResult(e->m_osmId, certainty);
+      // double const certainty =
+      //     static_cast<double>(ctx.GetNumUsedTokens()) /
```
ygorshenin: Why this code is commented?

geocoder/beam.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <algorithm>
+
+namespace geocoder
+{
+Beam::Beam(size_t capacity) : m_capacity(capacity) { m_entries.reserve(m_capacity); }
+
+void Beam::Add(Key const & key, Value value)
+{
+  if (m_capacity == 0)
```
ygorshenin: Strongly suggest to mark this check as UNLIKELY

geocoder/geocoder.cpp
```diff
@@ -1,19 +1,38 @@
...
 #include <utility>
 
 using namespace std;
 
 namespace
 {
+size_t const kMaxResults = 100;
+
+map<geocoder::Type, double> const kWeight = {
+    {geocoder::Type::Country, 10.0},
```
ygorshenin: Why don't you want to use direct address hashing?

geocoder/geocoder.cpp
```diff
@@ -52,25 +71,29 @@ geocoder::Type NextType(geocoder::Type type)
...
   {
-    for (auto const * pe : layer.m_entries)
-    {
-      // Note that the relationship is somewhat inverted: every ancestor
-      // is stored in the address but the nodes have no information
-      // about their children.
-      if (e.m_address[static_cast<size_t>(pe->m_type)] == pe->m_nameTokens)
-        return true;
-    }
+    // Note that the relationship is somewhat inverted: every ancestor
```
ygorshenin: The name of the function is misleading. HasPrent?

geocoder/geocoder.cpp
```diff
@@ -172,39 +205,92 @@ void Geocoder::Go(Context & ctx, Type type) const
...
+      //     static_cast<double>(ctx.GetNumTokens());
 
-        ctx.GetLayers().emplace_back(move(curLayer));
-        SCOPE_GUARD(pop, [&] { ctx.GetLayers().pop_back(); });
+      double certainty = 0;
+      for (auto const t : ctx.GetTokenTypes())
+      {
+        auto const it = kWeight.find(t);
+        CHECK(it != kWeight.end(), ());
+        certainty += it->second;
```
ygorshenin: Am I right that certainity is a log prob?

geocoder/beam.cpp
mpimenov: Yes, it is possible. See further requests.

geocoder/beam.cpp
mpimenov: ok
will do in a later PR

geocoder/geocoder.cpp
mpimenov: Good point, will fix.

geocoder/geocoder.cpp
mpimenov: No.
certainty is described in geocoder/result.hpp
It is just a number in [0.0, 1.0] at this stage, no more guarantees are given.

geocoder/geocoder.cpp
mpimenov: HasParent is better, will fix

geocoder/geocoder.cpp
ygorshenin: But why not to transform it to log-prob? It'll have clear semantic.

mpimenov: Can you explain what you mean in more detail?
Here is what certainty is now:
1. Let's see how many query tokens are covered with the current parse and assign this fraction to certainty.
2. Turns out, it works badly since a single matched long street in the query (i.e., wrong city, wrong region, wrong locality, correct street) can shadow a more relevant result (correct city, correct locality, wrong street) in the case where we do not have good results in our database anyway. So let's make some parts of the query heavier (heuristically). This turns out to work more predictable.

geocoder/geocoder.cpp
mpimenov: See the comment about certainty below.
I've removed the commented code for now.

geocoder/geocoder.cpp
mpimenov: I added this discussion as the documentation to GetWeight in https://github.com/mapsme/omim/pull/10815

PR #9937, opened 2018-11-28
[geocoder] Dropping street synonyms when indexing.

mpimenov:


geocoder/hierarchy.cpp
```diff
@@ -199,6 +205,22 @@ void Hierarchy::IndexEntries()
...
+{
+  CHECK_EQUAL(e.m_type, Type::Street, ());
+  size_t const t = static_cast<size_t>(e.m_type);
+  m_entriesByTokens[MakeIndexKey(e.m_address[t])].emplace_back(&e);
+
+  for (size_t i = 0; i < e.m_address[t].size(); ++i)
+  {
+    if (!search::IsStreetSynonym(strings::MakeUniString(e.m_address[t][i])))
+      continue;
+    auto addr = e.m_address[t];
```
syershov: Possible to move this string outside the loop?

mpimenov: I don't think so, see the next line.

PR #9955, opened 2018-11-29
[android] Added global purchase validation observer to avoid collisti…

alexzatsepin:
https://jira.mail.ru/browse/MAPSME-8889


alexzatsepin:
@yoksnod PTAL

PR #9975, opened 2018-11-30
[iOS] fix deeplinks to catalog

beloal:
Do not require author name in catalog deeplink


iphone/Maps/Bookmarks/Catalog/CatalogWebViewController.swift
```diff
@@ -1,17 +1,16 @@
...
 
   init?(_ components: [String : String]) {
     guard let id = components["id"],
-      let name = components["name"],
-      let author = components["author_name"] else { return nil }
+      let name = components["name"] else { return nil }
     self.id = id
     self.name = name
-    self.author = author
+    author = components["author_name"];
```
kadetlessy: remove ;

PR #10056, opened 2018-12-12
[coding] Renamed the StringUtf8Multilang files.

mpimenov:
Also added some documentation on how StringUtf8Multilang is stored.

### DIFF
--- a/coding/transliteration.cpp
+++ b/coding/transliteration.cpp
@@ -1,5 +1,6 @@
 #include "coding/transliteration.hpp"
-#include "coding/multilang_utf8_string.hpp"
+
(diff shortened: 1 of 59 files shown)


mpimenov:
JTA

coding/transliteration.cpp
```diff
@@ -10,9 +11,8 @@
 #include "3party/icu/i18n/unicode/translit.h"
 #include "3party/icu/i18n/unicode/utrans.h"
 
-#include "std/unique_ptr.hpp"
-
 #include <cstring>
+#include <memory>
```
tatiana-yan: есть в hpp

mpimenov: done

PR #10065, opened 2018-12-13
[android] Modified events and params

yoksnod:


alexzatsepin:
@yoksnod cherry-pick on 85?

PR #10086, opened 2018-12-17
[search] Use HasName feature instead of ErrorsMade and NameScores for  categorial requests.

tatiana-yan:


search/ranking_info.hpp
```diff
@@ -48,6 +48,12 @@ struct RankingInfo
   // tokens are categorial ones.
   bool m_falseCats = false;
 
+  // True iff request is categorial.
```
mpimenov: s/request/the request/

search/ranking_info.hpp
```diff
@@ -48,6 +48,12 @@ struct RankingInfo
...
+  bool m_categorialRequest = false;
+
+  // True iff feature has name.
```
mpimenov: iff the feature has a name

tatiana-yan: done

search/ranking_info.hpp
tatiana-yan: done

tatiana-yan:
@mpimenov PTAL

PR #10096, opened 2018-12-18
[search] Emit every |batchSize| results.

tatiana-yan:


search/ranker.cpp
```diff
@@ -540,7 +540,11 @@ void Ranker::UpdateResults(bool lastUpdate)
     }
 
     if (!lastUpdate)
+    {
       BailIfCancelled();
+      if (count % m_params.m_batchSize == 0)
```
mpimenov: // For categorial requests, it is usual that several batches arrive
// in the first call to UpdateResults(). Emit them as soon as they are available
// to improve responsiveness.

tatiana-yan:
@mpimenov PTAL

PR #10157, opened 2018-12-26
[search][generator] Store corresponding street feature id as street i…

tatiana-yan:
…dentifier in HouseToStreetTable.


tatiana-yan:
JTALL

PR #10202, opened 2019-01-15
Remove requirement to build from a git checkout

vicpopov:
Sources can be obtained via download of tarball from github,
and in that case git rev-parse is not possible.
Also, every path in script is not dependent on current workdir.


bykoianko:
JTA

PR #10224, opened 2019-01-18
[routing] Add test for route=shuttle_train

gmoryes:
image


routing/routing_integration_tests/route_test.cpp
```diff
@@ -399,4 +399,20 @@ namespace
     Route const & route = *routeResult.first;
     integration::TestRouteTime(route, 17704.3);
   }
+
+  // Test on rouads with tag route=shuttle_train
```
bykoianko: `// Test on roads with tag route=shuttle_train`

gmoryes: done

PR #10254, opened 2019-01-25
[routing] m2::Point to LatLon in routing_integration_tests

gmoryes:


gmoryes:
JTRIT

mapsmetest:

PR #10275, opened 2019-01-29
[ios][mapsme-9511] Limited number of tags that the user can choose to share category

kadetlessy:
https://jira.mail.ru/browse/MAPSME-9511


beloal:
https://github.com/mapsme/omim/pull/10361

PR #10312, opened 2019-02-07
Update Hungarian translations

debyos:


mapsmetest:

debyos:
Done.

Gábor

mpimenov <notifications@github.com> ezt írta (időpont: 2019. febr. 7., Cs,
23:24):

data/categories.txt
```diff
@@ -1495,7 +1495,7 @@ nl:^Computerwinkel
 fi:^Tietokonekauppa
 fr:^Magasin d'informatique
 de:^Computergeschäft
-hu:^Számítógép üzlet
+hu:^Számítógépüzlet
```
mpimenov: Are you sure that we don't need the version with a space?
We currently do not insert spaces automatically.

data/categories.txt
```diff
@@ -1911,7 +1911,7 @@ nl:3^Metro
 fi:3^Metro
 fr:3^Métro
 de:2^U-Bahn|Metro|ÖPNV
-hu:3^Földalatti|3metro
+hu:3^Metró|Földalatti|3metro
```
mpimenov: Fill free to drop the last one: diacritics are ignored when looking for synonyms.

data/categories.txt
```diff
@@ -1962,7 +1962,7 @@ nl:3^Veerpont|veerboot|veer|pont|overzetboot
 fi:3^Lautta|terminaali|kuljetus
 fr:3^Traversier|terminal
 de:2^Fähre|Terminal
-hu:2^Komp
+hu:2^Kompkikötő
```
mpimenov: Is it really so rare (or important) to start suggesting it after typing "Ko" (2 characters)?

data/categories.txt
```diff
@@ -2259,7 +2259,7 @@ nl:^Kerk|tempel|kathedraal|gebedsplaats|gebedshuis
 fi:^Kirkko|uskonnollinen kohde
 fr:^Église|lieu de culte|temple|basilique
 de:^Kirche|Dom|3Kathedrale|4Münster|Anbetungsstätte|Tempel
-hu:^Templom
+hu:^Templom|Székesegyház|Katedrális|Dóm|Bazilika|Istentisztelei hely
```
mpimenov: Should the last one be "Istentiszteleti hely"?

data/categories.txt
```diff
@@ -2831,7 +2831,7 @@ el:^Παραχώρηση αυτοκινήτου
 es:^Compartir coche
 fi:^Autojen yhteiskäyttö
 fr:^Car Sharing
-hu:^Gépjármű megosztása
+hu:^Gépjárműmegosztás
```
mpimenov: Same here: should we leave ```Gépjárműmegosztás|Gépjármű megosztás```?

data/categories.txt
```diff
@@ -3797,7 +3797,7 @@ nl:3^Toilet
 fi:^WC
 fr:3^Toilettes
 de:3^Toilette
-hu:3^Mosdó
+hu:3^WC
```
mpimenov: Please remove the "3".

data/categories.txt
```diff
@@ -5378,7 +5378,7 @@ nl:3^Auto reparatie|auto|wagen
 fi:^Autokorjaamo
 fr:3^Réparation d'automobiles|3garage|4station service|voiture|auto
 de:3^Autowerkstatt|Kfz|Reparaturwerkstatt|Auto|Werkstatt
-hu:3^Gépkocsi szervíz|4javítóműhely|kocsi
+hu:3^Autószerelő|Autószerviz|Gépkocsiszerviz|Autójavító|4javítóműhely|kocsi
```
mpimenov: Should we really remove the old synonym?

data/categories.txt
```diff
@@ -9228,7 +9228,7 @@ el:^Καλύβα|αγροικία
 es:^Cabaña
 fi:^Autiotupa
 fr:^Cabane sauvage
-hu:^Kunyhó a vadonban
+hu:^Menedékház (személyzet nélkül)
```
mpimenov: Note that this file contains synonyms for search: we will suggest the category if a user types in one of the strings separated by "|". Here someone needs to type in (almost exactly) ```Menedékház (személyzet nélkül)``` which I think is unlikely. Please try to find a shorted synonym: what would you type in the search bar if you wanted to find a wilderness hut?

mpimenov:
You may also be interested in taking a look at data/strings/strings.txt: https://github.com/mapsme/omim/blob/master/data/strings/strings.txt

data/categories.txt
debyos: Yes, I am sure.
In Hungarian, if you have a composed noun which is made of two parts or more than two parts but is at most 6 sillables long, it must be written together.

data/categories.txt
debyos: OK, I will drop it (or you can do as well).

data/categories.txt
debyos: No, sorry, I just took over from the previous version.

data/categories.txt
debyos: Yes, you're right. Thanks for spotting!

data/categories.txt
debyos: The same rule here like above (this is the so-called 6:3 rule: 6 sillables and 3 parts).

data/categories.txt
debyos: OK, I will do it.

data/categories.txt
debyos: It is there but written correctly: _Gépkocsiszerviz_ (the third expression, again, written without spaces)

data/categories.txt
debyos: It can be just _menedékház_ ("személyzet nélkül" means "without personnel" which is not indispensable)

PR #10339, opened 2019-02-13
No issue android studio 3.1.1

yoksnod:
Fabric plugin produce the similar issue like here https://github.com/firebase/firebase-android-sdk/issues/198#issuecomment-456455002
It will be fixed later


yoksnod:
JTANDROID

PR #10340, opened 2019-02-13
[storage] Removing the deleted country from DiffManager.

mpimenov:
Suppose that a user has one mwm and the app has been updated.
It now offers to download the diff, however instead
the user deletes the map and tries to download it again.
The download fails because according to the DiffManager it's
not the mwm but the mwmdiff that should be downloaded which
results in a size mismatch. Moreover, if the download is retried,
the diff is downloaded (because the MapOptions are correct now)
but it cannot be applied because the old mwm has already been removed.

The problem is that we do not account for the change in mwm status
since the start of the application (and, in particular, the DiffManager)
when taking the decision whether to download a diff or a full mwm.

### DIFF
--- a/storage/storage.cpp
+++ b/storage/storage.cpp
@@ -564,6 +564,8 @@ void Storage::DeleteCountry(CountryId const & countryId, MapOptions opt)
   bool const deferredDelete = m_willDelete(countryId, localFile);
   DeleteCountryFiles(countryId, opt, deferredDelete);
   DeleteCountryFilesFromDownloader(countryId);
+  m_diffManager.RemoveDiffForCountry(countryId);
+
   NotifyStatusChangedForHierarchy(countryId);
 }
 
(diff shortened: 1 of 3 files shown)


storage/storage.cpp
```diff
@@ -565,6 +565,8 @@ void Storage::DeleteCountry(CountryId const & countryId, MapOptions opt)
   DeleteCountryFiles(countryId, opt, deferredDelete);
   DeleteCountryFilesFromDownloader(countryId);
   NotifyStatusChangedForHierarchy(countryId);
+
+  m_diffManager.RemoveDiffForCountry(countryId);
```
milchakov: опционально. может может поменять местами 567 и 569? чтоб все методы удаления были рядом.

mpimenov:
JTA

mpimenov:
JTA

PR #10433, opened 2019-02-27
Categories.txt improved Dutch and German translations

demasterr:
Improved the translations for Dutch and German by adding new translations, fixing some wrong translations, and reordering some of the translations.

*Below the most important remarks of changes are listed. The changes with warning signs may need an extra review or contain a question.*

A few remarks for the Dutch translations:
- The earlier  used 'autodelen' for 'car sharing' is probably not the most common phrase in Dutch. 'Carpool' is a better translation, as this is the known concept for sharing your car.
- Hospital received an extra synonym for Belgian people. In the Netherlands (Holland) this word is not used, however, Belgian people may use 'hospitaal' as a synonym for 'hospital'.
- The translation of 'Dumpster|trash' was now translated in Dutch as 'uitschot'. This word is more used for calling a person trash, not so much for throwing away stuff. We don't want people to get the impressions that there are some trash people at a location ;)
- For hospital there was the Dutch translation 'tandarts', which is a 'dentist'. This probably shouldn't belong here.
- ⚠️ 'County' was translated as 'graafschap' for Dutch. This is a very old translation for it and will not be used in normal day usage. 'Pronvincie' is a better translation here. The word may be used differently in other countries though, which would mean that this translation may then be wrong.
- Island was translated in Dutch as 'eilandje', which would mean that it is only used for small islands. 'Eiland' is the right translation for all sizes.
- ⚠️ Racetrack was changed to 'renbaan' for Dutch. This is a better translation UNLESS racetrack is also used for cars. 'Renbaan' is only used for animal racing. So the question is, is racetrack used for animals only or also for cars?
- ⚠️ The German translation of Hvac|Air conditioner seems a bit off. Heizung translates to heater. Air conditioner directly translated gives Klimaanlage. However, I'm not German myself so I cannot fully guarantee that this is the right translation for this point.
- Nursery|child care was in German translated to Kinderzimmer, which is a child's room
- A defibrillator is in Dutch called an AED. I don't expect most Dutch people to know the word 'defibrillator'. Though, defibrillator is an accepted Dutch word, so it is still valid here.
- ⚠️ Variety Store did not have a Spanish translation. It now contained `<Todo a cien>`. I replaced it with the Google Translate suggestion. However, this translation could be wrong! Anyhow, it can never be any worse than seeing `<Todo a cien>` in your GUI.
EDIT: at least one Spanish person confirmed this translation for me.

Could someone explain the usage of `place-locality`? I believe the Dutch translation is wrong here. However, I do think the current translation is wrong.

### DIFF
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -2189,7 +2189,7 @@ ru:^Место для пикника|3Пикник
 ar:^مكان نزهة|سياحة|مناظر
 cs:3^Piknik|picnic|picnik
 da:3^Picnic
-nl:3^Picnicplaats|3picnicweide|3picnictafel
+nl:3^Picknickplaats|picknickweide|picknicktafel
 fi:3^Piknikpaikat
 fr:3^Terrain de pique-nique|3aire de pique-nique
 de:3^Picknickplatz|Grillplatz
@@ -2707,15 +2707,20 @@ fa:خرابه های تاریخی
 historic-ship|@tourism
 en:^Ship|boat
 ru:^Корабль|судно|лодка
+nl:^Schip|boot
+de:^Schiff|Boot
 
 historic-tomb|@tourism
 en:^Tomb|grave|memorial
 ru:^Гробница|склеп|захоронение|могила|кладбище
+nl:^Tombe|graf|gedenkteken
+de:^Grab|Gedenkstätte
 
 leisure-dog_park
 en:3^Dog area|Dog park
 ru:^Место для выгула собак|Площадка для собак|4Собачья площадка
 cs:3^Psí hřiště
+nl:4^Hondenpark|Hondengebied
 de:4^Hundezone|Hundeauslaufzone
 fi:Koirapuisto
 he:גינת כלבים
@@ -2837,7 +2842,7 @@ it:^Car Sharing
 ja:^カーシェアリング
 ko:^카셰어링
 nb:^Kjøre sammen
-nl:^Autodelen
+nl:3^Carpool|Carpoolen|Auto delen|Autodelen
 pl:^Carsharing
 pt:^Partilha de carro
 pt-BR:^Compartilhamento de carro
@@ -2858,7 +2863,7 @@ ru:4^Прокат авто|3авто|3машина|прокат|авто нап�
 ar:^تأجير سيارات|سيارة|تأجير|إيجار
 cs:3^Půjčovna aut|auto|pronájem|nájemné
 da:4^Biludlejning|bil
-nl:3^Autoverhuur|autodelen|auto|huur
+nl:3^Autoverhuur|4huurauto|auto delen|auto|huur
 fi:^Autovuokraamo
 fr:4^Location de voitures|auto|loueur|location|partage
 de:3^Autoverleih|Mietauto|Auto|Fahrzeug Mieten|Verleih|KFZ-Vermietung|Fahrzeugvermietung
@@ -2891,7 +2896,7 @@ ru:^Кинотеатр|3кино
 ar:^سينما
 cs:3^Kino|biograf
 da:3^Biograf
-nl:3^Bioscoop|cinema
+nl:3^Bioscoop|filmtheater|cinema
 fi:^Elokuvateatteri
 fr:3^Cinéma
 de:3^Kino|Cinema|Filmtheater|Lichtspielhaus
@@ -2957,7 +2962,7 @@ ru:3^Ночной клуб|3дискотека|ночной бар|диско|т
 ar:^نادي ليلي|رقص
 cs:4^Noční klub|3disco|3klub
 da:3^Natklub|dans|danseklub
-nl:3^Discotheek|3disco|dansen
+nl:3^Discotheek|3disco|dansen|nachtclub
 fi:^Yökerho
 fr:3^Discothèque|3boîte de nuit|danse
 de:3^Nachtclub|Nachtklub|Nachtbar|Nachtlokal|Disko|Tanzen|Club
@@ -3214,7 +3219,7 @@ ru:4^Больница
 ar:^مستشفى
 cs:4^Nemocnice
 da:4^Hospital
-nl:4^Ziekenhuis
+nl:4^Ziekenhuis|hospitaal
 fi:^Sairaala
 fr:4^Hôpital
 de:4^Krankenhaus
@@ -3248,9 +3253,9 @@ ru:4госпиталь|врач|доктор|медицинский центр|�
 ar:^عيادة|طبيب|دكتور
 cs:klinika|pohotovost|zdravotnické centrum|3lékař|doktor
 da:sygehus
-nl:kliniek|hospitaal|tandarts|3dokter
+nl:kliniek|hospitaal|3dokter|3eerste hulp
 fr:clinique|3médecin|3docteur
-de:Spital|3Arzt|Doktor|Medizinische Einrichtung|Medizinisches Zentrum|Gesundheitswesen|Ambulanz|Gesundheitsdienstleistungen|Erste Hilfe
+de:Spital|Klinik|3Arzt|Doktor|Medizinische Einrichtung|Medizinisches Zentrum|Gesundheitswesen|Ambulanz|Gesundheitsdienstleistungen|Erste Hilfe
 hu:rendelőintézet|3orvos
 ja:クリニック|医師|医者|ドクター|救急|診療
 ko:의사|의사가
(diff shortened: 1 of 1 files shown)


mapsmetest:

data/categories.txt
```diff
@@ -3825,7 +3831,7 @@ amenity-toilets|@toilet
 en:wc|5restroom|4bathroom|loo|lavatory|U+1F6BD|U+1F6BE|U+1F4A9|U+1F6BB|U+1F6B9|U+1F6BA
 ar:^حمام|مرحاض
 cs:wc|toalety
-nl:wc
+nl:WC
```
HenkGrent: Consider adding "Toilet"

demasterr: Good one, will add!

mpimenov:
Hello. Please fill in the CLA: https://maps.me/cla/

demasterr:
Done!

data/categories.txt
```diff
@@ -7447,7 +7455,7 @@ ru:^Карта
 ar:^خريطة سياحية
 cs:^Turistická mapa
 da:^Turistkort
-nl:^Toeristische kaart
+nl:^Toeristische kaart|toeristen kaart
```
casperboone: I think this should be 'toeristenkaart'

data/categories.txt
```diff
@@ -7254,10 +7262,10 @@ ru:^Рыбный магазин|рыба|морепродукты|рыбная 
 ar:^سماك
 cs:^Prodej ryb
 da:^Fiskehandler
-nl:^Visboer
+nl:^Visboer|vis markt|vis|zeevoedsel|schelpdier
```
casperboone: I think this should be 'vismarkt'

data/categories.txt
```diff
@@ -6612,7 +6620,7 @@ ru:^Телефон для экстренных вызовов
 ar:^هاتف الطوارئ
 cs:^Tísňového volání
 da:^Nødtelefon
-nl:^Praatpaal
+nl:^Praatpaal|4nood telefoon
```
casperboone: I think this should be 'noodtelefoon'

data/categories.txt
```diff
@@ -6420,7 +6428,7 @@ ru:^Автомойка
 ar:^مغسلة سيارات
 cs:^Myčka aut
 da:^Bilvask
-nl:^Car wash
+nl:^Car wash|auto was
```
casperboone: I think 'autowasstraat' or 'wasstraat' are more appropriate translations

data/categories.txt
```diff
@@ -6231,6 +6238,7 @@ ru:^Копировальный центр|печать|полиграфия
 ar:^محل نسخ|طباعة
 cs:^Kopírovací obchod|Tiskárny
 da:^Kopieringsbutik|Trykkeri
+nl:^Copy shop|Kopieer winkel|printer|drukker
```
casperboone: I think this should be 'kopieerwinkel'

data/categories.txt
```diff
@@ -5407,7 +5414,7 @@ ru:4^Кемпинг
 ar:^تخييم|مكان التخييم|أرض التخييم
 cs:4^Kempování|camping|kemp|camp
 da:^Camping|campingplads|teltplads|lejrplads
-nl:4^Camping|kamperen
+nl:4^Camping|kamperen|kampeer plek
```
casperboone: I think this should be 'kampeerplek'

data/categories.txt
```diff
@@ -3629,7 +3635,7 @@ ru:^Мусорный контейнер|мусорка
 ar:^قمامة|مهملات|سلة مهملات|صندوق قمامة
 cs:^Odpadky|koš|popelnice|uložiště odpadu
 da:^Bortskaffelse af affald|skraldespand|skrald
-nl:^Uitschot|afval|prullenbak|vuilnis
+nl:^Afval container|afval|prullenbak|vuilnis
```
casperboone: I think this should be 'afval container'

demasterr: Thank you!

data/categories.txt
demasterr: Fixed!

data/categories.txt
demasterr: Fixed!

data/categories.txt
demasterr: Yes auto was is probably more slang. I'll take your suggestion 👍

data/categories.txt
demasterr: Yes, fixed!

data/categories.txt
demasterr: Fixed!

data/categories.txt
demasterr: Fixed!

demasterr:
@mpimenov The CLA was signed. Could the PR be reviewed?

data/categories.txt
```diff
@@ -2189,7 +2189,7 @@ ru:^Место для пикника|3Пикник
 ar:^مكان نزهة|سياحة|مناظر
 cs:3^Piknik|picnic|picnik
 da:3^Picnic
-nl:3^Picnicplaats|3picnicweide|3picnictafel
+nl:3^Picknickplaats|3picknickweide|3picknicktafel
```
mpimenov: Are you sure all three synonyms need the "3" prefix? Do we need to suggest all three of them after typing 3 letters?

data/categories.txt
```diff
@@ -5471,7 +5478,7 @@ ru:^Офис|компания|контора|фирма|бизнес|бюро|у
 ar:^مكتب|شركة
 cs:^Kancelář
 da:^Kontor|firma|virksomhed
-nl:^Kantoor|ambt
+nl:^Kantoor|ambt|bedrijf|bedrijspand|bedrijfskantoor
```
mpimenov: s/bedrijspand/bedrijfspand/ ?

data/categories.txt
```diff
@@ -6073,7 +6080,7 @@ da:WiFi|Wi-Fi
 nl:WiFi|Wi-Fi
 fi:WiFi|Wi-Fi
 fr:WiFi|Wi-Fi
-de:WLAN|WiFi|Wi-Fi
+de:WiFi|Wi-Fi|WLAN
```
mpimenov: nit: the order doesn't matter here

mpimenov:
@demasterr I can't see you in our CLA list. Are you sure you have used this github account name?

data/categories.txt
demasterr: You are right, only the first should be suggested. Fixed!

data/categories.txt
demasterr: Fixed!

data/categories.txt
demasterr: Alright, in that case, I'll revert this change.

mpimenov:
Regarding the circumflex: sorry for the confusion, the correct way to make a translation to be shown in the UI is now to edit https://github.com/mapsme/omim/blob/master/data/strings/types_strings.txt

demasterr:
@mpimenov I signed under the GitHub Account `demasterr`. I've signed it again to be sure.

demasterr:
Is categories.txt not used any more? Otherwise it may be wise to delete the file from the master.

I'll have a look if I can move the changes suggested in this PR to https://github.com/mapsme/omim/blob/master/data/strings/types_strings.txt somewhere this week.

mpimenov:
Also, note the all strings shown in the UI should be in this directory, should you want to fix them too: https://github.com/mapsme/omim/blob/master/data/strings

mpimenov:
Please squash all the commits into one.

mpimenov:
We'll look into making it more clear, but for now:
* Strings from the strings/ directory are shown in the UI.
* Strings from the categories* files are used as synonyms for search.

mpimenov:
> I'll have a look if I can move the changes suggested in this PR to https://github.com/mapsme/omim/blob/master/data/strings/types_strings.txt somewhere this week.

The changes here are fine but you may want to repeat some of them there too.

demasterr:
@mpimenov 
> Please squash all the commits into one.

I've no idea what I was doing now. Could you advise me on how to do this?

mpimenov:
git rebase -i
(interactive rebase)

demasterr:
@mpimenov It took some time but I finally managed to do it. Missed the `-f` with the push 🤦‍♂️

demasterr:
I also found one typo while doing the rebase. I had `3Car poolen`. I changed it to `Carpoolen`. The suggestion after 3 characters was also removed, as it was not needed.

mpimenov:
Please change this large commit message to a shorter "Categories.txt improved Dutch and German translations" or similar.

demasterr:
@mpimenov This should be better 👍

PR #10473, opened 2019-03-04
[storage] Fixed a crash.

mpimenov:
Consider the following scenario.
1. Diffs for A and B are downloaded but not applied.
   Currently it is possible because we do not delete a
   diff whose application process was cancelled.
2. The update button is clicked for both A and B, adding
   them both to the queue.
3. The cancel button is clicked for both A and B.

The cancellation of A sends a signal to the diff
manager and kicks the downloader to start applying B.
The diff manager receives the signal to cancel A and
resets the m_latestDiffRequest. Then, the cancel request
to B clears the queue but never reaches the diff manager
(because of the m_latestDiffRequest mismatch), and, once
the diff to B has been successfully applied, Storage cannot
remove B from queue because it has already been removed.

### DIFF
--- a/storage/storage.cpp
+++ b/storage/storage.cpp
@@ -700,6 +700,12 @@ void Storage::DownloadNextFile(QueuedCountry const & country)
   // switch to next file.
   if (isDownloadedDiff || p.GetFileSizeByFullPath(readyFilePath, size))
   {
+    if (m_latestDiffRequest)
+    {
+      // No need to kick the downloader: it will be kicked by
+      // the current diff application process when it is completed or cancelled.
+      return;
+    }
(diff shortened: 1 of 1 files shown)


storage/storage.cpp
```diff
@@ -1604,10 +1610,14 @@ void Storage::ApplyDiff(CountryId const & countryId, function<void(bool isSucces
           break;
           case DiffApplicationResult::Cancelled:
           {
-            CHECK(!IsDiffApplyingInProgressToCountry(countryId), ());
+	    DownloadNextCountryFromQueue();
+          }
+          break;
```
vmihaylenko: добавь, пожалуйста, break внутрь фигурных скобок

mpimenov: done

PR #10501, opened 2019-03-09
[android] Resource leak in ZipLogTask

demasterr:
If an exception would be thrown, the streams would not be closed. The use-try-with-resources will close the resources. Therefore, the `out.close()` has also become unnecessary.
Furthermore, some streams were never closed. All resources will now be closed using the try-with-resources statement.

In my commit messages, I falsely used the words 'memory leak' which should have been 'resource leak'.

⚠️ IMPORTANT: This change assumes the use of Java SE7 or higher.


mapsmetest:

PR #10502, opened 2019-03-09
Fixed possible NullPointerException mSearchController

demasterr:
nSearchController.hide() could throw a NullPointerException. Furthermore, the check later would then be redundant and should be moved earlier.


mapsmetest:

PR #10506, opened 2019-03-11
[android] Refactored banner states to process 'not determined' state …

alexzatsepin:
… more precisely in case when ad is loaded but place page is in not stable state (settling/dragging/hidden)

https://jira.mail.ru/browse/MAPSME-9890

### DIFF
--- a/android/src/com/mapswithme/maps/widget/placepage/BannerController.java
+++ b/android/src/com/mapswithme/maps/widget/placepage/BannerController.java
@@ -40,7 +40,7 @@ import static com.mapswithme.util.statistics.Statistics.PP_BANNER_STATE_DETAILS;
 import static com.mapswithme.util.statistics.Statistics.PP_BANNER_STATE_PREVIEW;
 
 
-final class BannerController
+final class BannerController implements PlacePageStateObserver
 {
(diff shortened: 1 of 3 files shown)


android/src/com/mapswithme/maps/widget/placepage/BannerController.java
```diff
@@ -390,9 +387,9 @@ private void fillViews(@NonNull MwmNativeAd data)
 
   private void animateActionButton()
   {
-    View view = mOpened ? mActionLarge : mActionSmall;
+    View view = isDetailsState(mState) ? mActionLarge : mActionSmall;
```
demasterr: This statement is a bit redundant as you do the same check one if-statement later. I would suggest to just use the if-else below to reduce the complexity of the code. This would mean that animator in the if becomes:
```
 animator = ObjectAnimator.ofObject(mActionLarge, "backgroundColor", new ArgbEvaluator(),
                                         colorFrom, colorTo, colorFrom);
```
and animator in the else becomes
```
animator = ObjectAnimator.ofFloat(mActionSmall, "alpha", 0.3f, 1f);
```

alexzatsepin: Done.

PR #10514, opened 2019-03-11
[android] Removed identical check in equals

demasterr:
Changes:
* In CatalogTag.java the equals method contained two identical evaluations of equals with an OR. Here I assumed the `mId` to be unique and, therefore, no more checks are required. Alternatively, the `mLocalizedName` could be included in the equality, as well as the `mColor`.


mapsmetest:

yoksnod:
@demasterr, thx, it was autogenerated method's mistake

PR #10524, opened 2019-03-12
Unnecessary length check removed

HenkGrent:
Length can not return a value less than 0, therefore this check is not needed.
Some analysis using Sonar Cube revealed this condition check in the code.

I have signed the CLA btw.


mapsmetest:

demasterr:
Maybe SonarQube could be added to the continuous integration of the project

PR #10548, opened 2019-03-18
[iOS] retrun armv7 support

beloal:


mapsmetest:

PR #10551, opened 2019-03-18
[strings] Add types strings for service types

tatiana-yan:


mapsmetest:

PR #10566, opened 2019-03-20
[map] no need to migrate UGC API

milchakov:
В минорку


mapsmetest:

PR #10590, opened 2019-03-26
[android][core][ndk] Added ARM64 support

alexzatsepin:
https://jira.mail.ru/browse/MAPSME-9894

The arm64-v8a support is added in this PR. You have to:
1. Update NDK to r19c version https://developer.android.com/ndk/downloads
2. If build is failed after the first step please update Cmake to last version throught Android Sdk Manager (which is embedded in Android studio or probable you may find it in SDK folder).

Important note: if you need to make build for the specific arch then you should pass the argument to the gradlew command -Parm32, -Parm64 or -Px86 to make your build process faster.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -118,7 +118,7 @@ def getDate() {
 
 def osName = System.properties['os.name'].toLowerCase()
 
-project.ext.versionCodes = ['armeabi-v7a': 1, 'x86': 2]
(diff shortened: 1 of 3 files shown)


alexzatsepin:
JTALL

vmihaylenko:
JTMAC

Alexuser1:
check crashes at MAPSME-10086, no new crashes found.

PR #10670, opened 2019-04-10
[indexer] Rename CachingPopularityLoader to CachingRankTableLoader, c…

tatiana-yan:
…onvert section name to parameter to use the same code for ratings table.


tatiana-yan:
JTALL

PR #10873, opened 2019-05-14
[categories] Removed the '^' symbol.

mpimenov:
Now that we have data/strings/types_strings.txt, there is
no need in the special syntax to mark the displayed name in categories.txt.

### DIFF
--- a/data/categories.txt
+++ b/data/categories.txt
@@ -1,17 +1,16 @@
 # Category groups referenced in many categories, should be defined before being referenced.
 #
 # Syntax:
-# @   - string that starts with this symbol indicates the beginning of the group translations
-#       which are used for the search.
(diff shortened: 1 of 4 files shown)


tatiana-yan:
JTMAC

data/categories.txt
```diff
@@ -9,7 +9,6 @@
 # 1-9 - digits in front of a word indicate the number of symbols than need to be typed in a search query to make
 #       this word appear in the list of suggestions. Located immediately after the separating characters ':' or '|'.
 #       Acceptable use only one digit for one word.
-# ^   - symbol that indicates that the translation should be used in the user interface to identify the category in
```
milchakov: еще две строчки вниз нужно удалить

mpimenov: done

PR #10888, opened 2019-05-16
[std] Drop string_view.hpp

cc-engineering:


cc-engineering:
JTALL

mpimenov:
Rationale:
1. We want to be on the safer side with using std::experimental on Android and iOS, which is to say, not use it at all.
2. We could reimplement string_view under omim/std/ but collectively agreed not to do so. While it looks like a better decision in general, this particular class seems too dangerous.

PR #10936, opened 2019-05-23
[Strings] Change of Palestine name in Download maps menu with accordance to OpenStreetMap

ForkWithMe:
MAPSME-10648


data/countries_names.txt
```diff
@@ -23342,66 +23342,66 @@
   sk = Palau
 
 [Palestine]
```
mpimenov: What about [Palestine Description]?

PR #10942, opened 2019-05-23
[search] Fix search synonyms for China

tatiana-yan:


generator/search_index_builder.cpp
```diff
@@ -77,7 +77,9 @@ class SynonymsHolder
         for (size_t i = 1; i < tokens.size(); ++i)
         {
           strings::Trim(tokens[i]);
-          // synonym should not has any spaces
+          // For consistency synonym should not have any spaces.
+          // In case we'll have "Russia" -> "Russian Federation" mapping
+          // feature with name "Russia" will match request "federation". It is wrong.
           ASSERT(tokens[i].find_first_of(" \t") == string::npos, ());
```
mpimenov:
```
          // For consistency, synonyms should not have any spaces.
          // For example, the hypothetical "Russia" -> "Russian Federation" mapping
          // would have the feature with name "Russia" match the request "federation". It would be wrong.
          ASSERT(tokens[i].find_first_of(" \t") == string::npos, ());
```

tatiana-yan: done

PR #11090, opened 2019-06-21
[iOS Sound TTS] new order for languages

ForkWithMe:
New sorting by alphabet with few exceptions in accordance with Android version for consistency of user experience.


ForkWithMe:
There is no need in launching of tts_languages.py, as originally it was used for generating languges.hpp and it will be deleted later by bykoianko.

PR #11335, opened 2019-08-02
[generator:regions:test] Fix: remove temporary files

cc-engineering:


mpimenov:
Consider using platform/platform_tests_support/scoped_file.hpp

PR #11425, opened 2019-08-13
[routing] Changed IndexRouter::AreMwmsNear

gmoryes:
In previous version start and finish could have only one
mwm related to them, but after adding several fake
segments, start and finish could be linked with
more than one mwm, because of that AreMwmsNear 
returns false when start or finish placed near mwm's 
border. Because of that LeapsOnly is chosen instead of
Joints mode.

### DIFF
--- a/routing/routing_integration_tests/route_test.cpp
+++ b/routing/routing_integration_tests/route_test.cpp
@@ -559,4 +559,14 @@ namespace
 
     TEST_EQUAL(route.second, RouterResultCode::NoError, ());
   }
+
+  UNIT_TEST(AreMwmsNear_HelsinkiPiter)
+  {
+    TRouteResult route =
+        integration::CalculateRoute(integration::GetVehicleComponents(VehicleType::Car),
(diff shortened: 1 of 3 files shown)


gmoryes:
пример маршрута (строит после текущего PR):
image

краш сейчас на релизе:
image

тест добавил

gmoryes:
JTRIT

gmoryes:
JTRIT

routing/routing_integration_tests/route_test.cpp
```diff
@@ -559,4 +559,14 @@ namespace
 
     TEST_EQUAL(route.second, RouterResultCode::NoError, ());
   }
+
+  UNIT_TEST(NoCrash_HelsinkiPiter)
```
mpimenov: nit: s/NoCrash/AreMwmsNear/

all tests are supposed to be "no crash"

gmoryes: fixed

PR #11429, opened 2019-08-13
[routing] Changed IndexRouter::AreMwmsNear

gmoryes:
In previous version start and finish could have only one
mwm related to them, but after adding several fake
segments, start and finish could be linked with
different mwms, because of that AreMwmsNear returns
false when start or finish placed near mwm's border.
Because of that LeapsOnly is chosen instead of
Joints mode.


gmoryes:
JTRIT

PR #11491, opened 2019-08-20
[search] Fixed an out-of-bounds error.

mpimenov:
Tracer assumed that every token will have a type but we
do not have a special type for unrecognized tokens and
set their type to COUNT instead. With relaxed parsing,
some tokens were left unclassified and this led to an error.


mpimenov:
JTALL

PR #11506, opened 2019-08-22
[iOS] Added connection error placeholder to the catalog web view

niakrise:
https://jira.mail.ru/browse/MAPSME-11589


iphone/Maps/Bookmarks/Catalog/CatalogConnectionError.swift
```diff
@@ -0,0 +1,47 @@
+import UIKit
+
+class CatalogConnectionError: UIView {
```
beloal: CatalogConnectionErrorView

PR #11517, opened 2019-08-23
[vulkan] Workaround for vkCreatePipelineCache unspecified return codes

rokuz:
https://fabric.io/mapsme/android/apps/com.mapswithme.maps.pro/issues/a2aa6fd202395e5678b4e4afe0fc9708?time=last-ninety-days


rokuz:
JTALL

PR #11560, opened 2019-08-28
Add aloha stage define to list of known defines

vicpopov:


omim_config.h
```diff
@@ -11,3 +11,4 @@
 //#define STAGE_DLINKS
 //#define STAGE_BOOKING_EXTENDED_INFO
 //#define STAGE_PICTURES_SERVER
+//#define STAGE_ALOHALYTICS_SERVER
```
mpimenov: мб отсортировать?

PR #11596, opened 2019-08-30
Added and fixed Polish strings

strebski:


strebski:
CLA signed.

data/strings/types_strings.txt
```diff
@@ -12603,9 +12658,11 @@
 
   [type.natural.tree_row]
     en = Tree row
+    pl - Rząd drzew
```
mpimenov: s/-/=/

data/strings/types_strings.txt
```diff
@@ -13143,6 +13207,7 @@
 
   [type.place]
     en = place
+    pl = miejsce
```
mpimenov: s/miejsce/Miejsce/

data/strings/types_strings.txt
```diff
@@ -2219,7 +2223,7 @@
     ja = 神社
     ko = 성지
     nb = Alter
-    pl = Kapliczka
+    pl = Kapliczka szintoistyczna
```
mpimenov: Please remove the trailing space.

mpimenov:
@strebski ping

data/strings/types_strings.txt
strebski: Yes, there is a unnecessary space, I'll fix it, thanks for info

strebski: @mpimenov Have no idea how to edit my commit :-( Could you assist?

mpimenov: It's ok if you make a separate commit with the changes.

strebski: > It's ok if you make a separate commit with the changes.

OK, done - all fixes! And I've added some other localized strings. Already should be fine.

PR #11606, opened 2019-09-02
[bookmarks][android][iOS] Limit indexed bookmarks count.

darina:


map/bookmark_manager.cpp
```diff
@@ -1542,9 +1543,30 @@ bool BookmarkManager::IsVisible(kml::MarkGroupId groupId) const
...
-  // TODO(@darina) Implement.
   CHECK_THREAD_CHECKER(m_threadChecker, ());
-  return true;
+  CHECK(m_callbacks.m_getSearchAPI != nullptr, ());
+
+  if (m_callbacks.m_getSearchAPI().IsIndexingOfBookmarkGroupEnabled(groupId))
+    return true;
+
+  size_t indexedBookmarksCount = 0;
+  for (auto indexableGroupId : m_callbacks.m_getSearchAPI().GetIndexableGroups())
```
mpimenov: const

PR #11613, opened 2019-09-03
[search] [bookmarks] Limited the number of bookmark results.

mpimenov:


search/bookmarks/processor.cpp
```diff
@@ -261,8 +261,13 @@ void Processor::Search(Params const & params) const
   BailIfCancelled();
   sort(idInfos.begin(), idInfos.end());
 
+  size_t numEmitted = 0;
   for (auto const & idInfo : idInfos)
+  {
+    if (numEmitted >= params.m_maxNumResults)
```
darina: numEmitted не инкрементируется

mpimenov: done

PR #11676, opened 2019-09-12
Updated and fixed Polish strings

strebski:
More natural and male/female neutral forms

### DIFF
--- a/data/sound.txt
+++ b/data/sound.txt
@@ -48,7 +48,7 @@
     it = Gira a destra.
     ja = 右に曲がります。
     ko = 우회전하십시오.
-    pl = Jedź w prawo.
+    pl = Skręć w prawo.
     pt = Vire à direita.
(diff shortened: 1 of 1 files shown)


bykoianko:
Thank you, but to make it work you need
- to install twin
- to run https://github.com/mapsme/omim/blob/master/tools/unix/generate_localizations.sh
then you see some difference in: https://github.com/mapsme/omim/tree/master/data/sound-strings
- then to make a new commit with the diff.

mpimenov:
@bykoianko I think we can do this part ourselves.

bykoianko:
> @bykoianko I think we can do this part ourselves.

All right. I did a PR using the commit by Mr. @strebski 

https://github.com/mapsme/omim/pull/11700

@strebski @mpimenov PTAL

bykoianko:
Thank you @strebski. I put your commit to #11700.

PR #11685, opened 2019-09-12
Updated and polished Polish strings

strebski:


data/categories.txt
```diff
@@ -5520,6 +5529,7 @@ it:Ufficio aziendale
 ja:会社事務所
 ko:회사
 nb:Firmakontor
+pl:Biuro przedsiębiorstwa
```
mpimenov: Note that this file contains synonyms for the search, so we expect a user to type them in.
Is it reasonable to search for "Biuro przedsiębiorstwa"?

https://github.com/mapsme/omim/wiki/categories.txt

PR #11763, opened 2019-09-23
[std] Removed most of the remaining old style std/ includes.

mpimenov:


mpimenov:
JTALL

3party/sdf_image/sdf_image.cpp
```diff
@@ -44,7 +44,7 @@ namespace
   }
 
 }
-#define BIND_GRADIENT(f) bind(&f, _1, _2, _3, _4, _5, _6, _7, _8)
+#define BIND_GRADIENT(f) std::bind(&f, _1, _2, _3, _4, _5, _6, _7, _8)
```
vmihaylenko: ```std::placeholders``` for ```_1, 2, ...```?

mpimenov: There is a ```using namespace std::placeholders;``` at the top.

PR #11871, opened 2019-10-09
Added and fixed Polish phrases

strebski:
Some autocorrection from previous update

### DIFF
--- a/data/strings/types_strings.txt
+++ b/data/strings/types_strings.txt
@@ -3392,7 +3392,7 @@
 
   [type.boundary.administrative.county]
     en = boundary-administrative-county
-    pl = Granica państwa
+    pl = Granica hrabstwa
 
   [type.boundary.administrative.municipality]
     en = boundary-administrative-municipality
@@ -10679,6 +10679,7 @@
 
   [type.landuse]
     en = landuse
+    pl = Zagospodarowanie ziemi
 
   [type.landuse.allotments]
     en = Allotments
@@ -11177,6 +11178,7 @@
 
   [type.leisure]
     en = leisure
+    pl = Wypoczynek
 
   [type.leisure.common]
     en = leisure-common
@@ -13275,6 +13277,7 @@
 
   [type.olympics]
     en = olympics
+    pl = Olimpiada
 
   [type.olympics.attraction]
     en = olympics-attraction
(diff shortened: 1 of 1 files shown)


data/strings/types_strings.txt
```diff
@@ -3359,27 +3384,34 @@
...
     en = boundary-administrative-city
+    pl = Granica miasta
 
   [type.boundary.administrative.country]
     en = boundary-administrative-country
+    pl = Granica państwa
 
   [type.boundary.administrative.county]
     en = boundary-administrative-county
+    pl = Granica państwa
```
mpimenov: Are you sure?
https://wiki.openstreetmap.org/wiki/Tag:place%3Dcounty

data/strings/types_strings.txt
```diff
@@ -11093,19 +11153,23 @@
...
     en = landuse-retail
     pl = Obszar handlu
 
   [type.landuse.salt_pond]
     en = landuse-salt_pond
+    pl = Słone jezioro
 
   [type.landuse.village_green]
     en = landuse-village_green
+    pl = Nawsie
```
mpimenov: are you sure?
https://pl.wikipedia.org/wiki/Nawsie

data/strings/types_strings.txt
strebski: You're right. Should be "pl = Hrabstwo", like it is described on Polish version of mentioned wiki site. Thanks for control!

data/strings/types_strings.txt
strebski: Yes, in this issue "Nawsie" is correct. Another meaning of this word: https://pl.m.wikipedia.org/wiki/Nawsie_(urbanistyka)
Also check
https://wiki.openstreetmap.org/wiki/Pl:Tag:landuse%3Dvillage_green

data/strings/types_strings.txt
strebski: "pl = Granica hrabstwa", of course.

data/strings/types_strings.txt
mpimenov: Looks like a niche term but ok.

strebski: "Nawsie" is not popular Polish word, you are right :-)

data/strings/types_strings.txt
strebski: County boundary already fixed!

PR #11879, opened 2019-10-10
[search] Add place to street synonyms

tatiana-yan:
place is used as street synonym in UK


mpimenov:
An example: https://en.wikipedia.org/wiki/Portland_Place

mpimenov:
Another one: https://en.wikipedia.org/wiki/Cadogan_Place

From a cursory look, these are usually named after people.

PR #12073, opened 2019-11-19
[base] Made the call to CancellationStatus update the status before return.

mpimenov:
Review fixes for https://github.com/mapsme/omim/pull/11933.


milchakov:
а, увидел твой ответ в другом pr

PR #12077, opened 2019-11-19
Translations

maksimandrianov:
так как ждем ответа по pl у @strebski, строки пока не генерировал

### DIFF
--- a/data/strings/types_strings.txt
+++ b/data/strings/types_strings.txt
@@ -2000,11 +2000,11 @@
     tr = Park alanı
     uk = Паркувальне місце
     vi = Chỗ để xe
-    el = Χώρος στάθμευσης
-    sk = Parkovacie miesto
     sw = Sehemu ya maegesho
     zh-Hans = 停车地点
     zh-Hant = 停車地點
+    el = Χώρος στάθμευσης
+    sk = Parkovacie miesto
     fa = جای پارک
 
   [type.amenity.parking_space.permissive]
@@ -10832,7 +10832,7 @@
     ja = 教会付属墓地
     ko = 교회
     nb = Kirkegård
-    pl = Dziedziniec przykościelny
+    pl = Teren kościelny
     pt = Adro
     pt-BR = Igreja
     ro = Curtea Bisericii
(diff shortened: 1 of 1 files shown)


maksimandrianov:
@strebski please take a look

tatiana-yan:
generate_localizations?

data/strings/types_strings.txt
```diff
@@ -1975,7 +1975,37 @@
...
+    pt-BR = Estacionamento
+    ro = Loc de parcare
+    es = Plaza de aparcamiento
+    es-MX = Plaza de aparcamiento
+    sv = Parkeringsplats
+    th = พื้นที่จอดรถ
+    tr = Park alanı
+    uk = Паркувальне місце
+    vi = Chỗ để xe
+    el = Χώρος στάθμευσης
```
mpimenov: el, sk, fa after zh-Hant

mpimenov:
@strebski PTAL

strebski:
Pardon, I was outside Github last data:
[strings] type.landuse.churchyard
pl = Teren przykościelny

[strings] type.amenity.parking_space
pl = Miejsce parkingowe

Cannot update strings.txt myself, working only on smartphone - or give me pls 2 days :-)

mpimenov:
@strebski The translation by our contractor is in conflict with yours. Which do you think is better?

strebski:
"Miejsce parkingowe" is indisputable. Simply correct.
We can think about "Teren przykościelny" - alternative is "Teren kościelny"; but first (my proposition) is strictly related with place (church and it's surroundings) when second option is related (in Polish language) more with property (as a terrain which belongs to the church).
So I suggest my both propositions as correct in our map context.

mpimenov:
The kościelny/przykościelny look similar enough to me, I think we'll leave the "przykościelny" which according to you has the meaning "church and it's surroundings".

I'm more interested in the distinction between "Dziedziniec" and "Teren".

mpimenov:
@strebski We'll go with your original suggestion for now. Thanks.

strebski:
"Dziedziniec" is very old-fashioned and elegant word, more related with castles and manors.
"Teren" is neutral and most popular (simple) word.
All best!

PR #12159, opened 2019-11-28
[omim] Add param for clang-format for sorting each block separately

gmoryes:
https://releases.llvm.org/8.0.0/tools/clang/docs/ClangFormatStyleOptions.html


mpimenov:
(IncludeBlocks: Preserve in an upcoming PR)

PR #12172, opened 2019-11-28
[iOS] Fixed layout in Subscription Views

Polas:
Fixed view controller orientation for iphone, discount label constraints, description stackView


ToshUxanoff:
JTIOS

PR #12198, opened 2019-12-03
[routing] New filter for cross mwm segments.

gmoryes:
Drop twin segments if they have another direction,
this could happend because of different mwms versions.


bykoianko:
JTRIT

bykoianko:
JTRQ

bykoianko:
JTMAC

PR #12359, opened 2019-12-25
[search] Add SkipAlways to TypesSkipper.

tatiana-yan:


tatiana-yan:
JTLINUX

data/mixed_nodes.txt
```diff
@@ -1,3 +1,5 @@
+# Add types which should be skipped for search index to search::TypesSkipper:m_skipAlways
```
mpimenov: nit: s/for/when building/

tatiana-yan: done

PR #12450, opened 2020-01-17
[generator] Fix generator integration tests

gmoryes:
Tests were broken after ignoring barriers on some highways

I see two types of disappearing accesses:
1) Because of  "ignoring barriers on some highways"
2) Because there are lots of ways with tag barrier=gate or something else, for example this: https://www.openstreetmap.org/way/97663995

The 2) is good for us, because such ways anyway break road graph and we will just store less info about way-accesses

Also there was a bug. If elem has "foot=yes" and "access=no", we will skip "yes" tag and apply only "no", so now we prefer "yes" tags if they exist


gmoryes:
JTGITR

PR #12518, opened 2020-01-29
[search_quality] Add categories filter modes.

tatiana-yan:


search/search_quality/aloha_to_samples_tool/aloha_to_samples_tool.cpp
```diff
@@ -256,5 +268,5 @@ int main(int argc, char * argv[])
       }
     }
   }
-  return 0;
+  return EXIT_SUCCESS;
```
mpimenov: include cstdlib

tatiana-yan: done

PR #12632, opened 2020-02-19
[isolines][core][geometry] Isolines fixes.

darina:
https://jira.mail.ru/browse/MAPSME-13196


mpimenov:
> int64_t constexpr kMinDataVersion = 200209;

Can we store this in MwmTraits?

darina:
> > int64_t constexpr kMinDataVersion = 200209;
> Can we store this in MwmTraits?

Переделала на MwmTraits. @mpimenov PTAL

PR #12664, opened 2020-02-25
cherry-pick [minizip] return to static linking for android (because of mystic cra…

milchakov:
…sh logs).

cp for https://github.com/mapsme/omim/pull/12663


milchakov:
@mpimenov PTAL

PR #12687, opened 2020-02-28
Mapsme 13288 update ndk via buildgradle

alexzatsepin:
Updated NDK version to 21 via build.gradle. Build will failed if the specified NDK is not found. Now it's the last stable version - 21. 

Important note: ndk.dir in local.properties must be point on ndk of same version as build.gradle ndk version or deleted from local.properties.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,9 +8,10 @@ buildscript {
   }
 
   dependencies {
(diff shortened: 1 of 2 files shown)


ToshUxanoff:
JTANDROID

ToshUxanoff:
JTANDROID

PR #12714, opened 2020-03-05
[pykmlib] Track points with altitudes.

darina:


kml/pykmlib/bindings.cpp
```diff
@@ -125,6 +126,42 @@ struct LocalizableStringAdapter
   }
 };
 
+std::string LatLonToString(ms::LatLon const & latLon);
+
+struct PointWithAltitudeAdapter
+{
+  static m2::PointD const & GetPoint(geometry::PointWithAltitude & ptWithAlt)
```
milchakov: `geometry::PointWithAltitude const &`?  и в методе ниже

darina: поправила

PR #12776, opened 2020-03-21
Add FREENOW_CLIENT_ID and FREENOW_CLIENT_SECRET definitions

VMironiuk:
Fix `Use of undeclared identifier 'FREENOW_CLIENT_ID'` and `Use of undeclared identifier 'FREENOW_CLIENT_SECRET'` issues during compilation.


milchakov:
thanks

PR #13057, opened 2020-05-10
Autodetect and use ccache on Linux and macOS

shaforostoff:
ccache allows to speed up builds from scratch (very useful for CI) and also in cases when developer switches between branches a lot. On macOS it can be installed using brew.sh

ld.gold reduces linking time in half, but is only available for ELF objects (Linux; note, I did not check if it works for android)


biodranik:
LGTM :) +100500 to ccache

mpimenov:
Hello. Please sign the CLA: https://maps.me/cla

CMakeLists.txt
```diff
@@ -117,6 +117,24 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(EXECUTABLE_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR})
 set(LIBRARY_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR})
 
+if (UNIX)
+  find_program(CCACHE_FOUND ccache)
+  if(CCACHE_FOUND)
```
mpimenov, line 122: space after if

CMakeLists.txt
```diff
@@ -117,6 +117,24 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(EXECUTABLE_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR})
 set(LIBRARY_OUTPUT_PATH ${CMAKE_CURRENT_BINARY_DIR})
 
+if (UNIX)
```
tatiana-yan, line 120: PLATFORM_MAC?

CMakeLists.txt
```diff
@@ -117,6 +117,24 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
+  find_program(CCACHE_FOUND ccache)
+  if(CCACHE_FOUND)
+    message(STATUS "Using ccache")
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
+    set(CMAKE_C_COMPILER_LAUNCHER   "${CCACHE_PROGRAM}")
+    set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
+  endif(CCACHE_FOUND)
```
mpimenov, line 128: s/endif(CCACHE_FOUND)/endif()/

CMakeLists.txt
```diff
@@ -117,6 +117,24 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
+endif()
+if (PLATFORM_LINUX)
```
mpimenov, line 130: empty line before this one, please

CMakeLists.txt
tatiana-yan, line 122: can you please add space after `if`?

CMakeLists.txt
```diff
@@ -117,6 +117,24 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
+  find_program(GOLD_FOUND ld.gold)
+  if(GOLD_FOUND)
```
tatiana-yan, line 132: can you please add space after `if`?

CMakeLists.txt
mpimenov, line 122: please add a flag that would explicitly enable ccache and log the error if ccache was not found

CMakeLists.txt
shaforostoff: UNIX will be true on both linux and mac, or would you prefer your custom variables to be used instead?

CMakeLists.txt
shaforostoff: fixed. Should I resolve discussion myself, or is it responsibility of the reviewer?

CMakeLists.txt
shaforostoff: fixed. Should I resolve discussion myself, or is it responsibility of the reviewer?

CMakeLists.txt
shaforostoff: fixed. Should I resolve discussion myself, or is it responsibility of the reviewer?

CMakeLists.txt
shaforostoff: fixed. Should I resolve discussion myself, or is it responsibility of the reviewer?

CMakeLists.txt
shaforostoff: fixed

shaforostoff:
> Hello. Please sign the CLA: https://maps.me/cla

i have signed it, please check

CMakeLists.txt
tatiana-yan: Custom variables are better because it explicitly indicates that we do not want to apply it for iOS and Android (which are unix-like systems).

We do not change instruments for mobile platforms without complex research & tests.

CMakeLists.txt
tatiana-yan: It is up to reviewer.

shaforostoff: so you want ccache to be used only when appropriate flag is passed to cmake?

In other words, do you want the default value of this flag to be OFF or ON?

mpimenov: Yes: USE_CCACHE, which should be false by default.

CMakeLists.txt
shaforostoff: understood. changed.

CMakeLists.txt
shaforostoff: done

CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ option(PYBINDINGS "Create makefiles for building python bindings" OFF)
 option(SKIP_DESKTOP "Skip building of desktop application" OFF)
 option(BUILD_MAPSHOT "Build mapshot tool" OFF)
 option(USE_PCH "Use precompiled headers" OFF)
+option(USE_CCACHE "Cache object files with ccache" OFF)
```
mpimenov, line 75: one second thought, maybe add USE_GOLD too (also off by default) in case it doesn't work for somebody?

shaforostoff: ld.gold is part of binutils since 2008, and, to my knowledge, the only known unsolvable problem it has is linking kernel modules.

ld.gold might reveal some other problems, like forgetting to explicitly link to pthread or rt, according to https://stackoverflow.com/questions/3476093/replacing-ld-with-gold-any-experience but then those problems should be fixed instead, I believe.

So unless you insist on this option, I would prefer not to add it

PR #13058, opened 2020-05-10
fix build on macOS catalina

shaforostoff:


mpimenov:
@shaforostoff ping

shaforostoff:
i can build_omim.sh -r of latest master fine, however I still would prefer the warning fixes to be merged. I can change the title to reflect the reality

indexer/ftypes_matcher.hpp
```diff
@@ -238,7 +238,7 @@ class IsPlaceChecker : public BaseChecker
 
 class IsBridgeChecker : public BaseChecker
 {
-  virtual bool IsMatched(uint32_t type) const override;
+  bool IsMatched(uint32_t type) const override;
```
mpimenov, line 241: Why?

indexer/ftypes_matcher.hpp
```diff
@@ -282,7 +282,7 @@ class IsCoastlineChecker : public BaseChecker
 class IsHotelChecker : public BaseChecker
 {
 public:
-  enum class Type
+  enum class Type: unsigned
```
mpimenov, line 285: Why?

indexer/ftypes_matcher.hpp
```diff
@@ -334,7 +334,7 @@ class IsWifiChecker : public BaseChecker
 class IsEatChecker : public BaseChecker
 {
 public:
-  enum class Type
+  enum class Type: uint32_t
```
mpimenov, line 337: Why?

indexer/ftypes_matcher.hpp
shaforostoff: it is a bad idea to have a class with virtual methods but with no virtual destructor. ideally, besides removing virtual specified, this class should also be marked as final

indexer/ftypes_matcher.hpp
shaforostoff: this a compile warning fix:
comparison of integer expressions of different signedness: ‘std::underlying_type_t<ftypes::IsHotelChecker::Type>’ {aka ‘int’} and ‘long unsigned int’

indexer/ftypes_matcher.hpp
shaforostoff: same as above

PR #13063, opened 2020-05-12
[bookmarks] Added a new method of getting the category id by the serv…

darina:
…er id.


darina:
@beloal @yoksnod @alexzatsepin FYI

Issue #11976, opened 2019-10-27
Software renderer not building


utack:
Hi devs,  

I am trying to build 2a88c6ca5f90a14fdc6fb6ed47ec8dc86548187d on arch linux currently.

In the CMakeLists I enabled mapshot, to generate png images of maps.

Unfortunately the software render fails to build:

/home/jakob/Downloads/omim/software_renderer/cpu_drawer.hpp: In constructor ‘software_renderer::CPUDrawer::PointShape::PointShape(const PointD&, dp::Anchor, software_renderer::CPUDrawer::EShapeType, const software_renderer::DrawRule&, int)’:
/home/jakob/Downloads/omim/base/src_point.hpp:44:21: error: ‘SrcPoint’ is not a member of ‘software_renderer::base’
   44 | #define SRC() base::SrcPoint(__FILE__, __LINE__, __FUNCTION__, "()")

Could you please look into it? I am hoping to generate some maps to print.
Thank you very much

gmoryes:
Could you give please the full log?

utack:
Of course I can provide the full log, thank you for looking into it!

Freshly cloned, at fee8ebf513d9a4da930a2a067a95a9628dcb3f78, then configure.sh without arguments

Changed one line in CMakeLists to build mapshot and hence the software renderer:
option(BUILD_MAPSHOT "Build mapshot tool" ON)

Replaced PROCESSES=* in ./tools/unix/build_omim.sh with PROCESSES=1 to only run 1 build-task and keep the log in order

then started a fresh debug build
./tools/unix/build_omim.sh -cd

The full log of this is attached
[buildlog.txt](https://github.com/mapsme/omim/files/3802271/buildlog.txt)

gmoryes:
> Of course I can provide the full log, thank you for looking into it!
> 
> Freshly cloned, at [fee8ebf](https://github.com/mapsme/omim/commit/fee8ebf513d9a4da930a2a067a95a9628dcb3f78), then configure.sh without arguments
> 
> Changed one line in CMakeLists to build mapshot and hence the software renderer:
> option(BUILD_MAPSHOT "Build mapshot tool" ON)
> 
> Replaced PROCESSES=* in ./tools/unix/build_omim.sh with PROCESSES=1 to only run 1 build-task and keep the log in order
> 
> then started a fresh debug build
> ./tools/unix/build_omim.sh -cd
> 
> The full log of this is attached
[...]

PR #13151, opened 2020-05-26
[software_renderer] Fix build.

mpimenov:
Missing std includes resulted in build problems on some systems
as reported here: https://github.com/mapsme/omim/issues/11976

### DIFF
--- a/software_renderer/rect.h
+++ b/software_renderer/rect.h
@@ -3,8 +3,11 @@
 
 #include "software_renderer/point.h"
 
+#include <algorithm>
+#include <cmath>
 #include <iostream>
-#include <climits>
+#include <limits>
+#include <vector>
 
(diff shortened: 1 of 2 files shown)


mpimenov:
JTALL

software_renderer/rect.h
```diff
@@ -3,8 +3,10 @@
 
 #include "software_renderer/point.h"
 
+#include <algorithm>
```
tatiana-yan, line 6: std::min далее?

tatiana-yan, line 6: не нужно include cmath и s/fabs/std::fabs далее?

mpimenov: done

mpimenov:
JTALL

PR #13339, opened 2020-06-24
[guides on map] zoom out by suggested zoom fix

milchakov:


milchakov:
@darina @vicpopov PTAL

PR #13377, opened 2020-07-07
[map] Remove using namespace routing from transit reader.

mesozoic-drones:
По аналогии с #13376.


mesozoic-drones:
JTALL

mesozoic-drones:
JTANDROID

PR #13540, opened 2020-08-11
[cmake] fix build with USE_PCH=On on Linux/gcc

Dushistov:
The project builds using c++17 standard, so if compile headers with
-std=c++14 compiler complains about "optional" and so on things,
that part of c++17 standard


tomilov:
JTLINUX
JTLINUXGCC

tatiana-yan:
JTLINUX

tatiana-yan:
JTLINUXGCC

Issue #13561, opened 2020-08-18
Add "tobacco shop" category


FinixFighter:
Hi! Using the app I found another category not implemented yet, the "tobacco shop" one. Could it be added?

ThanoS-max:
I guess adding tobacco shop would be not necessary because in some countries it is not allowed to mention tobacco shops.

FinixFighter:
> I guess adding tobacco shop would be not necessary because in some countries it is not allowed to mention tobacco shops.

The category could be disabled for those countries; furthermore, maybe those countries don't even have tobacco shops, so the problem doesn't exist? Can you please give me a reference to this for a country? Just for curiosity.

ThanoS-max:
That is again true because the country I live in (India) does not support Google Street View due to some privacy issues.

PR #13564, opened 2020-08-18
Adding the tobacco shop category

FinixFighter:
Adding the tobacco shop category


FinixFighter:
#13561

tatiana-yan:
Hello.
Categories.txt is search synonyms file. It does not add any osm objects to offline data.

PR #13581, opened 2020-08-20
speedup compilation with help of cmake 3.6 pch support

Dushistov:
Since cmake 3.16 there is cross-platform support of precompiled
headers in cmake. So I add code to check version of cmake,
and if it is recent enough build system switches to cmake's pch
support.

I tested with such command:
touch qt/main.cpp && time ninja MAPS.ME
Before: ~ 35 seconds on my machine (average)
After: ~ 21 seconds (average)

What make this speedup possible:
1. With cmake's pch support generation of precompiled headers
  happens only if something changed, while current PCH script cause
  recompilation of headers every run of ninja. It takes ~ 4 seconds
2. Now it is possible to set precompiled headers per target,
   and I include map/framework.hpp for "qt".

### DIFF
--- a/cmake/OmimHelpers.cmake
+++ b/cmake/OmimHelpers.cmake
@@ -273,74 +273,87 @@ function(add_pic_pch_target header pch_target_name
 endfunction()
 
 function(add_precompiled_headers header pch_target_name)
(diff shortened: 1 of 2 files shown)


pastk:
Looks like https://github.com/organicmaps/organicmaps could benefit from this PR?

PR #13599, opened 2020-08-24
[iOS] Added precondition in map search

Polas:
Для поиска креша.
image


beloal:
Исправлено в https://github.com/mapsme/omim/pull/13646

PR #13835, opened 2020-10-19
[pybindings] do not search for python libraries

vicpopov:
We don't need to find them since we don't link to them anymore (since #13695)


mpimenov (approved):
LGTM

maksimandrianov (approved):
LGTM

PR #13916, opened 2020-11-03
Arsenal strings

milchakov:
https://jira.mail.ru/browse/MAPSME-15209

### DIFF
--- a/data/strings/strings.txt
+++ b/data/strings/strings.txt
@@ -49364,3 +49364,21 @@
   [taxi_category_order]
     en = Order
     ru = Заказать
+
+  [arsenal_telemed_map_downloader_title]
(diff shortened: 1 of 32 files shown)


data/strings/strings.txt
```diff
@@ -49364,3 +49364,21 @@
   [taxi_category_order]
     en = Order
     ru = Заказать
+
+  [arsenal_telemed_map_downloader_title]
+    en = Онлайн консультации профессиональных врачей 24/7
```
mpimenov, line 49369: note: we must have an "en" translation so this is a workaround, the correct locale is used in runtime

mpimenov (approved):
LGTM

PR #13925, opened 2020-11-05
[pybindings] pykmllib fix

milchakov:
Fix for
TypeError: No Python class registered for C++ class std::__1::vector<unsigned long long, std::__1::allocator<unsigned long long> >


mixilchenko:
Бинд и тест есть. Я ок

tomilov (approved):

PR #13940, opened 2020-11-10
[qt] Fix for ignoring function's val with 'warn_unused_result'.

mesozoic-drones:
Fix for warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result].


mesozoic-drones:
JTMAC

ldo2 (approved):

PR #14058, opened 2020-12-14
[bookmarks][looonely] Allow to turn on visibility for categories having visibilities of all constituents turned off.

tomilov:
Do not force visibility to false for categories that does not have visible constituents. 
https://jira.mail.ru/browse/MAPSME-15349


milchakov (approved):

milchakov:
название ветки зачетное))

PR #14102, opened 2021-01-16
[iOS] Add more keyboard shortcuts

ghost:
## Why?:
Lots of apps seem to implement their own shortcuts which creates confusion for the end-user. Zooming on an Apple Mac or iOS app has always used CMD + and CMD -.

So to be more in line with Apples defaults from apps (like Apple Maps) which use CMD + and CMD - I have added these as defaults.

A dedicated re-center button it useful when the map was moved or zoomed to quickly get back to the current location and default zoom level.

## Added:

`CMD +` for Zoom In (With an alternative of CMD = to also make it work on non numeric keyboards without much confusion.
`CMD -` for Zoom Out
`CMD 0` for recentering the map.

## Changed:
The original zoom in and zoom out on the arrow keys are still active but not shown in the help popup when holding down CMD for backwards compatibility. We might want to re-use these keys to move the map around sometime in the future.

### DIFF
--- a/iphone/Maps/Classes/MapViewController.mm
+++ b/iphone/Maps/Classes/MapViewController.mm
@@ -943,9 +943,14 @@ NSString *const kPP2BookmarkEditingSegue = @"PP2BookmarkEditing";
 }
 
 - (NSArray *)keyCommands {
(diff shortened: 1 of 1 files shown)


Fossil01:
Anyone wants to review this please? :-)

Fossil01:
Anyone?

